Issue #28, opened 2021-10-07
REGRESSION: OTA updates not functioning with expanded partition table


davepl:
**Problem**
The code contains OTA support that worked when the project was small enough to fit in the default partition table.  When the code grew, partitions_custom.csv was added to accommodate it.   Unfortunately doing so seems to have broken OTA support.

**Steps**

1. Build the SPECTRUM config, ensure ENABLE_OTA is on 
2.  Run it, note IP, set upload_port to it, upload
3.  Note that it connects but the OTA code bails internally when checking the partition table.  That's as far as I debugged.

BeldrothTheGold:
Oh, it looks like it might not be necessary to have a factory app. This might work? 

```
# ESP-IDF Partition Table
# Name,   Type, SubType,     Offset,      Size, Flags

nvs,        data,   nvs,     0x9000,    0x5000,
otadata,    data,   ota,     0xe000,    0x2000,
app0,       app,    ota_0,   0x10000,   0x150000,
app1,       app,    ota_1,   0x160000,  0x150000,
storage,    data,   spiffs,  0x330000,  0xF0000,   # for  4 MB
```

davepl:
I thought it was tested and working before it was checked in!  We only verify that it BUILDS.  I suppose given an ESP32 emulator… no, never mind.

It’s important to test any changes, as the pull requests get a code review and a buddy build but not a runtime test.

- Dave

BeldrothTheGold:
This seems to be resolved. I am no longer seeing crashes during OTA upload of the spectrum project.

rbergen:
Closing this as it was marked resolved over a year ago.

PR #36, opened 2021-10-10
Fix OTA regression

BeldrothTheGold:
## Description

Update partitions_custiom.csv to resolve OTA regression. Issue #28 

## Contributing requirements

### DIFF
--- a/partitions_custom.csv
+++ b/partitions_custom.csv
@@ -1,13 +1,12 @@
-# If you link in enough functionality that you exceed the default flash size,
-# this partition table can be enabled.  Assumes only 4M of flash available.
-#
 # ESP-IDF Partition Table
-# Name,   Type, SubType, Offset,  Size, Flags
+# This gives us some additional space for code. 
+# It should also fix the OTA regression.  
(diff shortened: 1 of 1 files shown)


davepl:
I just figured out  this is sort of a broadcast and not a direct reply, sorry!

 

I think (2) makes sense if you can at the same time generalize the two pieces of code that draw stats so we don’t wind up with N copies of.  There are 2 copies now, one for TFT and one for OLED.  They could be generalized if you added defs for DISPLAY_WIDTH and DISPLAY_HEIGHT, but you’ve also got to deal with the variable fonts and colors used on the OLED.

 

I’m fine with keeping it really simple, like 

 

drawstring(x, y, color, text, size = medium);

 

Then if USE_OLED is defined, drawstring used the M5.Lcdf.drawString and friends.  If USE_TFT is defined, it uses g_TFT.drawStr.  If you added support for small, medium, and large, that’d cover everything for now.

 

That might be roughly what you mean by display driver, I’ve just kept it down to one or two calls rather than making an API out of it, but go as wild as you feel necessary (and no wilder!).

 

Thanks,

Dave

Issue #73, opened 2021-10-27
std::bad_alloc exception in Arduino OTA handler


laufersteppenwolf:
## Bug report

**Problem**
The ESP32 repeatedly keeps on rebooting due to an unhandled exception (std::bad_alloc)
I don't know why exactly, but with the BROOKLYNROOM config, it happens the most.

The issue seems to be related to the Arduino OTA handler, see the stack trace below
src/main.cpp:674 relates to [this line](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/772a19ea361bc4f7c564d5504f82e55385b3d239/src/main.cpp#L674)

**Steps**

1.  build the BROOKLYNROOM config
2.  run the monitor
3.  catch the stacktrace

**Example**

**Notes**
Stack trace:
`
(I) (TerminateHandler)(C1) -------------------------------------------------------------------------------------
(I) (TerminateHandler)(C1) - NightDriverStrip Guru Meditation                              Unhandled Exception -
(I) (TerminateHandler)(C1) -------------------------------------------------------------------------------------
(I) (PrintOutputHeader)(C1) NightDriverStrip
(I)
(I) (PrintOutputHeader)(C1) -------------------------------------------------------------------------------------
(I) (PrintOutputHeader)(C1) M5STICKC: 0, USE_OLED: 1, USE_TFT: 0
(I) (PrintOutputHeader)(C1) Version 137: Wifi SSID: NETGEAR90 - ESP32 Free Memory: 68692, PSRAM:0, PSRAM Free: 0
(I) (PrintOutputHeader)(C1) ESP32 Clock Freq : 240 MHz
(I) (TerminateHandler)(C1) Terminated due to exception: std::bad_alloc

ELF file SHA256: 0000000000000000

Backtrace: 0x4008c84c:0x3ffb1e20 0x4008cac9:0x3ffb1e40 0x400dd3c9:0x3ffb1e60 0x4017f3f7:0x3ffb1e90 0x4017f45e:0x3ffb1eb0 0x4017e89b:0x3ffb1ed0 0x4017ee4e:0x3ffb1ef0 0x4017ea05:0x3ffb1f10 0x400e58dd:0x3ffb1f30 0x4010144e:0x3ffb1f70 0x400dd455:0x3ffb1f90 0x40104254:0x3ffb1fb0 0x4008db1a:0x3ffb1fd0
[...]

PR #75, opened 2021-10-28
Draft: Fix brooklynroom project build

BeldrothTheGold:
## Description
The purpose of this PR to fix the brooklynroom project's build errors. I also added a brooklynroom env in the platformio.ini file so the project can be added to the github CI if desired. 

Here are the build issues and how I solved them. So we're clear, I'm not sure these are the best ways to solve these issues. 

1) The FAN_SIZE setting for this project was previously set to NUM_LEDS, but there is no matching definition in faneffects.h for a FAN_SIZE > 16. This causes FanPixelsVertical not to be declared.   To solve this build error I have set the FAN_SIZE to 1 for the project in globals.h. 
   
2) I added #if USE_TFT sections around the M5 LCD update code in spectrumeffects.h to allow it to build on both heltec and m5 boards. I think this makes sense as NightDriverStrip is supposed to support I2S microphones. 

3) I added NUM_BANDS, NOISE_FLOOR, and NOISE_CUTOFF to the project in globals.h so audio.cpp and soundanalyzer.h would compile. 

4) I added variable declarations for mult, and prob above the AllEffects array in effects.cpp. @laufersteppenwolf noticed that several projects reference these variables in their effects, but they weren't declared.  So, we used this very simple solution to fix the build. 

As noted in #73 there seem to be issues with ArduinoOTA while using this project. This does not fix those issues. It just makes the project build and run. 

## Contributing requirements

### DIFF
--- a/src/effects.cpp
+++ b/src/effects.cpp
@@ -259,6 +259,8 @@ unique_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color)
 }
 #endif
 
+CRGB mult = CRGB::Black; float prob = 0.9; // These constants are referenced by several project's effects
+
 // AllEffects
 // 
 // A list of internal effects, if any.  
--- a/include/effects/spectrumeffects.h
+++ b/include/effects/spectrumeffects.h
@@ -118,8 +118,10 @@ class VUMeterEffect : public LEDStripEffect
 
         pGFXChannel->fillRect(0, yVU, MATRIX_WIDTH, 1, BLACK16);
 
+#if USE_TFT
         if (giInfoPage == 1)
             M5.Lcd.fillRect(0, 0, M5.Lcd.width(), 10, BLACK16);
+#endif
 
         if (iPeakVUy > 1)
         {
@@ -345,6 +347,7 @@ class SpectrumAnalyzerEffect : public VUMeterEffect
         else
             fillSolidOnAllChannels(CRGB::Black);
 
+#if USE_TFT
         if (giInfoPage == 1)
         {
             if (gbInfoPageDirty)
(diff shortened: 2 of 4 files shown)


src/effects.cpp
```diff
@@ -259,6 +259,8 @@ unique_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color)
 }
 #endif
 
+CRGB mult = CRGB::Black; float prob = 0.9; // These constants are referenced by several project's effects
```
rbergen, line 262: I'm wondering if setting these variables to fixed values in the cpp file is the best way to go. Maybe it would make sense to:
1. set their values using header defines,or
2. replace the references to the variables with such defines throughout effects.cpp, or even
3. replacing the references with the constants entirely?

BeldrothTheGold: Yup. I was thinking this wasn't a great solution. Though, the values that are used for these variables are probably highly dependent on preference, and the hardware used.

rbergen: As they are used in multiple projects' effects, it may make most sense to add a couple of defines to globals.h and use those in effects.cpp; that header already contains some "semi-configuration", anyway.

laufersteppenwolf: I agree
As I said before, the way I implemented those variables was for testing purposes only.
I, too, would add the defines to the globals header, and also add some default values, in case a project/config doesn't want to override it

BeldrothTheGold: Sounds good. I'll make the changes and force push when I have a chance.

davepl:
Thanks for sorting this out and fixing it for us!

PR #89, opened 2021-11-17
Add .click() handler to webserver checkboxes

xcalibur839:
## Description
Adds a .click() event handler to the checkboxes in index.html. 
When a checkbox is checked, EnableEffect() is called, otherwise DisableEffect() is called.

## Contributing requirements

### DIFF
--- a/data/index.html
+++ b/data/index.html
@@ -185,7 +185,7 @@
             });
         }
    
-        function EnableEffect(i) 
+        function DisableEffect(i) 
         {
             jQuery.support.cors = true;
             $.post(BasePageForRequests() +  "/disableEffect",
@@ -220,7 +220,14 @@
 
                 var inp = $('<input type="checkbox"/> ')
                     .addClass('single-checkbox')
-                    .appendTo(li);
+                    .appendTo(li)
+                    .click(function() {
+                        if ($(this).is(':checked')) {
+                            EnableEffect($(this).parent().attr('id'));
+                        } else {
+                            DisableEffect($(this).parent().attr('id'));
+                        }
+                    });
 
                 inp.prop('checked', effectData.Effects[i].enabled);
 


xcalibur839:
There are a couple of things to note with this PR:
- Disabling an effect that is currently running does not immediately stop the effect, it will continue to run until DEFAULT_EFFECT_INTERVAL has elapsed
- If you only have one effect in AllEffects[] and you disable that effect, the LEDs will appear to freeze in whatever their final state was after DEFAULT_EFFECT_INTERVAL has elapsed. This does not happen when there are multiple effects in AllEffects[] that are all disabled because of the fading between effects that sets all LEDs to Off/Black (i.e. the final state of the LEDs is Off/Black)

Fixing these issues will require additional changes to effecmanager.h (and possibly spiffswebserver.h, elsewhere).

rbergen:
@xcalibur839 Thank you for adding this, and the notes.

Concerning the notes, I think the decision to make here is whether they should be added as an issue (as in, [in the Issues section](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues)). Considerations are:

- Once this PR is merged, the notes will basically disappear out of view. The reason being that PRs are looked at far less often (I would say: hardly ever), once merged. Issues will remain visible until they themselves are closed. 
- The behavior you're describing is not an issue caused by the PR, but a description of how the "enable" and "disable" URLs work. Because of the PR, they just may become visible more often.

xcalibur839:
@rbergen Yes, that makes sense. These issues exist regardless of this PR, but this PR certainly makes them more visible. With that being said, this PR addresses an issue that is also not listed in the Issues section of the Repository. 

Would you like me to add all of the above as Issues to the Repo and then link this PR to the Checkbox Issue? I'm still very new to the proper procedures on GitHub, so if there is anything else I should have done or should be doing please don't hesitate to let me know!

rbergen:
@xcalibur839 I'd say we just move on with merging this PR (which I will do once I have confirmation you've read the comment I'm now writing) to solve the issue it solves; there's no longer a need to describe the issue that now that your PR has already very helpfully fixed it.

Concerning the delay and lingering of end state in relation to the disabling of effects, I think that opening an issue might indeed be appropriate. As both points relate to the disabling of effects, I also think it's perfectly okay to cover both of them in one single issue.

When it comes to "proper procedure" on GitHub: I wouldn't worry too much about violating it, if only because it's not set in stone. For one, what works best for a project depends on the nature of the project and the people involved in it. In general, open source projects really do appreciate genuine contributions like this one; this project certainly does. The peripheral details can be worked out organically as we go along.

xcalibur839:
@rbergen Sounds good, and thank you very much for the encouragement! I'll create that Issue now.

rbergen:
@xcalibur839 Thanks again for your contribution!

Issue #90, opened 2021-11-18
Unexpected behavior when effects are disabled via webserver


xcalibur839:
## Bug report

**Problem**

- Disabling an effect that is currently running does not immediately stop the effect, it will continue to run until DEFAULT_EFFECT_INTERVAL has elapsed
- If you only have one effect in AllEffects[] and you disable that effect, the LEDs will appear to freeze in whatever their final state was after DEFAULT_EFFECT_INTERVAL has elapsed. This does not happen when there are multiple effects in AllEffects[] that are all disabled, due to the fading between effects that sets all LEDs to Off/Black (i.e. the final state of the LEDs is Off/Black)

**Steps**

1.  Enable the webserver and Wi-Fi in globals.h and configure secrets.h
2.  Determine the IP address of the ESP32 and connect to the webserver with a web browser
3.  Disable all of the listed effects (or call the /disableEffect endpoint using the curl example below on every effect)

**Example**

curl -d 'effectIndex=0' 192.168.0.25/disableEffect

xcalibur839:
Working on [PR #93](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/93) has changed my view on this issue. For example, clicking on a disabled effect from the web page will immediately play that effect even though it is disabled. However, it will not play automatically as part of the CheckEffectTimerExpired()/NextEffect() loop.

So perhaps the first part of this issue is not actually an issue, but is working as intended. Specifically:
> Disabling an effect that is currently running does not immediately stop the effect

The effect is still disabled as far as the EffectManager auto play is concerned.

xcalibur839:
I think the biggest reason this behavior is unexpected is the lack of visual feedback for the currently running effect. When unchecking an effect in the list, there currently isn't any indication that unchecking the effect did anything until all of the effects are unchecked.

[PR #98](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/98) attempts to address this issue by providing visual feedback that unchecking an effect did register, as well as the current timeframe for the actively running effect. This behavior is very similar to something like Windows Media Player, where the currently playing song is selected, a progress bar shows how much time is remaining, and the play queue can be modified without stopping playback of the current song.

rbergen:
@xcalibur839 With #93 and #98 merged, I'm wondering what is left of this issue?

xcalibur839:
@rbergen I think you're correct and this is no longer an issue! Closing now, thank you for everything!

PR #93, opened 2021-11-23
Count enabled effects

xcalibur839:
## Description

With this fix, disabling all effects will now:
- Immediately turn off all LEDs
- No longer get stuck in NextEffect()

This partially addresses [Issue #90](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/90).

## Contributing requirements

### DIFF
--- a/include/effects/effectmanager.h
+++ b/include/effects/effectmanager.h
@@ -67,12 +67,13 @@ class EffectManager
 {
 	LEDStripEffect ** _ppEffects;
 	size_t            _cEffects;
+	size_t			 _cEnabled;
 
 	size_t			 _iCurrentEffect;
 	uint    		 _effectStartTime;
 	uint    		 _effectInterval;
 	bool             _bPlayAll;
-	
+
 	unique_ptr<bool []> _abEffectEnabled;
 	shared_ptr<LEDMatrixGFX> * _gfx;
 	unique_ptr<LEDStripEffect> _pRemoteEffect;
@@ -85,6 +86,7 @@ public:
 	EffectManager(LEDStripEffect ** pEffects, size_t cEffects, shared_ptr<LEDMatrixGFX> * gfx)
 		  : _ppEffects(pEffects),
 	  	    _cEffects(cEffects),
+			_cEnabled(0),
 		    _effectInterval(DEFAULT_EFFECT_INTERVAL),
 		    _bPlayAll(false),
 			_gfx(gfx)
@@ -157,6 +159,12 @@ public:
 			return;
 		}
 		_abEffectEnabled[i] = true;
+		
+		if (_cEnabled < 1)
+		{
+			ClearRemoteColor();
+		}
+		_cEnabled++;
 	}
 
 	void DisableEffect(size_t i)
@@ -167,6 +175,12 @@ public:
 			return;
 		}
 		_abEffectEnabled[i] = false;
+		
+		_cEnabled--;
+		if (_cEnabled < 1)
+		{
+			SetGlobalColor(CRGB::Black);
+		}
 	}
 
 	bool IsEffectEnabled(size_t i) const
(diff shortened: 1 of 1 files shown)


xcalibur839:
I'm not sure if this is the best approach for this fix, but it does seem to work well on my ESP32 WROOM and M5StickCPlus. 

Maybe something fancier could be done to reliably determine if all of the items in the _abEffectEnabled array are false without needing to iterate over the whole array. Maybe iterating over the array is the better option - I didn't test that approach due to scaling concerns.

rbergen:
I do think we need to reconsider this PR's approach for keeping track of how many effects are enabled. 

If `_cEnabled` is meant to keep track of how many effects have been enabled or disabled, then it should count _actual transitions_. The current code for the `EnableEffect` and `DisableEffect` set  a `true` or `false` value to the relevant member of the `_abEffectEnabled` array without considering if it is already set to the same value. This means that `_cEnabled` can become out of sync with the actual number of enabled effects if the `EnableEffect` or `DisableEffect` function is called multiple times in sequence for the same effect. This is exacerbated by the fact that `_cEnabled` is of type `size_t` and as such unsigned: if its value is 0 and then decreased due to a subsequent invocation of `DisableEffect`, it will roll over to a very high positive number.

I think we can stick to the current approach by adding a check if an effect is already enabled or disabled _before_ setting its state and counting the transition, and bailing out if an attempt is made to set an effect to a state it already has. Counting the set items in the `abEffectEnabled` array is an alternative indeed, and more resilient to (future) alternative approaches to setting or clearing the enabled states for effects besides calls to `EnableEffect`/`DisableEffect`.

xcalibur839:
Aha yes you're absolutely correct about the possibility for desynchronization, I'll submit a fix for that shortly.

What are your thoughts about this approach vs. iterating over the `_abEffectEnabled` array to get an exact count? The reason I didn't pursue that path in the first place was concern over the performance implications if there are a very large number of LEDs attached. But maybe those concerns are unwarranted, and the advantages outweigh any downsides.

Some thoughts:
I'm going to call the approach used in this PR the Turnstile approach and the alternative of iterating over the array the Iterative approach.


Iterative Pros:
- Count is effectively guaranteed to be accurate
- It doesn't matter how effects are enabled/disabled, the count will still be accurate


Iterative Cons:
- Potential performance implications when there are many LEDs attached (not tested or measured)
- The check will need to be invoked on every enable/disable to accomplish the same fix as the Turnstile approach


Turnstile Pros:
- Negligible performance hit, regardless of how many LEDs are attached
- No additional function calls necessary, everything is handled in `EnableEffect()`/`DisableEffect()`


Turnstile Cons:
- The count provided by `_cEnabled` may not accurately represent the count of enabled effects in the `_abEffectEnabled` array if it becomes desynchronized
- `EnableEffect()`/`DisableEffect()` must be used to enable/disable effects. Any other approach will cause `_cEnabled` to become desynchronized


There are probably other pros and cons to each approach that I'm not considering, so please don't consider this exhaustive and feel free to add or remove anything.

rbergen:
I'd say that because `_cEnabled`  is a private member of the class, it's okay to assume that member functions (public or otherwise) know when and how to use and update it. As it stands, the constructor already uses `EnableEffect()` itself, so the current pattern already is that all changes to the enabled state of effects does already take place via `EnableEffect()`/`DisableEffect()`. 

That makes what you call the Turnstile approach appropriate and less wasteful than iterating over the array.

xcalibur839:
That makes sense, thank you very much for the sanity check!

Those changes have now been added, as well as a public getter for the private `_cEnabled` member because I'm certain that information will be useful outside of the class as well.

rbergen:
I've given this one more look, and I think it's good to merge. Thanks again for your contribution!

PR #98, opened 2021-11-28
FEATURE: Webserver: Current Effect Status

xcalibur839:
## Description

This feature was designed to address parts of [Issue #90](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/90). The goal was to improve the visual feedback on the web page in a number of ways:
- The currently running effect will automatically appear selected (without triggering the selected event)
- A new status bar is included below the list of effects:
  - The left side displays the name of the currently running effect and a progress bar showing the amount of time remaining for that effect
  - The right side shows a count of currently enabled effects
- This information is updated on the web page every 2 seconds by polling the /getEffectList endpoint in RequestEffectList()

## Contributing requirements

### DIFF
--- a/data/index.html
+++ b/data/index.html
@@ -206,13 +206,30 @@
                 dataType: "text",
                 success: function (responseData, textStatus, jqXHR) {
                     var obj = JSON.parse(responseData);
-                    UpdateEffectList(obj);
+                    UpdateAll(obj);
                 },
                 error: function (responseData, textStatus, errorThrown) {
-                    UpdateEffectList(null);
+                    UpdateAll(null);
                     setTimeout(RequestEffectList, 2000);
                 }
             });
+        }
+
+        // UpdateAll
+        //
+        // Update the status on the page based on the state received from the server
+
+        var initialized = false;
+        function UpdateAll(obj)
+        {
+            if (!initialized)
+            {
+                UpdateEffectList(obj);
+                initialized = true;
+            }
+
+            UpdateEnabled(obj);
+            UpdateCurrentEffect(obj);
         }
     </script>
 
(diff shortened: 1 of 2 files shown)


data/index.html
```diff
@@ -162,14 +206,31 @@ <h2 class="demoHeaders">Dave's Fire NightDriverStrip</h2>
                 dataType: "text",
                 success: function (responseData, textStatus, jqXHR) {
                     var obj = JSON.parse(responseData);
-                    UpdateEffectList(obj);
+                    UpdateAll(obj);
                 },
                 error: function (responseData, textStatus, errorThrown) {
-                    UpdateEffectList(null);
+                    UpdateAll(null);
                     setTimeout(RequestEffectList, 2000);
```
rbergen, line 213: I think this `setTimeout` can be removed. The function will be called again after 2 seconds anyway due to the `setInterval` below.

xcalibur839: Good catch! Yes you're absolutely right, I'll remove that now!

data/index.html
```diff
@@ -192,13 +253,59 @@ <h2 class="demoHeaders">Dave's Fire NightDriverStrip</h2>
...
+            if (effectData.enabledCount > 0)
+            {
+                $('#currentEffectProgress').css('background-color', 'green');
+            }
+            else
+            {
+                $('#currentEffectProgress').css('background-color', 'red');
+            }
+
+            $.each($('#effectList li input'), function(i, v)
```
xcalibur839, line 285: This section was intended to sync the checked state of all of the check boxes, so that if updates are made on another computer they will still be accurately represented on the web page after the next update. 

This seems to be causing odd behavior from my mobile browser, where checking the checkbox appears to be delayed until the next update happens. Sometimes the checked state will jump back and forth a couple of times before settling. This is also iterating over the list of all effects on the page, which could be time intensive if there is a large number of effects.

I'm not sure if there is a better alternative to accomplish this, or if it should be removed.

rbergen: I'm not sure I can really help you with making a decision about this. I can't test it myself; I don't have the hardware, nor the mobile browser in question. That said, on the surface it sounds like something that might be caused by some caching-related oddity, which may be specific to the browser that exhibits the behavior you describe.

PR #99, opened 2021-12-04
Revise secrets.h handling

BeldrothTheGold:
## Description
I was mistaken that git could ignore changes to a tracked file.  It is possible to tell git to assume that a tracked file is unchanged in the index, but this is cumbersome, and causes issues when switching branches. 

To make handling this simpler, I've renamed the secrets.h file to secrets.example.h in the repo. 

I've updated the documentation to inform users that they need to copy the include/secrets.example.h file to include/secrets.h in the build instructions. Then users can edit secrets.h and git will actually ignore secrets.h in users local repo. 

I've also updated the CI to perform this step prior to build. I tested this in my own fork, and it works. 

## Contributing requirements


README.md
```diff
@@ -21,18 +21,21 @@ The simplest configuation, `DEMO`, assumes you have a single meter strip of 144
...
 - Enjoy the pretty lights
-- Start enabling features in the `globals.h` file like WiFi and WebServer.  Set your WiFi SSID and password in secrets.h.
+- Start enabling features in the `globals.h` file like WiFi and WebServer.
 - Connect to the ESP32's web user interface with a browser to its IP address
 
 ## Wifi Setup
-Set your WiFi SSID and password in include/secrets.h.
- - You can prevent git from tracking changes in this file by telling it to assume the file is unchanged. <br/>
-   `git update-index --assume-unchanged include/secrets.h`
+Ensure your WiFi SSID and password are set in include/secrets.h.
```
rbergen, line 34: I think we should add something after the "Ensure your WiFi SSID...", line, like: Please do make sure you set them in `include/secrets.h`, **not** `include/secrets.example.h`.

I'd really like to prevent someone not paying enough attention at exactly that moment, and later accidentally pushing their network details in a PR.

Similarly, I think we should include a comment in (now) include/secrets.example.h to instruct people to copy the file and only THEN add their credentials, just above the defines. Something like:
```cpp
// NOTE: do NOT enter your network details in this file (secrets.example.h)! Instead, copy this file to secrets.h, and set the below defines in that file.`
```

BeldrothTheGold: I've made the requested changes.

rbergen: Thank you. Also for putting this PR together, I think this is a much more robust approach, overall.

BeldrothTheGold: No problem, I wasn't really happy with how this was working. 

Also, on a side note.  I will submit a PR here shortly to significantly speed up the CI. (It will pre-download the base project dependencies to take advantage of the lib folder)

rbergen: Great, looking forward to it!

rbergen:
Looks good to me. Thanks again for contributing this!

PR #128, opened 2022-03-19
Enable response packet over sockets to give server our current status

davepl:
## Description

Adds a "ResponsePacket" that is sent back to the host every time we receive a data frame.  This packet includes information like our firmware version, current FPS, buffer size, buffer pos, and so on.

## Contributing requirements

* [Xttp://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [ X] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [ X] I selected `main` as the target branch.
* [X ] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

### DIFF
--- a/include/ntptimeclient.h
+++ b/include/ntptimeclient.h
@@ -39,7 +39,8 @@
 #include <mutex>
 
 // define the NTP server to connect too (replace . [dots] in IP addresses with , [commas])
-#define cszNTPServer  94, 199, 173, 123     // 0.pool.ntp.org
+#define cszNTPServer  192, 168, 1, 2
+//#define cszNTPServer  94, 199, 173, 123     // 0.pool.ntp.org
 //#define cszNTPServer 216, 239, 35, 12     // google time
 //#define cszNTPServer 17, 253, 16, 253     // apple time
 
@@ -139,6 +140,14 @@ class NTPTimeClient
 
 		uint32_t microsecs = ((uint64_t) frac * 1000000) >> 32;
 
+		// BUGBUG (davepl): I've been gettin back odd packets where the clock is year 2036 and micros is 0. I ignore those.
+
+		if (microsecs == 0)
+		{
+			debugW("Bogus NTP time received, ignoring.");
+			return false;
+		}
+
 		debugV("NTP clock: Raw values sec=%u, usec=%u", frac, microsecs);
 		
 		tvNew.tv_sec = ((unsigned long)chNtpPacket[40] << 24) +       // bits 24 through 31 of ntp time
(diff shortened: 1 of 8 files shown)


include/ntptimeclient.h
```diff
@@ -39,7 +39,8 @@
 #include <mutex>
 
 // define the NTP server to connect too (replace . [dots] in IP addresses with , [commas])
-#define cszNTPServer  94, 199, 173, 123     // 0.pool.ntp.org
+#define cszNTPServer  192, 168, 1, 2
+//#define cszNTPServer  94, 199, 173, 123     // 0.pool.ntp.org
```
rbergen, line 43: This will make the NTP setup tailored to your specific local environment. Does it make sense to make that the premise for the public (version of the) repository?

src/network.cpp
```diff
@@ -328,6 +328,21 @@ bool ProcessIncomingData(uint8_t *payloadData, size_t payloadLength)
...
+            if (delta > 1000)
+            {
+                /*
+                debugI("Server is wildly different so adjusting time by %lf to %lf", delta, dServer);
+                auto tv = AppTime::TimevalFromTime(dServer);
+                settimeofday(&tv, nullptr);
+                time_t newtime = time(NULL);
+                debugI("New Time: %s", ctime(&newtime));
+                */
+            };
```
rbergen, line 344: With all actual logic commented out, does it make sense to check this in?

src/network.cpp
```diff
@@ -415,26 +430,16 @@ bool ProcessIncomingData(uint8_t *payloadData, size_t payloadLength)
...
-                debugI("Server clock: Creeping time back by %lf to to %lf, as clock was ahead by %lf\n", ClockCreepBackLimit, dOld - ClockCreepBackLimit, dNew-dOld);
-                timeval tv = AppTime::TimevalFromTime(dOld - ClockCreepBackLimit);
-                settimeofday(&tv, nullptr);
-            }
-            else
-            {                                                   // Clock received and it's old but not old enough to matter, so its a discard
-                debugV("Server clock: Updated clock TOO OLD, NOT written to ESP32 rtc: %ld.%ld, DELTA: %lf", tvNew.tv_sec, tvNew.tv_usec, dOld - dNew );
-            }
-            
+            */
```
rbergen, line 442: Same question as before.

rbergen (approved):

PR #130, opened 2022-04-09
Implement HueFireFanEffect class. Removed GreenFireFanEffect and BlueFireFanEffect classes.

hemi426:
## Description
Implements HueFireFanEffect, a replacement for existing BlueFireFanEffect and GreenFireFanEffect that allows for easy access to many more color options.

Removed BlueFireFanEffect and GreenFireFanEffect.

Fixed up effects.cpp to reflect changes. Didn't add any new colors as examples, but usage is clear from existing, re-worked effect entries.

## Contributing requirements

### DIFF
--- a/include/effects/faneffects.h
+++ b/include/effects/faneffects.h
@@ -779,6 +779,7 @@ class FireFanEffect : public LEDStripEffect
     bool    bReversed;          // If reversed we draw from 0 outwards
     bool    bMirrored;          // If mirrored we split and duplicate the drawing
     bool    bMulticolor;        // If each arm of the atomlight should have its own color
+    int     _hue;               // Hue "color" from FastLED, int or HSVHue color name
 
     PixelOrder Order;
 
@@ -807,7 +808,8 @@ class FireFanEffect : public LEDStripEffect
                   PixelOrder order = Sequential, 
                   bool breversed = false, 
                   bool bmirrored = false, 
-                  bool bmulticolor = false)
+                  bool bmulticolor = false,
+                  int hue = HUE_RED)
         : LEDStripEffect("FireFanEffect"),
           LEDCount(ledCount),
           CellsPerLED(cellsPerLED),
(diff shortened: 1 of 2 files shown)


hemi426:
Maybe not more blinken, but definitely more colors, and _reduced_ lines of code! Does that qualify as infinite blinken? :D

include/effects/faneffects.h
```diff
@@ -779,6 +779,7 @@ class FireFanEffect : public LEDStripEffect
     bool    bReversed;          // If reversed we draw from 0 outwards
     bool    bMirrored;          // If mirrored we split and duplicate the drawing
     bool    bMulticolor;        // If each arm of the atomlight should have its own color
+    int     _hue;               // Hue "color" from FastLED, int or HSVHue color name
```
rbergen, line 782: Please apply the naming convention of the class you're adding to/deriving from. Personally, I think that would translate to `iHue` in this case.

rbergen (commented):
I think that with a reduction in code, we can consider the post-change blinken-per-bits higher than the pre-change one. That's enough for me when it comes to that. :)

davepl:
Btw, like the idea of starting with a hue number, good one!  I’d agree iHue makes the most sense.  Or _iInitalHue if its never updated.

_iHue would be even better but only if ALL the members get renamed to underscore!

- Dave

include/effects/faneffects.h
hemi426: Absolutely...iHue is a better idea, especially since I've got a few other things in mind. I made the mistake of using _hue for testing the idea locally, then never fixed it when I submitted the PR.

hemi426:
Will eventually investigate a similar change for the FireEffect class. I have a lot to learn before attempting that one, though.

Thank you for a great bit of software and the community that's sprung up around it.

rbergen:
Thanks for this improvement!

(Just to avoid any confusion: I don't mean to claim any credits for your "thank you". That one is solidly @davepl's to take.)

PR #133, opened 2022-04-29
Update platformio.ini

damopinn:
Pin platform to espressif32 @ 3.5.0 to fix compile errors
## Contributing requirements

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -51,7 +51,7 @@ lib_deps        = ${common.lib_deps}
                   ${m5stick-c-plus.lib_deps}
 
 [env]
-platform        = espressif32
+platform        = espressif32@3.5.0
 framework       = arduino
 build_type      = debug
 build_unflags   = -std=gnu++11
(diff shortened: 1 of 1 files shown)


platformio.ini
```diff
@@ -121,7 +121,7 @@ build_flags   = -DLEDSTRIP=1
 ; so it sends power and ground as well as data on PIN 32.D
 
 [env:m5demo]
-platform      = espressif32wn
+platform      = espressif32@3.5.0
```
rbergen, line 124: I noticed that the original platform identifier here is espressif32**wn**, instead of espressif32. Have you established that the "non-wn" version works in this context?

damopinn: I get the error, and the same error on windows
Platform Manager: Installing espressif32wn
Error: Could not find the package with 'espressif32wn' requirements for your system 'linux_x86_64'
I googled this package and the only result is this one, so it seemed to be a typo. 
Another change is needed to compile a build flag is missing from env:m5demo
	-DM5STICKC=1
I'll create another PR or edit this one when I work out what to do.

rbergen: Ok, thank you for the explanation. I'll proceed with merging this PR as it is now, because as it stands the 4.x version of espressif32 breaks builds that were previously working.

rbergen (approved):

PR #146, opened 2022-08-26
Added an additional check to pass the assertion later

SpookedByRoaches:
The pixel picker loop tries to find a black pixel, but there is an assertion later that the found pixel has to not be in the litPixel deque. However, there is no check or anything if the newly picked iNew pixel is actually in the deque. I've added a few lines to delete the chosen pixel from the deque if it is chosen.

Thanks to @steve-git-hub for his help in [this](https://github.com/PlummersSoftwareLLC/NightDriverStrip/discussions/69) discussion.

## Contributing requirements

### DIFF
--- a/include/effects/misceffects.h
+++ b/include/effects/misceffects.h
@@ -295,6 +295,9 @@ class TwinkleEffect : public LEDStripEffect
 				size_t i = random(0, NUM_LEDS);
 				if (pPixels[i] != CRGB(0,0,0))
 					continue;
+				std::deque<size_t>::iterator found = find(litPixels.begin(), litPixels.end(), i);
+				if (litPixels.end() != found)
+					litPixels.erase(found);
 				iNew = i;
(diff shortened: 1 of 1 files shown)


rbergen:
@SpookedByRoaches Thanks for submitting this. 

This PR was not merged before, because @davepl had already made significant progress on a large development of the codebase. That has now been merged through #150. This does mean that this PR would need to be updated, as `/include/effects/misceffects.h` has moved to `/include/effects/strip/misceffects.h`.

Concerning the proposed change itself, personally I think it would be better to not remove the `iNew` pixel from the deque if it is already in it. Instead, I think it is better to leave the pixel on the deque where it is, and continue the search for a different black pixel, instead.

rbergen:
Closing this after the merge of #151, @SpookedByRoaches again, many thanks for flagging the bug and proposing a solution.

PR #150, opened 2022-11-01
Merge all Davepl changes into main

davepl:
## Description

Updates strips to firmware v032
## Contributing requirements

* [X] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [X] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [X] I selected `main` as the target branch.
* [X] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #151, opened 2022-11-10
TwinkleEffect: Check selected pixel isn't already lit

rbergen:
## Description

This is an alternative solution for the issue addressed in #146. It has also been written to align with the current state of the codebase (after the merge of #150).

## Contributing requirements


davepl (approved):
As I read it, will not use pixels that are not black, or already in the list.

PR #170, opened 2022-12-06
Interval

davepl:
## Description

Makes spectrum effects able to have a variable number of bands that may or may not match the actual sampler

## Contributing requirements


rbergen:
I see there's a number of new envs in platformio.ini. Do you think we should include one or more of them in CI?

davepl:
It’s taking a long time to build already, and I think we’re already covering the major variants, but perhaps adding the panlee one would be good?  That’s the only new “complicated” one I’d say!

Thanks
Dave

rbergen:
You're welcome!

The best way to go about it is to sync a branch you want to merge into
PlummersSoftwareLLC:main just before you open the PR. You can tell if there
is anything to sync by looking at the branch in GitHub. If it says the
"branch is x commits behind...", then you can sync your branch by hitting
the "Sync fork" button.

[image: image.png]

The longer you've been working on a branch independently from the main
project, the bigger the chance that things have diverted to a point it may
be problematic. An added benefit of this is that this will also identify
any merge conflicts you will otherwise run into when opening your PR. Of
course, after syncing the branch on GitHub, don't forget to "git pull" to
local.

Thanks for explaining the changes you made, there's some good stuff in
there! I could already tell by the PR that you've done quite a bit of work
again; the changed file count was 19 for starters. :)

Cheers,
Rutger

davepl:
What about in my case where it’s a fork, not a branch?  Like I’m working in

Davepl/nightdriverstrip

But others are working away in 

Plummerssosftawarellc/nightdriverstrip

“Back in my day” we would sync down changes, make them build with OUR changes, and then submit as a checkin.   But how do I pull changes from plummerssoftwarellc into my fork?

If I had to guess, I’d open a PR to pull Plummers into Davepl, then git pull to get those changes, but it would then ask me to stash, right?  So I have no idea!

 -Dave

PR #179, opened 2022-12-18
Add a fix for POST hasParam and getParam.

MatthewvanBuuren:
## Description
- hasParam and getParam do not require indicating POST=true in second parameter which actually causes that request to fail when a valid POST request has been sent with that param set. 
- - Validated locally with REST client and validated the correct behaviour occurs now.
-- FYI hasParam(const String& name, bool post, bool file) from WebRequest.cpp

## Contributing requirements

### DIFF
--- a/include/spiffswebserver.h
+++ b/include/spiffswebserver.h
@@ -180,11 +180,11 @@ class CSPIFFSWebServer
 
         // Look for the parameter by name
         const char * pszEffectInterval = "effectInterval";
-        if (pRequest->hasParam(pszEffectInterval, true, false))
+        if (pRequest->hasParam(pszEffectInterval))
         {
             debugI("found EffectInterval");
             // If found, parse it and pass it off to the EffectManager, who will validate it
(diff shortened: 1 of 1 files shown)


rbergen:
Hi @MatthewvanBuuren, I'm not sure I fully understand your definition of "correct". That is to say that if the intended behaviour of the endpoints in question is that they only accept the parameters when they are passed in an HTTP POST, then the current code is actually correct. What makes you conclude this is not the case?

MatthewvanBuuren:
Hi Rutger, I should double check this but I experienced that POST requests to disableEffect for example were received by the right request handler but inside the DisableEffect method where it validated hasParam, It would fail.. or rather it wouldn't get into that block even when it did have that parameter set. The official guide for that method suggests the current implementation is correct but for some reason it doesn't identify the parameter exists correctly. Since DisableEffect is already filtering for POST, Skipping defining the POSt Param on the hasParam fixed the issue for me.

PR #210, opened 2023-01-31
Build all environments/configs in CI

rbergen:
This makes GitHub CI (Actions) build all configurations included in platformio.ini.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,28 +33,9 @@ jobs:
         pio pkg update --no-save
     - name: Copy secrets
       run: cp include/secrets.example.h include/secrets.h
-    - name: Build chieftain
-      run: pio run -e chieftain
-    - name: Build demo
-      run: pio run -e demo
-    - name: Build heltecdemo
-      run: pio run -e heltecdemo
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -33,28 +33,9 @@ jobs:
...
-    - name: Build spectrum
-      run: pio run -e spectrum
-    - name: Build umbrella
-      run: pio run -e umbrella
-    - name: Build xmastrees
-      run: pio run -e xmastrees
+    - name: Enable all environments
+      run: |
+        cp platformio.ini platformio.ini.orig
+        grep -v ^default_envs platformio.ini.orig > platformio.ini
```
davepl, line 39: Exactly how I would have done it :-)

davepl: And small minds seldom differ, they say!  But yours is superior today because I have no idea what the pipe does, unless it's just the line continuation character?  And you can place one command on a line?

rbergen: The single pipe means "multiple commands will follow, on the next level of indentation". It can work like that because, like Python, YAML is also indentation-sensitive. Without the pipe, there can indeed be one command line immediately following the `run:`.

PR #217, opened 2023-02-12
Remove data directory from committed tree

rbergen:
The data/ directory contains files that need to be built per project, and are generated by the bake_site.py script. As such, the output the files in data/ represent should not be part of the repo codebase.


davepl:
I’ve removed these files in my own checkin in my fork, changed the partition table, and so on… it’ll likely be a merge.  Is anything going in their place for now?  Or anything else we should coordinate on?  

Cheers,
Dave

rbergen:
Nothing else will go in their place in the context of the repo (i.e., the checked-in codebase). The contents of the data directory are generated during every PlatformIO environment build that needs them, but that's a strictly "build-local" affair.

PR #223, opened 2023-02-14
ImprovSerial work - add SSID enumeration via RPC; templatized ImprovSerial so it can run on S2 as well

davepl:
## Description
Adds the RPC support needed for the web installer to pull the SSID list from the local esp32 device so it can provide those choices in a list.

## Contributing requirements


WebInstaller/index.html
```diff
@@ -49,10 +45,6 @@
         ></script>
```
rbergen, line 45: The PR introduces an esp32webtools directory with what seems like a "local copy" of ESP32 Web Tools. However, the HTML still retrieves the code it uses from unpkg.com. Unless I'm missing something, I'd say that either the script block here should use the local copy, or the local copy (i.e. esp32webtools directory with contents) can be removed?

davepl:
I’m not an HTML guy and couldn’t get it to work with the local copies of the files. The only reason I need/want a local copy is because it’s hardcoded to upload at 115200, which takes minutes.  If I could get the local copy to work, I could run it at 1500000, much much faster!

Long term we need to host it somewhere anyway, rather than rely in their package.  Just the small matter of getting it working; if I can’t, perhaps you or Louis will have an idea on why…

So for now there’s an unused local copy, but it’s working its way towards working…

- Dave

rbergen:
I'm not the biggest HTML expert either, but my guess is that the problem is with the local copy not being set up to be used as a module. That means it won't work if the &lt;script&gt; tag includes the type="module" attribute. Anyway, I can test that after the changes I'm working on have been merged as well. Until then, I now understand the apparent discrepancy and I'm fine with things as they are.

rbergen (approved):

davepl:
I tried again today and it works, so not sure what I did yesterday!

In my working copy, here’s my folder layout, hope it makes sense and matches what you’re doing!

The content of esp32webtools is the output from the web folder of our fork of esp32webtools.  Right now I’m manually copying the output files and checking them in, but one day they should be generated directly from the code, whether we include the web tools into the main project or keep it a separate repo.

But for now, to make this hierarchy here’s my manual steps:

1) Copy the firmware.bin and friends from the build output to the firmware/project folder
2) Copy the web folder from an esp32webtools build into the esp32webtools folder

Let me know if any of this conflicts with what you’re doing or needs to be different!

Thanks,
Dave

PS: If I knew how, what I’d like to do is have the esp32webtools code live in NightDriverStrip as a subfolder and get built/binplaced by the extra_scripts.py but still be linked to the original repo somehow so that we can update it when they do without manual copying.  I’m not sure about repos under repos, so haven’t tried that yet...


￼

PR #228, opened 2023-02-26
Rearrange repo files, fix issues with web installer WiFi setup, other small changes

rbergen:
This:

- Moves a number of existing files in the repo to directories that have recently been added to the tree.
- Fixes issues with setting up WiFi via the web installer
- Includes the NightDriverStrip version in the web installer's image footprint


rbergen:
Merging this based on positive test results on staging.

PR #230, opened 2023-02-28
Overhaul README.md

vaaski:
## Description

I added some things to the readme that I wished were there in the first place. This includes:

- updated build command
- link to the config section where applicable
- explain build_flags format
- build time on my machine ;)

I also auto-formatted the tables and such for better readability.
related: #229 

## Contributing requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 <img src="assets/NightDriverLogo-small.png" width="400" />
 
 <!-- markdownlint-disable MD036 /no-emphasis-as-heading -->
+
 _Davepl, 9/19/2021_
 
 - See [Discussions](http://github.com/PlummersSoftwareLLC/NightDriverStrip/discussions) for questions and comments.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,44 +127,46 @@ The project can be built using [PlatformIO](https://platformio.org/). There's a
...
+| Hardware Specific | Description                                         | Supported Boards             |
+| ----------------- | --------------------------------------------------- | ---------------------------- |
+| USE_M5DISPLAY     | Enable stats display on built in LCD                | M5Stick-C and M5Stick-C Plus |
+| USE_OLED          | Enable stats display on built in OLED               | Heltec Wifi Kit 32           |
+| USE_LCD           | Enable stats display on external ILI9341 LCD        | Wrover32                     |
+| USE_TFTSPI        | Enable stats display on external TTGO LCD           | esp32dev                     |
+| ENABLE_AUDIO      | Listen for audio from the microphone and process it | M5Stick-C and M5Stick-C Plus |
+| ENABLE_REMOTE     | IR Remote Control                                   | Requires IR Hardware         |
+
+example in platformio.ini (prefix the flags with `D`, e.g. `ENABLE_WIFI=1` becomes `DENABLE_WIFI=1`)
```
rbergen, line 166: If you want to mention the `D`, then please also include the dash (`-`). `-D` is the build flag to set a #define from the command line.

README.md
```diff
@@ -6,34 +6,35 @@
 <img src="assets/NightDriverLogo-small.png" width="400" />
 
 <!-- markdownlint-disable MD036 /no-emphasis-as-heading -->
+
```
rbergen, line 9: What may actually be best here is to change `markdownlint-disable` to `markdownlint-disable-next-line` and then keep it together with the line it is actually meant to address. That does mean the empty line you added will need to be removed again.

rbergen (commented):
Thank you for these changes, I think that on the whole they're an improvement. I do have two small comments; could you please look at them?

README.md
vaaski: Can't really comment on this since I don't use markdownlint, but sure, makes sense.

rbergen (approved):

PR #237, opened 2023-03-10
added Lilygo T-Display-S3 config

mbenna:
## Description
Lilygo T-Display-S3 config was added to platform.io to make it easy to compile for this platform.

## Contributing requirements

* [*] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [*] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [*] I selected `main` as the target branch.
* [*] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -9,7 +9,7 @@
 ; https://docs.platformio.org/page/projectconf.html
 
 [platformio]
-default_envs = 
+default_envs = lilygo-tdisplay-s3-demo
 
 ; ==================
 ; Base configuration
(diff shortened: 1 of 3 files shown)


platformio.ini
```diff
@@ -133,6 +133,14 @@ board           = esp-wrover-kit
 monitor_speed   = 115200
 upload_speed    = 1500000
 
+[dev_lilygo-tdisplay-s3]
+extends         = dev_esp32-s3
+build_flags     = -DLILYGOTDISPLAYS3=1
+                  -DPIN_SDA=16
+                  -DTOGGLE_BUTTON_1=0
+                  -DTOGGLE_BUTTON_2=14
+                  ${base.build_flags}
```
rbergen, line 142: Please build on top of the build flags of the section (device) you're extending:
 
```suggestion
                  ${dev_esp32-s3.build_flags}
```

That way, build flags can be added to the base section/device without then having to consider sections extending from it. (The fact dev_esp32-s3 currently sets no `build-flags` of its own is irrelevant; it will automatically inherit those from `base`).

mbenna: Good catch; I agree. Done.

I also changed the default data pin on this device because once WiFi is enabled the previous pin becomes unstable and results in sporadic led flashing.

rbergen (approved):
LGTM

PR #251, opened 2023-04-09
(De)serialize EffectManager to/from JSON

rbergen:
## Description

This adds the ability to serialize the EffectManager config (i.e. effects list and enabled state) to JSON, and deserialize EffectManager from said JSON upon construction. 

This implementation stores the generated JSON on the SPIFFS partition. That means that the config is cleared when a SPIFFS image without the JSON config is uploaded. At the same time, it also allows effects configuration to be included in such an image. 

Do note that at this time, it's an "all or nothing" deal: if a JSON file with effect manager config is present, then only the effects defined in it will be instantiated. A more hybrid approach is foreseen in a future update. At this time, the serialization capability is mainly an internal affair anyway, as NightDriverStrip does not yet provide means to change the configuration of effects at runtime aside from enabling or disabling them.

## Contributing requirements


include/effects.h
```diff
@@ -0,0 +1,151 @@
...
+#define EFFECT_STAR_MUSIC_PULSE                          6
+#define EFFECT_STAR_QUIET                                7
+#define EFFECT_STAR_BUBBLY                               8
+#define EFFECT_STAR_FLASH                                9
+#define EFFECT_STAR_COLOR_CYCLE                         10
+#define EFFECT_STAR_MULTI_COLOR                         11
+#define EFFECT_STAR_CHRISTMAS                           12
+#define EFFECT_STAR_HOT_WHITE                           13
+
+// Some common JSON properties to prevent typos
```
davepl, line 124: Is there real value in keep these to 3 chars, or would longer more descriptive names make the JSON easier to read?

include/effects/matrix/PatternWeather.h
```diff
@@ -250,8 +253,35 @@ class PatternWeather : public LEDStripEffect
...
+        if (strPostalCode.isEmpty())
+            strPostalCode = cszZipCode;
+
+        if (strCountryCode.isEmpty())
+            strCountryCode = cszCountryCode;
+    }
+
+    virtual bool SerializeToJSON(JsonObject& jsonObject) 
+    {
+        StaticJsonDocument<128> jsonDoc;
```
davepl, line 276: Will 128 always be large enough to serialize it here?

include/effects/matrix/spectrumeffects.h
```diff
@@ -329,6 +349,36 @@ class SpectrumAnalyzerEffect : public LEDStripEffect, virtual public VUMeterEffe
...
 
+    SpectrumAnalyzerEffect(const JsonObjectConst& jsonObject)
+        : LEDStripEffect(jsonObject), 
+          _numBars(jsonObject["nmb"]),
+          _colorOffset(0),
+          _scrollSpeed(jsonObject[PTY_SPEED]), 
+          _fadeRate(jsonObject["frt"]),
+          _palette(jsonObject[PTY_PALETTE].as<CRGBPalette16>()),
+          _peak1DecayRate(jsonObject["pd1"]),
+          _peak2DecayRate(jsonObject["pd2"])
```
davepl, line 360: Cool, I like how it can use the incoming object to init all of its params...

include/effects/strip/faneffects.h
```diff
@@ -936,6 +1005,45 @@ class FireFanEffect : public LEDStripEffect
     abHeat = std::make_unique<uint8_t[]>(CellCount());
   }
 
+  FireFanEffect(const JsonObjectConst& jsonObject)
```
davepl, line 1008: Is there any validation that the JSON object contains all the params, or what if it's missing a few, and they don't get initialized?

include/effects/strip/fireeffect.h
```diff
@@ -412,6 +511,38 @@ class SmoothFireEffect : public LEDStripEffect
...
+    }
+
+    virtual bool SerializeToJSON(JsonObject& jsonObject) 
+    {
+        StaticJsonDocument<128> jsonDoc;
+        
+        JsonObject root = jsonDoc.to<JsonObject>();
+        LEDStripEffect::SerializeToJSON(root);
+
+        jsonDoc[PTY_MIRORRED] = _Reversed;
```
davepl, line 534: Why the mix of macro and const json property ids?  Are some well known and hence used across multiple effects, while the 'hardcoded' ones are one-offs?

include/jsonserializer.h
```diff
@@ -0,0 +1,102 @@
...
+//              
+//---------------------------------------------------------------------------
+
+#pragma once
+
+#include <ArduinoJson.h>
+#include "jsonbase.h"
+#include "FastLED.h"
+
+struct IJSONSerializable
```
davepl, line 38: Using a struct means the functions will be public, just confirming that's what you want here!

include/effects.h
rbergen: Three premises for my implementation of this are:
1. Prioritizing minimum use of memory (and storage)
2. Each serializable class decides for itself how it serializes itself into a JSON object
3. The JSON that's saved is really only for "internal use" by the program logic itself, so it does not have to be that human-friendly.

The first premise calls for short property names. The second means that each serializable class only needs to make sense of the JSON property names it uses for its own JSON objects. As the number of properties per class is pretty small, I think the shorter property names are not too confusing and thus defendable.

Of course, this does change if you want to optimize the JSON for consumption by humans, thereby effectively dropping premise 3. 

It's a choice, and we can choose either way. I guess it depends on what use (and composition) of the JSON you envisage. :)

include/effects/matrix/PatternWeather.h
rbergen: That's a fair question. What needs to be serialized is:

* Two property names of 2 characters each (6 bytes)
* Two property names of 3 characters each (8 bytes)
* One integer (8 bytes)
* "Weather" (8 bytes)
* One country code of 2 characters (3 bytes)
* The postal code or other location indicator

The knowns add up to 33 bytes in raw data, leaving 95 bytes for postal code/location and bytes for structure. I gathered that should be enough.

Now that you asked the question I've used the ArduinoJson Assistant, and it calculated that 82 bytes are needed for the serialization with my location data (postal code "4133EK", country code "nl"). That leaves 46 more bytes (&asymp; characters) for a different location specifier. I still think that's enough. :)

include/effects/matrix/spectrumeffects.h
rbergen: I do too. :)

include/effects/strip/fireeffect.h
rbergen: I agree that this is somewhat inconsistent. But yes, my consideration was exactly that.

include/effects/strip/faneffects.h
rbergen: There is no such validation at this time. In line with what I wrote earlier ("a serializable class is responsible for its own serialization"), only the serializable class can figure out what to do with invalid JSON. My reasoning was as follows:

- A serializable class (effect) can decide at deserialization time (construction) how to deal with missing/invalid properties.
- If an effect wants to prevent construction in case of certain scenario, its factory (i.e. its respective lambda in effectfactories,cpp) can perform the appropriate validation and decide to return nullptr instead of a new effect.

The reason I didn't implement any actual validation logic for individual effects is that it requires more knowledge of the logic of the effects than I currently possess.

include/jsonserializer.h
rbergen: Yes, it is. This is the "serializable" interface definition that declares one method, which must be public to be useful. As it is an interface definition, I expect any future additions to have to be public too.

PR #253, opened 2023-04-20
Effect improvements, device configuration and related changes

rbergen:
## Description

This adds:

- a number of effect improvements implemented by @davepl
- a facility to keep and persist a collection of centralized settings, which can be retrieved (HTTP GET) and modified (HTTP POST) using the existing /settings endpoint
- a /reset endpoint, which can be used to trigger a reset using HTTP POST of one or more of the following: effect config as added through #251 and #252, device config as described in the previous point, the board itself (i.e. reboot)

A number of effects have been modified to use the device settings that have now been implemented.

## Contributing requirements


include/effects/matrix/PatternWeather.h
```diff
@@ -71,9 +72,11 @@ class PatternWeather : public LEDStripEffect
 
 private:
 
+    String strLocationName       = "";
     String strLocation           = "";
-    String strPostalCode         = "";
     String strCountryCode        = "";
+    String strLatitude           = "0.0";
```
davepl, line 78: Polar!

include/effects/matrix/PatternWeather.h
```diff
@@ -104,30 +110,70 @@ class PatternWeather : public LEDStripEffect
         return (K - 273.15) * 9.0f/5.0f + 32;
     }
 
-    inline float KelvinToCelcius(float K)
+    inline float KelvinToCelsius(float K)
```
davepl, line 113: You'd think, being Canadian, I could have spelled it more goodly.

include/effects/matrix/PatternWeather.h
```diff
@@ -247,41 +296,27 @@ class PatternWeather : public LEDStripEffect
         }
         auto task = pObj->weatherTask;
         pObj->weatherTask = nullptr;
+        pObj->updateInProgress = false;
         debugW("Weather thread exiting");
         vTaskDelete(task);
     }
 
+    bool HasLocationChanged()
```
davepl, line 304: What if LAT/LONG changes, does that matter?

src/webserver.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    auto response = new AsyncJsonResponse(false, JSON_BUFFER_BASE_SIZE);
+    response->addHeader("Server","NightDriverStrip");
+    auto root = response->getRoot();
+    JsonObject jsonObject = root.to<JsonObject>();
+    
+    g_aptrDeviceConfig->SerializeToJSON(jsonObject);
+    jsonObject["effectInterval"] = g_aptrEffectManager->GetInterval();
+
+    response->setLength();
```
davepl, line 91: Doesn't AddCORESHeaderAndSendOKResponse do this same workflow?

davepl (approved):
Looks good!  A few very minor comments but nothing material, except maybe lat/long change not changing location?

include/effects/matrix/PatternWeather.h
rbergen: :) I missed it too in my review of the initial conversion code.

include/effects/matrix/PatternWeather.h
rbergen: LAT/LONG is only set by UpdateCoordinates(), which retrieves them from OpenWeatherAPI's geocoding endpoint using Location and Country. This function checks if the Location and Country used to retrieve the instances LAT/LONG are (still) in sync with Location and Country as set in the device settings.

src/webserver.cpp
rbergen: Kind of. AddCORESHeaderAndSendOKResponse creates its own response, using an optional text parameter. This method sends back a JSON response that's created earlier. This is the same approach used by other CWebServer endpoint methods that send back content in their response.

That said, I guess overall a few lines of code could be saved if AddCORESHeaderAndSendOKResponse was overloaded to also accept a response object, and that overload was used by all methods that return a content response.

PR #254, opened 2023-04-29
Refactor webserver and device config, change WiFi connect (wait) logic

### LINKED ISSUE #255: Splash screen does not time out
## Bug report

**Problem**
After flashing my board, the splash logo does not time out in a few seconds. If I erase_flash and then reprogram, it fixes it.  But it recurs.

**Steps**
1. Upload firmware, use it 
2.  ???
3.  Upload firmware
4. Note that splash screen doesn't time out. 

I wonder if perhaps the full interval is being stored for it?  There's some special logic that allows an effect to set a max duration, and perhaps that's being ignored?

If there's a way for me to fetch the JSON from SPIFFs, let me know.  If not, consider adding a "/dumpjson" REST call to do just that in order to ease future debugging?

rbergen:
## Description

This:

- Refactors the webserver and device config class to reduce duplication of code
- Modifies the WiFi connect (wait) logic, to allow for connections taking longer to settle
- Shows the splash screen effect at startup, regardless of the effect list loaded from JSON or by default
- Incorporates the changes of #257, and thus also fixes #255
- Made the Weather effect update task persistent, which fixes a memory leak
- Uses PSRAM for JsonDocument instances
- Preloads static device stats in the webserver, preventing freezes when the web interface stats view is enabled

Note: while reviewing this PR, one might want to ignore whitespace changes. In one commit, trailing whitespace was removed for most files in the `include` and `src` directories.

## Contributing requirements

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -2,7 +2,7 @@
 //
 // File:        deviceconfig.h
 //
-// NightDriverStrip - (c) 2018 Plummer's Software LLC.  All Rights Reserved.  
+// NightDriverStrip - (c) 2018 Plummer's Software LLC.  All Rights Reserved.
 //
 // This file is part of the NightDriver software project.
 //
@@ -10,12 +10,12 @@
 //    it under the terms of the GNU General Public License as published by
 //    the Free Software Foundation, either version 3 of the License, or
 //    (at your option) any later version.
-//   
+//
 //    NightDriver is distributed in the hope that it will be useful,
 //    but WITHOUT ANY WARRANTY; without even the implied warranty of
 //    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 //    GNU General Public License for more details.
-//   
+//
 //    You should have received a copy of the GNU General Public License
 //    along with Nightdriver.  It is normally found in copying.txt
 //    If not, see <https://www.gnu.org/licenses/>.
@@ -46,7 +46,7 @@ class DeviceConfig : public IJSONSerializable
     bool use24HourClock;
     bool useCelsius;
 
-/*    
+/*
     void WriteToNVS(const String& name, const String& value);
     void WriteToNVS(const String& name, bool value);
 */
@@ -66,7 +66,7 @@ class DeviceConfig : public IJSONSerializable
     template <typename T>
     void SetIfPresentIn(const JsonObjectConst& jsonObject, T& target, const char *tag)
     {
-        if (jsonObject.containsKey(tag)) 
+        if (jsonObject.containsKey(tag))
             target = jsonObject[tag].as<T>();
     }
 
@@ -93,7 +93,7 @@ class DeviceConfig : public IJSONSerializable
         jsonDoc[TIME_ZONE_TAG] = timeZone;
         jsonDoc[USE_24_HOUR_CLOCK_TAG] = use24HourClock;
         jsonDoc[USE_CELSIUS_TAG] = useCelsius;
-    
+
         return jsonObject.set(jsonDoc.as<JsonObjectConst>());
     }
 
@@ -111,9 +111,9 @@ class DeviceConfig : public IJSONSerializable
         SetIfPresentIn(jsonObject, use24HourClock, USE_24_HOUR_CLOCK_TAG);
         SetIfPresentIn(jsonObject, useCelsius, USE_CELSIUS_TAG);
 
-        if (jsonObject.containsKey(TIME_ZONE_TAG)) 
+        if (jsonObject.containsKey(TIME_ZONE_TAG))
             return SetTimeZone(jsonObject[TIME_ZONE_TAG], true);
-   
+
         if (!skipWrite)
             SaveToJSON();
 
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -72,13 +72,13 @@ class DeviceConfig : public IJSONSerializable
 
   public:
 
-    static constexpr const char * LOCATION_TAG = NAME_OF(location);
-    static constexpr const char * LOCATION_IS_ZIP_TAG = NAME_OF(locationIsZip);
-    static constexpr const char * COUNTRY_CODE_TAG = NAME_OF(countryCode);
-    static constexpr const char * OPEN_WEATHER_API_KEY_TAG = NAME_OF(openWeatherApiKey);
-    static constexpr const char * TIME_ZONE_TAG = NAME_OF(timeZone);
-    static constexpr const char * USE_24_HOUR_CLOCK_TAG = NAME_OF(use24HourClock);
-    static constexpr const char * USE_CELSIUS_TAG = NAME_OF(useCelsius);
+    static constexpr const char * LocationTag = NAME_OF(location);
+    static constexpr const char * LocationIsZipTag = NAME_OF(locationIsZip);
+    static constexpr const char * CountryCodeTag = NAME_OF(countryCode);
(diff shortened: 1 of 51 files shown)


include/effectmanager.h
```diff
@@ -72,25 +73,32 @@ class EffectManager : public IJSONSerializable
...
+
+            // This is a hacky way to ensure that we start the correct effect after the temporary one
+            _iCurrentEffect--;
+        }
     }
 
-    void ClearEffects() 
+    void ClearEffects()
     {
         for (auto effect : _vEffects)
```
davepl, line 103: Watch for a merge conflict here, we no longer need to delete the items individually now that they're in vectors...

include/deviceconfig.h
```diff
@@ -66,25 +66,34 @@ class DeviceConfig : public IJSONSerializable
...
+
+        jsonDoc[LocationTag] = location;
+        jsonDoc[LocationIsZipTag] = locationIsZip;
+        jsonDoc[CountryCodeTag] = countryCode;
+        jsonDoc[OpenWeatherApiKeyTag] = openWeatherApiKey;
+        jsonDoc[TimeZoneTag] = timeZone;
+        jsonDoc[Use24HourClockTag] = use24HourClock;
+        jsonDoc[UseCelsiusTag] = useCelsius;
 
-        jsonDoc[NAME_OF(location)] = location;
```
davepl, line 80: Love the NAME_OF!

include/effectmanager.h
```diff
@@ -310,19 +324,18 @@ class EffectManager : public IJSONSerializable
 
     void StartEffect()
     {
+        // If there's a temporary effect override from the remote control active, we start that, else
+        // we start the current regular effect
+
+        LEDStripEffect *pEffect = _ptrRemoteEffect ? _ptrRemoteEffect.get() : _vEffects[_iCurrentEffect];
```
davepl, line 330: Perhaps this can become shared_ptr after merged so that it holds a reference to the remote effect.  Also, we should really name remoteEffect to "tempEffect" or similar.  I don't think it's naming is very clear that its from the temp remote control effect!

include/effects/matrix/PatternLife.h
```diff
@@ -117,7 +117,7 @@ class PatternLife : public LEDStripEffect
     // Seed: 92465, Generations: 1626
 
     static constexpr long bakedInSeeds[] =
```
davepl, line 119: Are these new seeds you discovered or where's this set of a few come from?

include/effects/matrix/PatternWeather.h
```diff
@@ -114,7 +113,7 @@ class PatternWeather : public LEDStripEffect
     {
         return K - 273.15;
     }
-    
+
```
davepl, line 116: This reads backwards to me?  If UseCelcius is true, shouldn't it do K->C instead of K->F?

include/effects/matrix/PatternWeather.h
```diff
@@ -339,45 +343,45 @@ class PatternWeather : public LEDStripEffect
...
+                    debugW("Spawning thread to check weather...");
+                    xTaskCreatePinnedToCore(WeatherTaskEntryPoint, "Weather", 4096, (void *) this, NET_PRIORITY, &weatherTask, NET_CORE);
                 }
                 else
                 {
-                    debugW("Skipping weather fetch because previous update still in progress");
+                    debugW("Triggering thread to check weather now...");
+                    // Resume the weather task. If it's not suspended then there's an update running and this will do nothing,
+                    //   so we'll silently skip the update this would otherwise trigger.
+                    vTaskResume(weatherTask);
```
davepl, line 363: Clever!

include/webserver.h
```diff
@@ -69,200 +84,73 @@ class CWebServer
...
-
-            j["currentEffect"]         = g_aptrEffectManager->GetCurrentEffectIndex();
-            j["millisecondsRemaining"] = g_aptrEffectManager->GetTimeRemainingForCurrentEffect();
-            j["effectInterval"]        = g_aptrEffectManager->GetInterval();
-            j["enabledCount"]          = g_aptrEffectManager->EnabledCount();
-
-            for (int i = 0; i < g_aptrEffectManager->EffectCount(); i++) 
-            {
-                DynamicJsonDocument effectDoc(256);
-                effectDoc["name"]    = g_aptrEffectManager->EffectsList()[i]->FriendlyName();
```
davepl, line 114: Not a big deal, but I think EffectsList returns the entire array of effects each time we index into it.  Might be worth having it take a parameter of which effect and fetching just the one?

src/network.cpp
```diff
@@ -196,103 +196,99 @@ void IRAM_ATTR RemoteLoopEntry(void *)
...
+            return false;
+        }
 
-                WiFi.disconnect();
-                WiFi.mode(WIFI_STA);
-                WiFi.begin(WiFi_ssid.c_str(), WiFi_password.c_str());
+        for (uint iPass = 0; iPass < cRetries; iPass++)
+        {
+            Serial.printf("Pass %u of %u: Connecting to Wifi SSID: %s - ESP32 Free Memory: %u, PSRAM:%u, PSRAM Free: %u\n",
+                iPass + 1, cRetries, WiFi_ssid, ESP.getFreeHeap(), ESP.getPsramSize(), ESP.getFreePsram());
```
davepl, line 229: Come on, no pass 0?  We're all programmers here :-)

davepl (approved):
Looks good!  Got a lot easier to review when I found the "ignore whitespace" setting!

include/effectmanager.h
rbergen: They have to still be deleted here, because in this version the contents of the vector are still raw pointers. Beyond that you're right, there will be some merge conflicts to address when we merge the safe pointer switch.

include/deviceconfig.h
rbergen: Haha, I knew you would. :)

include/effectmanager.h
rbergen: Changing the naming of the effect did cross my mind, but as I didn't - and don't - know the history of the special one effect, I chose not to touch it. If you also think it should be changed, then I think we can take care of that when we implement the switch to shared_ptr as well.

include/effects/matrix/PatternLife.h
rbergen: I did not touch this code. I think they're coming from a set of commits I pulled in mid-way from your fork?

include/effects/matrix/PatternWeather.h
rbergen: Thank you!

include/webserver.h
rbergen: It depends. The endpoint is currently used by the current web interface, which retrieves and uses the data of the whole list. We could fetch and retain the effects list once before the loop, to reduce the number of EffectsList() invocations.

src/network.cpp
rbergen: You're probably right, but "0 to 4 out of 5" still causes a short circuit in my old-fashioned, simple little brain. :)

include/effectmanager.h
```diff
@@ -130,7 +149,7 @@ class EffectManager : public IJSONSerializable
     {
         ClearEffects();
         _vEffects.reserve(cEffects);
-        
+
         for (int i = 0; i < cEffects; i++)
         {
             _vEffects.push_back(pEffects[i]);
```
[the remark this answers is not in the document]
rbergen: I know. This will be ironed out when we merge the transition to safe pointers that you implemented.

PR #257, opened 2023-04-30
Fix #255

### LINKED ISSUE #255: Splash screen does not time out

rbergen:
## Description

This fixes #255, by:

- Not ignoring an effect's MaximumEffectTime() when the default effect interval is 0
- Persisting and restoring the current effect index

It also adds an /effectsConfig webserver endpoint that returns the contents of the effects config JSON file.

## Contributing requirements

### DIFF
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -470,11 +470,8 @@ public:
 
     uint GetInterval() const
     {
-        // This allows you to return a MinimumEffectTime and your effect won't be shown longer than that
-
-        if (_effectInterval == 0)
-            return std::numeric_limits<uint>::max();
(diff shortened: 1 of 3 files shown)


rbergen:
Closing this in favor of #254.

PR #269, opened 2023-05-16
Fix remote control falter and embed effect JPEGs

### LINKED ISSUE #263: Wrong caption name when stepping through effects with the remote
## Bug report

When stepping rapidly through the effects, the caption name isn't updated correctly and will often show the last effect, not the current one

**Problem**
Should show the current effect

**Steps**
Step through the effects reasonably rapidly with the remote, observe wrong title

rbergen:
## Description

This:

- Restores responsiveness to successive IR remote control effect changes. As such, it fixes #263 
- Embeds JPG files used by effects in the firmware image
- Adds `override` specifier to overrides of LEDStripEffect::Init()

## Contributing requirements


platformio.ini
```diff
@@ -193,6 +193,18 @@ lib_deps        = https://github.com/Xinyuan-LilyGO/TTGO-T-Display
 build_flags     = -DUNITY_INCLUDE_DOUBLE
                   -DUNITY_DOUBLE_PRECISION=1e-12    ; Make doubles real 8 byte doubles
 
+[matrix_config]
```
davepl, line 196: Only catch I can see here is that Weather isn't actually dependent on running on a MATRIX.   With enough pixels, it could run on a WS2812B array as well.  But I don't forsee that happening any time soon.

davepl (approved):
Looks good!

platformio.ini
rbergen: That's a good point. If and when that changes we'll have to restructure this and other things. For one, in the NightDriverStrip source code, the inclusion of all PatternXxxx effects is conditional on USE_MATRIX being set.

davepl:
That’s why I’ve tried to refactor all matrix-specific functionality into LEDMatrixGFX.  Then effects should be written to the GFXBase class and should work on any type of matrix.

USE_MATRIX should really be USE_HUB75, but that’s “for the next version” I think :-)

- Dave

PR #287, opened 2023-05-25
Combine default and JSON effect factories, load unserialized effects

rbergen:
## Description

This:

- Combines the creation of the default effect list and the JSON effect factories into one function (`LoadEffectFactories()` in `effects.cpp`).
- Updates `EffectManager` to first load the persisted effects list from JSON, and then add any effects from the default effect factory list that are not included in the JSON file.
- Updates README.md accordingly.

With this, it is no longer necessary to clear the serialized-to-JSON effects list/config when adding a new effect. At the moment, newly added effects are always added to the end of the effects list. An update I plan to implement soon(ish) will make it possible to change the effect order to one's own liking using the device API.

## Contributing requirements


include/effectmanager.h
```diff
@@ -96,6 +95,61 @@ class EffectManager : public IJSONSerializable
...
     }
 
+    void LoadJSONAndMissingEffects(const JsonArrayConst& effectsArray)
+    {
+        std::set<int> loadedEffectNumbers;
+
+        // Create effects from JSON objects, using the respective factories in g_EffectFactories
+        auto& jsonFactories = g_EffectFactories.GetJSONFactories();
+
+        for (auto effectObject : effectsArray)
```
davepl, line 105: auto & perhaps?

include/effectmanager.h
```diff
@@ -96,6 +95,61 @@ class EffectManager : public IJSONSerializable
...
+                _vEffects.push_back(pEffect);
+                loadedEffectNumbers.insert(effectNumber);
+            }
+        }
+
+        // Now add missing effects from the default factory list
+        auto &defaultFactories = g_EffectFactories.GetDefaultFactories();
+
+        // We iterate manually, so we can use where we are as the starting point for a later inner loop
+        for (auto iter = defaultFactories.begin(); iter != defaultFactories.end(); iter++)
```
davepl, line 126: auto & again I think?

include/effectmanager.h
```diff
@@ -96,6 +95,61 @@ class EffectManager : public IJSONSerializable
...
+
+            // We found an effect (number) in the default list that we have not yet loaded from JSON.
+            //   So, we go through the rest of the default factory list to create and add to our effects
+            //   list all instances of this effect.
+            std::for_each(iter, defaultFactories.end(), [&](const EffectFactories::NumberedFactory& numberedFactory)
+                {
+                    if (numberedFactory.EffectNumber != effectNumber)
+                        return;
+
+                    auto pEffect = numberedFactory.Factory();
```
davepl, line 142: auto &

davepl (approved):
There are a couple of spots I called out where I think you're copying elements out of the collection instead of taking a reference, but it's also possible that the "auto" keyword is smart enough to PREFER a reference and do it automatically, but I'm not certain... just calling it out for review, not as an error.

include/effectmanager.h
rbergen: Here I'm following the examples in ArduinoJson's documentation. As all JSON "objects" are in fact elaborate pointers, it doesn't hurt much anyway.

include/effectmanager.h
rbergen: I tried that first, but then the code refused to compile.

davepl:
I went through the pull requests and then did the merge.  Everything is merged into the main tree of davepl/NightDriverStrip, and once we’ve tested it a bit (it runs on mesmerizer!) we can do a PR from davepl to PlummersSoftwareLLC, I think!

- Dave

include/effectmanager.h
rbergen: No, that also won't compile. In this case I believe it is because `numberedFactory.Factory()` returns a `shared_ptr` that is created in the function.

davepl:
Certainly a minor point, but out of curiosity, why wouldn’t the compiler allow you to take a reference to a shared_ptr I wonder?  I wonder if it’s because it doesn’t bump the ref count when you do so, and your reference could “go away” if decremented to zero elsewhere…

- Dave

rbergen:
I think the issue is not the type, but the fact that the shared_ptr that the reference would refer to goes out of scope (and therefore disappears) as soon as the function that creates it ends. It would as such create a dangling reference to what was a local temporary variable while the function that created it was still active.

PR #288, opened 2023-05-26
Make effects movable via the API

rbergen:
## Description

This adds the ability to change the order of the effects in the effects list, and exposes it in an endpoint to which the current and newly desired effect indices can be passed. Any effect order changes made are persisted across reboots.
The API documentation and the Postman collection file have been extended accordingly.

## Contributing requirements

### DIFF
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -482,6 +482,29 @@ public:
         return _vEffects[i]->IsEnabled();
     }
 
+    void MoveEffect(size_t from, size_t to)
+    {
+        if (from >= _vEffects.size() || to >= _vEffects.size())
+        {
+            debugW("Invalid index for MoveEffect");
+            return;
+        }
+
+        if (from == to)
+            return;
+        else if (from > to)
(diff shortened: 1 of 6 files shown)


include/effectmanager.h
```diff
@@ -482,6 +482,31 @@ class EffectManager : public IJSONSerializable
...
 
+    void MoveEffect(size_t from, size_t to)
+    {
+        if (from >= _vEffects.size() || to >= _vEffects.size())
+        {
+            debugW("Invalid index for MoveEffect");
+            return;
+        }
+
+        if (from == to)
```
davepl, line 493: What about vector::move?  I've never used it but it feels like it could do this without rotating the others around?

davepl (approved):
LGTM!

include/effectmanager.h
rbergen: As I understand it, vector::move is used to move the items from one vector to another (that is, it prevents copying when migrating them).
vector::rotate is intended for the purpose we need here. And used as it is in this code it moves (pardon, rotates) exactly one effect, leaving all others alone.

PR #293, opened 2023-05-31
Expose LEDStripEffect settings, copy effects and remove copies

rbergen:
## Description

This:

- Exposes the friendly name and maximum effect time as settings of LEDStripEffect (and as such, any derived effect in principle).
- Adds the API-exposed ability to create a copy of an existing effect, optionally with modified settings from the original.
- Adds the API-exposed ability to delete a previously created copied effect.
- Updates documentation and Postman collection file accordingly.

## Contributing requirements

### DIFF
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -69,7 +69,6 @@ template <typename GFXTYPE>
 class EffectManager : public IJSONSerializable
 {
     std::vector<std::shared_ptr<LEDStripEffect>> _vEffects;
-    size_t _cEnabled = 0;
 
     size_t _iCurrentEffect = 0;
     uint _effectStartTime;
@@ -114,6 +113,9 @@ class EffectManager : public IJSONSerializable
 
             if (pEffect != nullptr)
             {
+                if (effectObject[PTY_COREEFFECT].as<int>())
+                    pEffect->MarkAsCoreEffect();
+
                 _vEffects.push_back(pEffect);
                 loadedEffectNumbers.insert(effectNumber);
             }
@@ -141,7 +143,11 @@ class EffectManager : public IJSONSerializable
 
                     auto pEffect = numberedFactory.Factory();
                     if (pEffect != nullptr)
+                    {
+                        // Effects in the default list are core effects. These can be disabled but not deleted.
+                        pEffect->MarkAsCoreEffect();
                         _vEffects.push_back(pEffect);
+                    }
                 }
             );
 
@@ -153,7 +159,6 @@ class EffectManager : public IJSONSerializable
     void ClearEffects()
     {
         _vEffects.clear();
-        _cEnabled = 0;
     }
 
 public:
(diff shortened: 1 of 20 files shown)


include/effectmanager.h
```diff
@@ -507,6 +509,83 @@ class EffectManager : public IJSONSerializable
...
+        // We don't allow core effects to be deleted
+        if (_vEffects[index]->IsCoreEffect())
+            return false;
+
+        if (index == _iCurrentEffect)
+            NextEffect();
+
+        DisableEffect(index, true);
+
+        _vEffects.erase(_vEffects.begin() + index);
```
davepl, line 579: Fancy, for some reason I've never done this before!

include/types.h
```diff
@@ -57,7 +57,8 @@ struct SettingSpec
         Float,
         Boolean,
         String,
-        Palette
+        Palette,
+        Color
```
davepl, line 61: You're well on your way to a complete OLEAUTO type spec!

davepl (approved):
LGTM!

include/effectmanager.h
```diff
@@ -507,6 +509,83 @@ class EffectManager : public IJSONSerializable
         SaveEffectManagerConfig();
     }
 
+    // Creates a copy of an existing effect in the list. Note that the effect is created but not yet added to the effect list;
+    //   use the AppendEffect() function for that.
+    std::shared_ptr<LEDStripEffect> CopyEffect(size_t index)
```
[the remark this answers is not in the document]
rbergen: I think you found the answer a number of lines down in the code, but for the record: it copies the properties of the effect as they are set at the time the effect is copied. It does this by serializing the original effect to JSON and creating the copy by deserializing it.

include/effectmanager.h
rbergen: Neither had I. :)

include/globals.h
```diff
@@ -143,6 +143,7 @@
 // C Helpers and Macros
 
 #define NAME_OF(x)          #x
+#define ACTUAL_NAME_OF(x)   ((#x) + 1)
```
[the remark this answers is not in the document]
rbergen: That's fair feedback, which also points out that the macro name is not ideally chosen. It does what NAME_OF does, but just skips the first character; it will yield the actual property name in case member variables are prefixed with an underscore. I just couldn't get myself to name it something like `NAME_OF_WITHOUT_FIRST_CHARACTER`.

Anyway, I just noticed that VS Code shows the comment with the symbol when one hovers the mouse over the latter, if the comment immediately precedes it. So, that's what I'll do.

include/types.h
rbergen: Haha, why did I have a feeling you would bring that up?

rbergen:
Thanks for the review! I'll add the ACTUAL_NAME_OF comment and merge when CI completes.

PR #297, opened 2023-06-01
Isolate current effect save

rbergen:
## Description

This:

- Isolates the saving/storage of the current effect index from the rest of the effects config
- Makes enabling/disabling the saving (remembering) of the current effect (index) a device setting

## Contributing requirements

### DIFF
--- a/src/deviceconfig.cpp
+++ b/src/deviceconfig.cpp
@@ -96,6 +96,13 @@ DeviceConfig::DeviceConfig()
         "The hostname or IP address of the NTP server to be used for time synchronization.",
         SettingSpec::SettingType::String
     );
+    settingSpecs.emplace_back(
+        NAME_OF(rememberCurrentEffect),
+        "Remember current effect",
+        "A boolean that indicates if the current effect index should be saved after an effect transition, so the device resumes "
+        "from the same effect when restarted. Enabling this will lead to more wear on the flash chip of your device.",
+        SettingSpec::SettingType::String
+    );
 
     writerIndex = g_ptrJSONWriter->RegisterWriter(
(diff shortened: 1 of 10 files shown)


Louis-Riel (approved):

rbergen:
Thank you @Louis-Riel!

davepl (approved):
LGTM!  Was there any conclusion on whether writing to NVS is better than SPIFFS?  I'd imaging they're the same...

Louis-Riel:
NVS is normally used to store crypto keys and such, so the api's are tooled as such. It would fit better to put then in spiff. Also, the advantage of putting them in SPIFF, is that this is a file system, and if we choose to, we could just point the web server to serve that json from the file system, no NVS load-unload and all that. I'm pretty sure the speed is the same, spiff is just tooled better I think.

rbergen:
I think @davepl is referring to using NVS vs SPIFFS to track the "current effect" specifically. SPIFFS is mildly wasteful with flash pages when writing small files (like the 1 to 2 byte case of the current effect index). However, as the NVS partition is currently only 8192 bytes in size, I think that bottom-line it's much of a muchness.

src/deviceconfig.cpp
```diff
@@ -96,6 +96,13 @@ DeviceConfig::DeviceConfig()
         "The hostname or IP address of the NTP server to be used for time synchronization.",
         SettingSpec::SettingType::String
     );
+    settingSpecs.emplace_back(
+        NAME_OF(rememberCurrentEffect),
+        "Remember current effect",
+        "A boolean that indicates if the current effect index should be saved after an effect transition, so the device resumes "
+        "from the same effect when restarted. Enabling this will lead to more wear on the flash chip of your device.",
+        SettingSpec::SettingType::String
```
mggates39, line 104: I think this should be Boolean.  I have generated a PR to update.  I forgot I could comment on closed PRs.

PR #321, opened 2023-06-12
Save memory

rbergen:
## Description

This PR includes a number of changes that reduce memory usage, specifically on devices that run many effects. For one, the Mesmerizer project in its current state leaves so little available RAM that the web server fails to start at boot.

Note that the ways to create and return effect setting specs, and for an effect to perform data fetches from the Internet change with the merge of this PR. For both, `PatternSubscribers` can (still) be used as a reference.

Furthermore, this PR includes a new version of the CI (GitHub Actions) script. It now uses the recommended installation script to install PlatformIO.

## Contributing requirements

### DIFF
--- a/include/network.h
+++ b/include/network.h
@@ -110,7 +110,7 @@
     class NetworkReader
     {
       // We allow the main network reader task entry point function to access private members
-      friend void IRAM_ATTR NetworkTaskEntry(void *);
+      friend void IRAM_ATTR NetworkReaderTaskEntry(void *);
 
     private:
 
--- a/include/network.h
+++ b/include/network.h
@@ -144,7 +144,7 @@
 
       // Add a reader to the collection. Returns the index of the added reader, for use with FlagReader().
       //   Note that if an interval (in ms) is specified, the reader will run for the first time after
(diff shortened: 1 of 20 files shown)


include/network.h
```diff
@@ -107,6 +107,54 @@
       return str_sprintf("%02X:%02X:%02X:%02X:%02X:%02X", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
     }
 
-
-
+    class NetworkReader
```
davepl, line 110: Maybe a comment, like what is NetworkReader and what does it do?   Just a summary would be handy!

davepl (approved):
That's a lot of code!  But I didn't spot any errors!

davepl:
Done, here’s what I think I did:

Merged your changes into PS/Main
Picked up those changes into davepl/main
Updated davepl/main with LEDViewer changes

So everything should be assembled in my davepl/main branch, and once it passes CI, we can merge it into PS/main, I think.

With it running both our combined changes, I’m seeing 104K free at this point.

Thanks!
Dave

PR #326, opened 2023-06-14
Fix minor typos in comments and code.

robertlipe:
## Description
Correct various minor typos in comments and code found by fresh eyes reading
code for the first time.

(This is NOT to statt a style war, It's also not to claim I never make typos.)

## Contributing requirements

### DIFF
--- a/src/drawing.cpp
+++ b/src/drawing.cpp
@@ -273,7 +273,7 @@ void ShowStrip(uint16_t numToShow)
 
 // CalcDelayUntilNextFrame
 //
-// Returns the amount of time to wait patiently until it is time to draw the next frame, up to one second max
+// Returns the amount of time to wait patiently until it's time to draw the next frame, up to one second max
 
 int CalcDelayUntilNextFrame(double frameStartTime, uint16_t localPixelsDrawn, uint16_t wifiPixelsDrawn)
 {


include/effects/matrix/PatternPongClock.h
```diff
@@ -290,7 +290,7 @@ class PatternPongClock : public LEDStripEffect
         }
 
         // move bat 1 towards target
-        // if bat y greater than target y move down until hit 0 (dont go any further or bat will move off screen)
+        // if bat y greater than target y move down until hit 0 (don't go any further or bat will move off-screen)
```
rbergen, line 293: I argue that "off" and "screen" should not be connected with a hyphen, as it doesn't concern a compound adjective.

```suggestion
        // if bat y greater than target y move down until hit 0 (don't go any further or bat will move off screen)
```

include/effects/matrix/PatternPongClock.h
```diff
@@ -309,7 +309,7 @@ class PatternPongClock : public LEDStripEffect
         LEDMatrixGFX::backgroundLayer.fillRectangle(BAT1_X - 1, bat1_y, BAT1_X, bat1_y + BAT_HEIGHT, rgb24(255,255,255));
         //      }
 
-        // move bat 2 towards target (dont go any further or bat will move off screen)
+        // move bat 2 towards target (dont go any further or bat will move off-screen)
```
rbergen, line 312: Same comment as before. Also, "dont" does need an apostrophe.

```suggestion
        // move bat 2 towards target (don't go any further or bat will move off screen)
```

src/drawing.cpp
```diff
@@ -148,7 +148,7 @@ void MatrixPreDraw()
 
 // WiFiDraw
 //
-// Draws forom WiFi color data if available, returns pixels drawn this frame
+// Draws from WiFi color data if available, returns pixels drawn from this frame
```
rbergen, line 151: The second "from" is correctly absent in the original sentence. The function returns the pixels drawn while composing the current frame, so "drawn this frame".

```suggestion
// Draws from WiFi color data if available, returns pixels drawn this frame
```

src/drawing.cpp
```diff
@@ -238,7 +238,7 @@ uint16_t LocalDraw()
 
 // ShowStrip
 //
-// ShowStrip sends the data to the LED strip.  If its fewer than the size of the strip, we only send that many.
+// ShowStrip sends the data to the LED strip.  If it is fewer than the size of the strip, we only send that many.
```
rbergen, line 241: I think the originally intended spelling is "it's".

```suggestion
// ShowStrip sends the data to the LED strip.  If it's fewer than the size of the strip, we only send that many.
```

src/drawing.cpp
```diff
@@ -273,7 +273,7 @@ void ShowStrip(uint16_t numToShow)
 
 // CalcDelayUntilNextFrame
 //
-// Returns the amount of time to wait patiently until its time to draw the next frame, up to one second max
+// Returns the amount of time to wait patiently until it is time to draw the next frame, up to one second max
```
rbergen, line 276: Same comment as before.

```suggestion
// Returns the amount of time to wait patiently until it's time to draw the next frame, up to one second max
```

rbergen (commented):
I wouldn't go near accusing you of starting a style war, but from a pure typo perspective I'd say you are pushing the envelope a little, here and there. 🙂

That's all good, but there's a few things I think do need to be changed before we merge. They're mentioned in the comments below. I've included a suggested alternative with each comment.

src/drawing.cpp
rbergen: What I'm saying is that Dave clearly intended to write "it's", in the meaning of the first bullet of on the page you linked to:

> If you mean it is or it has*, then use it’s

He just missed the apostrophe.

rbergen:
> As a general question: when do we get out of new contributor jail so our presubmits actually run so that reviewers can sensibly review patches?

Without wanting to make a big deal out of it, I think your choice of words in this particular comment is a bit passive-aggressive. Although I can personally handle it, I do think we can do without. 

Then to answer your question: as we do in other PlummersSoftwareLLC projects, we choose to review possible submissions first and then run CI on them. The reasons are that:

1. We're using free computational capacity from GitHub, and CI runs on this project are quite expensive in the computational sense. We're just mindful of what we use here, also considering that there are other projects within the organization that also have CI enabled (for one, the Primes project now starts >150 jobs every time we run CI on it).
2. We also allow contributors to propose changes to the CI workflow itself. As the impact of such changes can be huge, we really want to review those before we run them in the context this project.

In fact, if we could disable automatic Action execution for the maintainers then we would; I just haven't found the button in the GitHub Settings pages to make it work that way.

What's good to know is that if you want to run CI on whatever it is you aim to merge, you don't need to wait until you open your PR here. You can just enable GitHub Actions on your fork; GitHub will then kick off CI every time you push anything to it. This is how at least Dave and I work as well; we don't open a PR on the main tree until CI on our fork has completed successfully first.

I hope this helps clarify things somewhat.

Cheers,
Rutger

rbergen:
> Fair. It was unnecessarily aggressive-aggressive. Sorry that I'm being grumpy

Don't worry about it. Water under the bridge.

> Interesting. I don't think I've met any other project on GH that runs that way.

That's a fair comment. We actually settled on this way of working on the Primes project, where there were periods during which we had dozens of PRs per week, all from different people. In fact, occasionally we had over a dozen PRs _per day_. We ran into some issues there that we found were much easier to manage by adopting this approach. Having established that, we stuck with it.

On an earlier comment (and that will be my last comment on that one 🙂):

> So please integrate that with or without the it->it is change.

I think it was an "its" to "it is"/"it's" change. The real problem is that the sentence it's part of doesn't work grammatically either way. I just couldn't and can't be bothered to fix that root cause.

With that said, I'll run the gauntlet and then proceed to merge!

PR #336, opened 2023-06-26
Add range to SettingSpec, parallelize CI

### LINKED ISSUE #333: Settings Issue - Need min, max, etc
For most settings, like HueSpeed and ScrollSpeed and so on, sometimes only a range of reasonable values makes sense.  I imagine the UI will eventually display a slider for such settings, and it also needs to know a min and max range.

So, I think properties like PTY_HUESPEED may need either system-wide ranges or, better yet, effect-defined ranges.  

No idea how best to handle or implement this, just calling out that it's a likely need!

rbergen:
## Description

This:

- Pins the espressif32 version to 6.3.0 or over - this stops builds from breaking if the most recent release is < 6.1.
- Parallelizes CI for the vast majority of builds, which brings their overall runtime down to approximately 10 minutes. Only builds that include a push to the web installer Pages site are still sequential.
- Adds optional range specifiers (minimum and maximum value) to the `SettingSpec` class. Credits to @robertlipe for bringing `std::optional` to my attention again. This fixes #333.

## Contributing requirements


davepl (approved):
LGTM!

PR #345, opened 2023-07-05
Fix S3 crash, make invalid pixel draws errors

davepl:
## Description
Increases stack size to prevent S3 from crashing in JSONWriter
Makes writing out of pixel bounds at least an error message
Adds DrawSafeCircle which doesn't assume it's OK to write out of bounds
Fixes a number of effects that did write out of bounds


rbergen (approved):
LGTM!

PR #346, opened 2023-07-06
8M Flash, 60Hz, enable OTA, improve OTA

davepl:
8M Flash, 60Hz, enable OTA, improve OTA

Adds new partition tabe file with 8M suffice for mesmerizer that makes use of 8M of flash, which makes enough room for a second OTA partition.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -17,8 +17,8 @@ default_envs =
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     = /dev/cu.usbserial-8411101
-monitor_port    = /dev/cu.usbserial-8411101
+upload_port     =
+monitor_port    =
 upload_speed    = 1500000
 
 build_flags     = -std=gnu++17
--- a/platformio.ini
+++ b/platformio.ini
@@ -9,7 +9,7 @@
 ; https://docs.platformio.org/page/projectconf.html
 
 [platformio]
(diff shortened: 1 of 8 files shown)


platformio.ini
```diff
@@ -9,16 +9,16 @@
...
 ; ==================
 ; Base configuration
 ;
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     =
-monitor_port    =
+upload_port     = /dev/cu.usbserial-8411101
+monitor_port    = /dev/cu.usbserial-8411101
```
rbergen, line 21:
```suggestion
upload_port     =
monitor_port    =
```

rbergen (commented):
I suggest we continue without port paths :), but LGTM beyond that!
If you agree with the ports thing, you can just commit my suggested change. With that in, this can be merged as far as I'm concerned.

src/main.cpp
```diff
@@ -713,6 +710,9 @@ void loop()
             Serial.println(strOutput);
         }
 
-        delay(5);
+        // Once an update is underway, we loop tightly on ArduinoOTA.handle.  Otherweise we delay a bit to share the CPU.
```
rbergen, line 713:
```suggestion
        // Once an update is underway, we loop tightly on ArduinoOTA.handle.  Otherwise we delay a bit to share the CPU.

```

PR #351, opened 2023-07-09
Refactor globals, drop YouTubeSight

### LINKED ISSUE #273: Subscriber effect blocks draw thread
## Bug report
When the API request fails, it blocks the main draw thread for about a second, pausing drawing

Not sure if this a priority issue or if there just isn't CPU time for everything that needs to happen, but at a minimum we should "grow" the delay between attempts so that its not several times per minute if it's failing.

rbergen:
## Description

This wraps a number of currently individually declared global variables (objects) into one system container class `SystemContainer`, made available through global variable `g_ptrSystem`. The container class provides member functions for setting up the objects it contains, checking their initialization status (ready or not) and getting references to them. For one, this allows the time of initialization of the objects to be chosen by the developer, while allowing dependent code to check for and handle not-yet-available scenarios.

Note that:

- Not all global variables (those with names starting with `g_` by project convention) are included in the container class. For one, variables that are global because of where/how they are declared but effectively used in one CPP file only - and possibly the associated "primary" header file - are considered an "internal affair" of the CPP file in question, and not made more broadly available.
- To prevent code duplication, `SystemContainer` is composed largely through a number of macros at the top of `systemcontainer.h`. This may make the code of the class itself somewhat opaque. As VSCode's intellisense allows the class' user to interact with the expansions of the macros (i.e. the resulting member functions), I consider this to be acceptable.

In addition to the introduction of `SystemContainer`, this PR:

- Replaces `YouTubeSight` with a straightforward `HTTPClient` GET. This should fix #273. If it doesn't, it at least simplifies the implementation of `PatternSubscribers`. 
- Updates the on-board API documentation
- Makes a number of effects use `LEDStripEffect`'s `g()` function to retrieve `GFXBase`, instead of pulling it from `EffectManager`

## Contributing requirements


davepl (approved):
LGTM!

PR #362, opened 2023-07-15
Power Throttling

davepl:
Add power throttling, fixes some page flip bugs


include/gfxbase.h
```diff
@@ -732,13 +731,6 @@ class GFXBase : public Adafruit_GFX
         _targetPalette = CRGBPalette16(CRGB::Black, CRGB::Blue, CRGB::Aqua, CRGB::White);
     }
 
-    // write one pixel with the specified color from the current palette to coordinates
```
davepl, line 735: No callers

include/globals.h
```diff
@@ -123,8 +123,8 @@
 #define FLASH_VERSION          37    // Update ONLY this to increment the version number
 
 #ifndef USE_MATRIX                   // We support strips by default unless specifically defined out
-    #ifndef USESTRIP
-        #define USESTRIP 1
+    #ifndef USE_STRIP
```
davepl, line 126: Just making the name USE_STRIP match USE_MATRIX

include/socketserver.h
```diff
@@ -316,7 +316,7 @@ class SocketServer
                 // one big read one time would work best, and we use that to copy it to a regular RAM buffer.
 
                 #if USE_PSRAM
-                    std::unique_ptr<uint8_t []> _abTempBuffer = std::make_unique<uint8_t []>(MAXIMUM_PACKET_SIZE);
+                    std::unique_ptr<uint8_t []> _abTempBuffer = std::make_unique<uint8_t []>(MAXIMUM_PACKET_SIZE+1);    // Plus one for uzlib buffer overreach bug
```
davepl, line 319: UZLib has a bug wherein it assumes you've allocated one more byte than needed.

include/effectmanager.h
```diff
@@ -96,6 +96,7 @@ class EffectManager : public IJSONSerializable
             _clearTempEffectWhenExpired = true;
 
             // This is a hacky way to ensure that we start the correct effect after the temporary one
+            // REVIEW: Explain, I don't get it!
```
rbergen, line 99: The switching to the next effect is taken care of by NextEffect(), which starts with increasing _iCurrentEffect. We therefore need to decrease it here, to make sure that the first effect after the temporary one that is shown at startup - i.e. the NightDriver logo - is the one we want (either the then current one when the chip was powered off, or the one at index 0).

I've pushed a change to my issue-349 branch that adds this comment at this place in the code.

include/effects/matrix/PatternSubscribers.h
```diff
@@ -78,6 +78,12 @@ class PatternSubscribers : public LEDStripEffect
             return;
         }
 
+        if (youtubeChannelGuid.isEmpty())
+        {
+            debugW("No YouTube subscriber guid, so skipping check...");
+            return;
+        }
+
```
rbergen, line 86: That's a good check. Should have added that myself!

include/ledmatrixgfx.h
```diff
@@ -82,6 +82,29 @@ class LEDMatrixGFX : public GFXBase
     {
         matrix.setBrightness(amount);
     }
+    
+    // EstimatePowerDraw
+    //
+    // Estimate the total power load for the board and matrix by walking the pixels and adding our previously measured
+    // power draw per pixel based on what color and brightness each pixel is/
+
+    int EstimatePowerDraw()
```
rbergen, line 91: I'm glad I got an out-of-band explanation of this, otherwise I don't think I would have followed. :)

include/socketserver.h
rbergen: Nice.

src/main.cpp
```diff
@@ -189,6 +189,13 @@ DRAM_ATTR bool g_bUpdateStarted = false;
 DRAM_ATTR AppTime g_AppTime;                                                        // Keeps track of frame times
 DRAM_ATTR bool NTPTimeClient::_bClockSet = false;                                   // Has our clock been set by SNTP?
 
+#if USE_MATRIX
+    extern int g_MatrixPowerMilliwatts;                                             // Matrix power draw in mw
+    extern uint8_t g_MatrixScaledBrightness;                                        // 0-255 scaled brightness to stay in limit
+#endif
+
+extern DRAM_ATTR std::unique_ptr<EffectManager<GFXBase>> g_ptrEffectManager;       // The one and only global effect manager
```
rbergen, line 197: I don't think we need this any more. It's now part of SystemContainer (g_ptrSystem).

```suggestion
```

rbergen (commented):
I've made a few comments, see below. I'd recommend the removal of g_ptrEffectManager is applied before we merge - I've suggest a change to that effect in the comment in question.

Furthermore, I'm not sure why the simh module was added to this particular project? I've removed it before and the code seems to build fine without it.

include/effectmanager.h
rbergen, line 99: You know what, I'll just suggest a change to add it here. :)

```suggestion
            //   The switching to the next effect is taken care of by NextEffect(), which starts with
            //   increasing _iCurrentEffect. We therefore need to decrease it here, to make sure that
            //   the first effect after the temporary one is the one we want (either the then current
            //   one when the chip was powered off, or the one at index 0).
```

PR #369, opened 2023-07-20
Optimize drawing and framerate

davepl:
Includes a number of drawing improvements like XY() and changes the matrix framerate.

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -1234,11 +1234,11 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 #endif
 
 #ifndef MATRIX_REFRESH_RATE
-#define MATRIX_REFRESH_RATE 120
+#define MATRIX_REFRESH_RATE 180
 #endif
 
 #ifndef MATRIX_CALC_DIVIDER
-#define MATRIX_CALC_DIVIDER 2
+#define MATRIX_CALC_DIVIDER 3
 #endif
 
 
@@ -1610,6 +1610,13 @@ inline String str_sprintf(const char *fmt, ...)
 // C Helpers
 //
 // Simple inline utility functions like random numbers, mapping, conversion, etc
+
+inline static uint8_t randombyte(uint8_t lower = 0, uint8_t upper = 255)
+{
+    auto result = (lower + ((upper - lower) * rand()) / RAND_MAX);
+    return result;
+}
+
 
 inline static float randomfloat(float lower, float upper)
 {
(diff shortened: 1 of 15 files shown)


include/globals.h
```diff
@@ -1477,21 +1477,24 @@ inline String str_sprintf(const char *fmt, ...)
...
+    static std::mt19937 gen(rd());
+
+    if constexpr (std::is_integral<T>::value) {
+        std::uniform_int_distribution<T> distrib(lower, upper);
+        return distrib(gen);
+    } else if constexpr (std::is_floating_point<T>::value) {
+        std::uniform_real_distribution<T> distrib(lower, upper);
+        return distrib(gen);
+    }
 }
```
rbergen, line 1498: Nice, very elegant!

src/main.cpp
```diff
@@ -164,25 +164,21 @@
 
 void IRAM_ATTR ScreenUpdateLoopEntry(void *);
 
-// Global support objects and global values
+//
+// Global Variables
+//
+
 DRAM_ATTR std::unique_ptr<SystemContainer> g_ptrSystem;
+DRAM_ATTR std::unique_ptr<EffectFactories> g_ptrEffectFactories = nullptr;
```
rbergen, line 172: I intentionally didn't include g_ptrEffectFactories in the main.cpp globals. In my mind it's a global that's local to network.cpp - which has effectmanager.h as one of it's two primary defines. I suggest we leave it in network.cpp.

rbergen (commented):
I think we should leave g_ptrEffectFactories in effects.cpp, see my second comment below. Other than that, it looks good to me. 

(And I'll even be able to get over a merge with g_ptrEffectFactories in main.cpp, if you strongly prefer it to live there. :) )

src/main.cpp
rbergen: Suggested change here:

```suggestion
```

src/effects.cpp
```diff
@@ -258,7 +258,6 @@ std::shared_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color)
 #define STARRYNIGHT_PROBABILITY 1.0
 #define STARRYNIGHT_MUSICFACTOR 1.0
 
-DRAM_ATTR std::unique_ptr<EffectFactories> g_ptrEffectFactories = nullptr;
 std::map<int, JSONEffectFactory> g_JsonStarryNightEffectFactories =
```
rbergen, line 261: Suggestion to move g_ptrEffectFactories back to here:

```suggestion
DRAM_ATTR std::unique_ptr<EffectFactories> g_ptrEffectFactories = nullptr;

std::map<int, JSONEffectFactory> g_JsonStarryNightEffectFactories =
```

PR #365, opened 2023-07-17
Issue 349 part 1, refactor globals continued, power limit setting

rbergen:
## Description

This:
- Restructures the code that performs hardware initialization for all 3 device types (strip, matrix, hexagon). Note that this has been tested with a matrix (Mesmerizer) and a single strip (Spectrum), but not with other scenarios (like multiple strips and hexagon). This is the first step towards the resolution of #349.
- Adds the devices (`g_aptrDevices`) and buffer managers (`g_aptrBufferManager`) to the `SystemContainer` class (`g_ptrSystem`)
- Bundles a number of "simple value" global variables in a new class `Values`, made available as `g_Values`.
- Sets the default value for saving the current effect index to `true`
- Implements a power limit device setting, and as such fixes #363
- Expands the use of PSRAM, where useful

## Contributing requirements

* [x] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [x] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [x] I selected `main` as the target branch.
* [x] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #370, opened 2023-07-22
Add PrepareFrame() and PostProcessFrame() to GFXBase and subclasses

### LINKED ISSUE #349: WORKITEM: InitializeHardware, PrepareFrame, PresentFrame
I think BaseGFX should likely grow by three virtual functions:

InitializeHardware, PrepareFrame, PresentFrame

They would be overridden as follows:

LEDStripGFX::InitializeHardware would do the AddLeds calls
LEDMatrixGFX::InitializeHardware would do the matrix creation

LEDStripGFX::PrepareFrame needs not exist, as there's no prep
LEDMatrixGFX::PrepareFrame would do the matrix prep code currently in drawing.cpp

PresentFrame would call ShowLEDs() on the strip and do whatever the drawing code does for matrices today in the LEDMatrixGFX::PresentFrame call

This would simplify the init and drawing codepaths a fair bit and remove ifdef conditional code.

rbergen:
## Description

This fixes #349, to the extent #365 hadn't already. It does this by:

- Adding PrepareFrame() and PostProcessFrame() virtual functions to `GFXBase`
- Overloading both in `LEDMatrixGFX`
- Overloading PostProcessFrame() in `LEDStripGFX`

These changes have been tested on a Mesmerizer board running Mesmerizer, and an M5StickCPlus running Spectrum.

Also, this:

- Declares globally defined variables that are actually only meant to be used by one compilation unit (.cpp file) `static` and gives them an `l_` prefix (instead of `g_` for "true" globals).
- Fixes a regression that NTP time was no longer updated at regular intervals.

**Note:** I am aware this PR will cause merge conflicts with #369. I will address these when either that PR or this one has been merged. I'm opening this now so people can take a look at it, and/or test it if they want to.

## Contributing requirements

### DIFF
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -60,13 +60,11 @@ bool ReadCurrentEffectIndex(size_t& index);
 
 std::shared_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color);
 std::shared_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color, CRGB color2);
-extern DRAM_ATTR std::unique_ptr<EffectFactories> g_ptrEffectFactories;
 
 // EffectManager
 //
 // Handles keeping track of the effects, which one is active, asking it to draw, etc.
 
-template <typename GFXTYPE>
 class  EffectManager : public IJSONSerializable
 {
     std::vector<std::shared_ptr<LEDStripEffect>> _vEffects;
@@ -80,7 +78,7 @@ class  EffectManager : public IJSONSerializable
     bool _clearTempEffectWhenExpired = false;
     bool _newFrameAvailable = false;
 
-    std::vector<std::shared_ptr<GFXTYPE>> _gfx;
+    std::vector<std::shared_ptr<GFXBase>> _gfx;
     std::shared_ptr<LEDStripEffect> _tempEffect;
 
     void construct(bool clearTempEffect)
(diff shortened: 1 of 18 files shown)


include/effectmanager.h
```diff
@@ -354,11 +354,6 @@ class  EffectManager : public IJSONSerializable
         return _gfx[index];
     }
 
-    size_t gCount()
-    {
-        return _gfx.size();
-    }
-
```
rbergen, line 361: This was used in one place, where FastLED.count() was a better alternative.

include/effects/strip/faneffects.h
```diff
@@ -450,8 +450,6 @@ class FanBeatEffect : public LEDStripEffect
   }
 };
 
-extern void ShowTM1814();
-
```
rbergen, line 454: I removed all references to ShowTM1814(). It didn't seem to be used anywhere, and CI is happy without it.

include/effects/strip/faneffects.h
```diff
@@ -479,11 +477,8 @@ class CountEffect : public LEDStripEffect
         if (t >= OPEN_LEN)
           t -= OPEN_LEN;
       }
-#if ATOMISTRING
```
rbergen, line 482: This macro is not defined anywhere.

include/ledmatrixgfx.h
```diff
@@ -63,12 +63,10 @@ class LEDMatrixGFX : public GFXBase
     static const uint8_t kDefaultBrightness = 255; // full (100%) brightness
     static const rgb24   defaultBackgroundColor;
 
-    #if USE_MATRIX
```
rbergen, line 66: This condition is already checked at the top of the file.

include/effectmanager.h
```diff
@@ -60,13 +60,11 @@ bool ReadCurrentEffectIndex(size_t& index);
 
 std::shared_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color);
 std::shared_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color, CRGB color2);
-extern DRAM_ATTR std::unique_ptr<EffectFactories> g_ptrEffectFactories;
 
 // EffectManager
 //
 // Handles keeping track of the effects, which one is active, asking it to draw, etc.
 
-template <typename GFXTYPE>
```
rbergen, line 69: EffectManager is only ever instantiated with GFXBase as the graphics type, so it doesn't need to be a template.

include/effectmanager.h
davepl: Nice cleanup!

src/drawing.cpp
```diff
@@ -439,18 +276,7 @@ void IRAM_ATTR DrawLoopTaskEntry(void *)
...
-            MatrixPostDraw(localPixelsDrawn + wifiPixelsDrawn);
-        #endif
-
-        #if USE_STRIP
-            if (wifiPixelsDrawn)
-                ShowStrip(wifiPixelsDrawn);
-            else if (localPixelsDrawn)
-                ShowStrip(localPixelsDrawn);
-        #endif
+        graphics->PostProcessFrame(wifiPixelsDrawn, localPixelsDrawn);
```
davepl, line 279: So the Screen class also has a prepare/submit frame virtual function.  Not for this checkin, but I'd suggest unifying those since it inherits from GFXBase as well.  In other words, make Screen override Prepare/Post frame calls, and then in the one or two places in screen.h where the current begin/endframe, replace it woth Prepare/Post

src/effects.cpp
```diff
@@ -258,8 +258,8 @@ std::shared_ptr<LEDStripEffect> GetSpectrumAnalyzer(CRGB color)
 #define STARRYNIGHT_PROBABILITY 1.0
 #define STARRYNIGHT_MUSICFACTOR 1.0
 
-DRAM_ATTR std::unique_ptr<EffectFactories> g_ptrEffectFactories = nullptr;
-std::map<int, JSONEffectFactory> g_JsonStarryNightEffectFactories =
+static DRAM_ATTR std::unique_ptr<EffectFactories> l_ptrEffectFactories = nullptr;
```
davepl, line 261: Prefix of l is weird!  We always uses s_ but I suppose some might confuse that with string, so I'll live with it :-)

davepl (approved):
Looks good!

src/drawing.cpp
rbergen: Ah, I didn't know that, thanks for the pointer! I'll look into that once on this PR CI has passed, I've tested it works on my devices, and merged it.

src/effects.cpp
rbergen: Maybe I can redeem myself a little bit by saying hand-on-heart that I did consider s_ as well. :)

Anyway, `static` still is the wrong keyword for a compilation unit variable.

davepl: Wait, what?  Am I thinking K&R or what?  I was sure static meant "at global scope but accessible only here".

davepl:
That’s what I thought it meant, and so I’m confused!  

Pretty sure you’d get a linker duplicate inclusion error if there were two copies included, though, unless you’re referring to a scenario I haven’t considered!

- Dave

rbergen:
Ok, now I'm confused by the confusion. When I said "compilation unit variable" I meant "at global scope but accessible only here" - to me those are semantically equivalent. If the actual experts are shouting at their screens now I've said that, I do apologize. :)

davepl:
I think we all think the same thing, but there was a ocmment in the review that it meant something other than what I thought it meant, and I’m confused as to where I was confused :-)

That said, I’ve never seen the linker behavior you’re speaking of, but I don’t know compilers that well!

 -Dave

PR #376, opened 2023-07-25
Fix NTP update and restore use of (checked) NUM_CHANNELS

rbergen:
## Description

This fixes two bugs that were recently introduced:

- The static variable that keeps track of the last successful NTP time update is now in the right place and should as such indeed retain values across invocations.
- NUM_CHANNELS is used once again as the number of available channels, because FastLED.count() returns a value that is one too high if an on-board LED is present. Many thanks to @davepl for finding that one.
- To prevent an invalid NUM_CHANNELS causing problems, defining it to a value > 8 will now trigger a compile-time error.

## Contributing requirements


rbergen:
Closing this, as the changes contained herein were included in #380.

PR #380, opened 2023-07-25
Fix response packet, lantern effect, add 3D model, remote control repeat, more

davepl:
Fix response packet, lantern effect, add 3D model, remote control repeat and throttling, fix defref array bug, fix FPS calc, etc


include/effects/strip/fireeffect.h
```diff
@@ -199,16 +199,23 @@ class FireEffect : public LEDStripEffect
...
+            auto sum = 0;
+            for (int j = 0; j < CellsPerLED; j++)
+                sum += heat[i*CellsPerLED + j];
+            auto avg = sum / CellsPerLED;
 
-            CRGB color = GetBlackBodyHeatColor(heat[i*CellsPerLED]/(float)std::numeric_limits<uint8_t>::max());
+            #if LANTERN
+                CRGB color = CRGB(avg, avg * .45, avg * .08);
+            #else
+                CRGB color = GetBlackBodyHeatColor(avg/(float)std::numeric_limits<uint8_t>::max());
```
davepl, line 210: What a fancy way to write 255.  I think I'll take the rest of the day off!

include/effects/strip/fireeffect.h
```diff
@@ -199,16 +199,23 @@ class FireEffect : public LEDStripEffect
 
         for (int i = 0; i < LEDCount; i++)
         {
+            auto sum = 0;
```
davepl, line 202: Now uses the average of the pixels in the group, rather than just the first value, and looks MUCH more realistic.

src/ledmatrixgfx.cpp
```diff
@@ -88,7 +88,7 @@ void LEDMatrixGFX::PrepareFrame()
             backgroundLayer.drawString(MATRIX_WIDTH / 2 - (3 * output.length()), MATRIX_HEIGHT / 2 - 5, rgb24(255, 255, 255), rgb24(0, 0, 0), output.c_str());
         #endif
 
-        auto graphics = g_ptrSystem->EffectManager()[0];
```
davepl, line 91: It's equivalent, but I'm not a fan of operator overloading, and it hid a bug, so changed to the g() function which was there anyway.

src/screen.cpp
```diff
@@ -248,24 +248,26 @@ void CurrentEffectSummary(bool bRedraw)
     static auto lastSerial = 0;
     auto yh = 2; // Start at top of screen
```
davepl, line 250: Now it fits on the regular M5 stick, where the screen is only 160x80 or similar

include/effects/strip/fireeffect.h
rbergen: Haha, yeah, unleashing `std::numeric_limits` on `uint8_t` is like using a motorized mower on a single blade of grass. :)

src/ledmatrixgfx.cpp
rbergen: That's a good clean-up, and increases consistency with how the first graphics channel is retrieved everywhere else already. I'm now wondering why that didn't click when I looked at this before.

davepl:
The lazy man in me already considered that, and decided that since everything in the code is duration based, it should still work.  I’m not aware of anything that fires “every X periods but not greater than”, so nothing I can think of should be missed.  It’s a valid concern if we had timers or alarms based on time of day, but do we?

_ Dave

rbergen (approved):
LGTM!

PR #382, opened 2023-07-28
Warnings

davepl:
Fix numerous compiler warnings.  Updates to flash v40.  5 min NTP window.

### DIFF
--- a/include/effectfactories.h
+++ b/include/effectfactories.h
@@ -115,10 +115,7 @@ class EffectFactories
     void AddEffect(int effectNumber, const DefaultEffectFactory& defaultFactory, const JSONEffectFactory& jsonFactory)
     {
         defaultFactories.emplace_back(effectNumber, defaultFactory);
-
-        // BUGBUG perf (davepl) Consider using 'jsonFactories.try_emplace without the first lookup
-        if (jsonFactories.count(effectNumber) == 0)
-            jsonFactories[effectNumber] = jsonFactory;
+        jsonFactories.try_emplace(effectNumber, jsonFactory);    
     }
 
     bool IsEmpty()
--- a/include/effectfactories.h
+++ b/include/effectfactories.h
@@ -116,6 +116,7 @@ class EffectFactories
     {
         defaultFactories.emplace_back(effectNumber, defaultFactory);
 
+        // BUGBUG perf (davepl) Consider using 'jsonFactories.try_emplace without the first lookup
         if (jsonFactories.count(effectNumber) == 0)
             jsonFactories[effectNumber] = jsonFactory;
     }
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -121,7 +121,7 @@ public:
         construct(false);
     }
 
-    EffectManager(std::vector<std::shared_ptr<GFXBase>>& gfx)
+    explicit EffectManager(std::vector<std::shared_ptr<GFXBase>>& gfx)
         : _gfx(gfx)
     {
         debugV("EffectManager Constructor");
@@ -527,7 +527,8 @@ public:
 
     const bool AreEffectsEnabled() const
     {
-        for (auto& pEffect : _vEffects)
+        // BUGBUG (davepl) Consider using std::any_of algorithm instead of a raw loop
+        for (const auto & pEffect : _vEffects)
             if (pEffect->IsEnabled())
                 return true;
 
@@ -618,8 +619,7 @@ public:
 
     void PreviousPalette()
     {
-        auto g = _gfx[0];
-        g->CyclePalette(-1);
+        g()->CyclePalette(-1);
     }
     // Update to the next effect and abort the current effect.
 
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -606,7 +606,7 @@ public:
 
             debugV("%ldms elapsed: Next Effect", millis() - _effectStartTime);
             NextEffect();
-            debugV("Current Effect: %s", GetCurrentEffectName());
+            debugV("Current Effect: %s", GetCurrentEffectName().c_str());
         }
     }
 
@@ -662,15 +662,15 @@ public:
 
         for (int i = 0; i < _vEffects.size(); i++)
         {
-            debugV("About to init effect %s", _vEffects[i]->FriendlyName());
+            debugV("About to init effect %s", _vEffects[i]->FriendlyName().c_str());
             if (false == _vEffects[i]->Init(_gfx))
             {
-                debugW("Could not initialize effect: %s\n", _vEffects[i]->FriendlyName());
+                debugW("Could not initialize effect: %s\n", _vEffects[i]->FriendlyName().c_str());
                 return false;
             }
-            debugV("Loaded Effect: %s", _vEffects[i]->FriendlyName());
+            debugV("Loaded Effect: %s", _vEffects[i]->FriendlyName().c_str());
         }
-        debugV("First Effect: %s", GetCurrentEffectName());
+        debugV("First Effect: %s", GetCurrentEffectName().c_str());
         return true;
     }
 
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -527,12 +527,7 @@ public:
 
     const bool AreEffectsEnabled() const
     {
-        // BUGBUG (davepl) Consider using std::any_of algorithm instead of a raw loop
-        for (const auto & pEffect : _vEffects)
-            if (pEffect->IsEnabled())
-                return true;
-
-        return false;
+        return std::any_of(_vEffects.begin(), _vEffects.end(), [](const auto& pEffect){ return pEffect->IsEnabled(); } );
     }
 
     const size_t GetCurrentEffectIndex() const
--- a/include/effects/matrix/PatternPongClock.h
+++ b/include/effects/matrix/PatternPongClock.h
@@ -131,8 +131,8 @@ class PatternPongClock : public LEDStripEffect
 
     virtual void Draw() override
     {
-        time_t ttime = time(0);
-        tm *local_time = localtime(&ttime);
+        const time_t ttime = time(0);
+        const tm *local_time = localtime(&ttime);
 
         g()->Clear();
 
--- a/include/effects/matrix/PatternPongClock.h
+++ b/include/effects/matrix/PatternPongClock.h
@@ -149,7 +149,11 @@ class PatternPongClock : public LEDStripEffect
         }
 
         LEDMatrixGFX::backgroundLayer.setFont(gohufont11b);
-        char buffer[3];
+
+        // The compiler warns that with a nul terminator, 4 bytes could be needed, which is true
+        // but we're only writing 3 bytes, but I'll waste the byte to avoid the warning.
+
+        char buffer[4];
 
         auto clockColor = rgb24(255,255,255);
         sprintf(buffer, "%2d", hours);
--- a/include/effects/matrix/PatternWeather.h
+++ b/include/effects/matrix/PatternWeather.h
@@ -430,7 +430,7 @@ public:
         // Figure out which day of the week it is
 
         time_t today = time(nullptr);
-        tm * todayTime = localtime(&today);
+        const tm * todayTime = localtime(&today);
         const char * pszToday = pszDaysOfWeek[todayTime->tm_wday];
         const char * pszTomorrow = pszDaysOfWeek[ (todayTime->tm_wday + 1) % 7 ];
 
--- a/include/effects/matrix/PatternWeather.h
+++ b/include/effects/matrix/PatternWeather.h
@@ -167,7 +167,7 @@ private:
 
         if (httpResponseCode <= 0)
         {
-            debugW("Error fetching coordinates for location: %s", configLocation);
+            debugW("Error fetching coordinates for location: %s", configLocation.c_str());
             http.end();
             return false;
         }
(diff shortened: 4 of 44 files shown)


include/effects/matrix/spectrumeffects.h
```diff
@@ -189,16 +189,17 @@ class SpectrumAnalyzerEffect : public LEDStripEffect, virtual public VUMeterEffe
 {
   protected:
 
-    bool      _bShowVU;
```
davepl, line 192: You'll see a lot of simple reordering of member vars to match the inheritance order of constructors and so on

include/ledstripeffect.h
```diff
@@ -110,8 +110,8 @@ class LEDStripEffect : public IJSONSerializable
         int colorIndex = 0;
 
         auto componentsArray = src.as<JsonArrayConst>();
```
davepl, line 112: value was hiding an out variable named value

davepl: outer, not out

include/effectmanager.h
```diff
@@ -527,11 +527,7 @@ class  EffectManager : public IJSONSerializable
 
     const bool AreEffectsEnabled() const
     {
-        for (auto& pEffect : _vEffects)
-            if (pEffect->IsEnabled())
-                return true;
-
-        return false;
+        return std::any_of(_vEffects.begin(), _vEffects.end(), [](const auto& pEffect){ return pEffect->IsEnabled(); } );
```
davepl, line 530: OK, maybe this is a bit too clever, but it's short!

include/effectfactories.h
```diff
@@ -115,9 +115,7 @@ class EffectFactories
     void AddEffect(int effectNumber, const DefaultEffectFactory& defaultFactory, const JSONEffectFactory& jsonFactory)
     {
         defaultFactories.emplace_back(effectNumber, defaultFactory);
-
-        if (jsonFactories.count(effectNumber) == 0)
-            jsonFactories[effectNumber] = jsonFactory;
+        jsonFactories.try_emplace(effectNumber, jsonFactory);
```
davepl, line 118: I assume the possible outcomes are 0 or 1 for existing factories, but please vet my assumption.  I mean it's a map, so... try_emplace should do the trick!

platformio.ini
```diff
@@ -19,21 +19,30 @@
...
 ; ==================
 ; Base configuration
 ;
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     = 
-monitor_port    = 
+upload_port     = /dev/cu.usbserial-8411301
+monitor_port    = /dev/cu.usbserial-8411301
```
rbergen, line 31: I think we really should clear default_envs, upload_port and monitor_port.

src/main.cpp
```diff
@@ -573,8 +575,7 @@ void loop()
                 strOutput += str_sprintf("Buffer: %d/%d, ", bufferManager.Depth(), bufferManager.BufferCount());
             #endif
 
-            auto& taskManager = g_ptrSystem->TaskManager();
-            strOutput += str_sprintf("CPU: %03.0f%%, %03.0f%%, FreeDraw: %4.3lf", taskManager.GetCPUUsagePercent(0), taskManager.GetCPUUsagePercent(1), g_Values.FreeDrawTime);
+            strOutput += str_sprintf("CPU: %03.0f%%, %03.0f%%, FreeDraw: %4.3lf", g_ptrSystem->TaskManager().GetCPUUsagePercent(0), g_ptrSystem->TaskManager().GetCPUUsagePercent(1), g_Values.FreeDrawTime);
```
rbergen, line 578: This change genuinely interests me. Why did you choose to replace the one reference variable with two invocations of the getter?

rbergen (commented):
There's a few changes I'm not exactly sure about. One could call those a matter of personal preference, so one could reason either way.

The one thing I think that does need to be fixed is getting rid of some of your local values in platformio.ini. ;)

include/effectfactories.h
davepl: I believe try_emplace is atomic; there's a race condition in the first version because it's done in two steps.  Is that a real concern?  Not here, no.

src/main.cpp
davepl: There was somethig lint on my laptop didn't like about it.  But it passes cppcheck and looks fine to me as it was, so I've put it back!

platformio.ini
davepl: I think it should be done automatically to compensate for human frailties.  DIdn't we already strip the env?  Can't we do all three?

src/main.cpp
davepl: Ah, now I remember.  I can make it a const auto reference, since all the member calls are const.  So now I have.

include/effectmanager.h
GPSBabelDeveloper: Optimizers love std::algorithm code like this because it completely explains the intent. Vectorizers munch on that kind of code.

It took me a while to warm up to it, too.

include/effects/matrix/PatternWeather.h
```diff
@@ -167,7 +167,7 @@ class PatternWeather : public LEDStripEffect
 
         if (httpResponseCode <= 0)
         {
-            debugW("Error fetching coordinates for location: %s", configLocation);
+            debugW("Error fetching coordinates for location: %s", configLocation.c_str());
```
GPSBabelDeveloper, line 170: Is it worth adding stream operators so we can use String arguments naturally? You caught a lot of these, but it's an easy mistake to make.

As cringy as it is for a variadic macro, maybe we catch the case of the second argument being a String and just offer a specialzation for it. I wouldn't get crazy and do it for all arguments, but the debugX("blah %s blah", String) is such a common thing that a safety check - or a hard error if we can figure that out - may be worth it.

include/ledmatrixgfx.h
```diff
@@ -126,10 +126,11 @@ class LEDMatrixGFX : public GFXBase
 
     virtual uint16_t xy(uint16_t x, uint16_t y) const override
     {
-        if (x >= 0 && x < _width && y >= 0 && y < _height)
+        // Note the x,y are unsigned so can't be less than zero
+        if (x < _width && y < _height)
```
GPSBabelDeveloper, line 130: When porting code from other systems, this means each effect ends up implementing its own xy and drawpixels that mask because the systems ones (unique to us and unlike smartmatrix or marc merlin's or soulmate or ....) have to be avoided.

I know this isn't changing in this submit. Just pointing out that it's a pain point to effects developers.

include/ledstripeffect.h
```diff
@@ -110,8 +110,8 @@ class LEDStripEffect : public IJSONSerializable
         int colorIndex = 0;
 
         auto componentsArray = src.as<JsonArrayConst>();
-        for (auto value : componentsArray)
-            colors[colorIndex++] = value.as<CRGB>();
+        for (auto v : componentsArray)
```
GPSBabelDeveloper, line 113: Would a & eliminate a construct/copy here?

(It annoys me that 'auto' can't really be trusted here to do the right thing - it would be So Clean....)

platformio.ini
```diff
@@ -30,10 +30,19 @@ default_envs =
...
 build_flags     = -std=gnu++17
                   -Dregister=                       ; Sinister: redefine 'register' so FastLED can use that keyword under C++17
 
+;                 -Wall                             ; Extra useful warnings - uncomment to enable and fix anything you find in OUR code!
+;                 -Wformat=2                        ;   but you'll set many warnings in the libraries, so don't enable this in the build
+;                 -Wformat-truncation               ;   until you're ready to fix them all :-)
+;                 -Wstack-usage=4096           
+;                 -g3
+;                 -Os
+;                 -ffunction-sections
```
GPSBabelDeveloper, line 43: Consider leaving the -ffoo-secctions and -Os turned on in the default builds. They cost a fraction of a second at link time, but save .text and .data size by discarding library code we don't call. Flash is a precious resource and economizing it worth it.

My tree has these turned on. No observed problems.

GPSBabelDeveloper (approved):
Nice cleanup

include/ledmatrixgfx.h
davepl: I've read that several times and each time I'm less sure what you mean!  Effects don't need to implement xy, they consume it.  It's handled by the virtual implementation on the GFXBase class instance.

Can you explain more?  Or differently?  Or type slower so I can understand? :-)

include/ledstripeffect.h
davepl: I'll go you one better with const auto & for the win!

include/ledmatrixgfx.h
GPSBabelDeveloper: Sorry. I can be scatter-brained. You're correct, it's not xy() itself that gets reimplemented; it's everything using it, which is a lot. (I also see that I'm logged into the "wrong" account for this project, but I suspect you recognize this is Robert)

If you're porting an effect from another environment (let's say Animatrix, though it's just an example), that code can call drawCircle, drawLine or whatever with X and Y that are "off screen" with immunity. No wrapping is done and no error is enforced; it's simply masked. (clamp or integer modulo would be fast ways to implement it) It's not an error case. I thus wound up implementing (copy-pasting) a lot of drawCircle-style implementations because the system one would fall into this throw. There were a few that called drawLine near the edges that would similarly fall into this throw.

My request here is to make this a debugX warning instead of a hard error. (Clamping it so you have a superpixel in one corner isn't great, but it's at least a hint to the effect developer.) In xy() it's only annoying, but in the drawing primitives, it's a big deal.

include/effects/matrix/PatternPongClock.h
```diff
@@ -149,7 +149,11 @@ class PatternPongClock : public LEDStripEffect
         }
 
         LEDMatrixGFX::backgroundLayer.setFont(gohufont11b);
-        char buffer[3];
+
+        // The compiler warns that with a nul terminator, 4 bytes could be needed, which is true
+        // but we're only writing 3 bytes, but I'll waste the byte to avoid the warning.
```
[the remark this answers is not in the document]
davepl: In Ada you could set the range on the type!

include/effects/matrix/PatternWeather.h
davepl: For now I've enabled the warning that will result if you try to do this wrong!

include/gfxbase.h
```diff
@@ -426,8 +426,8 @@ class GFXBase : public Adafruit_GFX
           We fill pixel with .75 worth of color
         */
 
-        uint8_t fade1 = (std::max(frac1, 1.0f - count)) * 255; // Fraction is how far past pixel boundary we are (up to our total size) so larger fraction is more dimming
-        uint8_t fade2 = (1.0 - frac2) * 255;                   // Fraction is how far we are poking into this pixel, so larger fraction is less dimming
+        uint8_t fade1 = (uint8_t) (std::max(frac1, 1.0f - count)) * 255; // Fraction is how far past pixel boundary we are (up to our total size) so larger fraction is more dimming
+        uint8_t fade2 = (uint8_t) (1.0f - frac2) * 255;                   // Fraction is how far we are poking into this pixel, so larger fraction is less dimming
```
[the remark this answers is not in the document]
davepl: Done

davepl:
Find a cool effect that needs it, I’ll gladly change it!

- Dave

platformio.ini
davepl: Done!

include/effectmanager.h
rbergen: The name of the function itself does. I just find that the semantics of the specification of the iterator begin and end, and the ratio of the lambda fluff around the _actual_ "thing" (is it enabled or not) distracting. But I already stated I'm happy to accept this as a matter of personal preference.

Maybe the underlying question is what level of comfort with this form of C++ we expect people to have if they want to contribute to the core of the project. I'm definitely not going to start that discussion, though. :)

include/effects/matrix/PatternWeather.h
rbergen: > As cringy as it is for a variadic macro, maybe we catch the case of the second argument being a String and just offer a specialzation for it.

I don't think we should, if I'm honest. If you do that for the "one string argument" case but not for the others, then it is even easier to forget the .c_str() for the second, third, ... String argument. Figuring out why the plain String does work for the first and not the others will then also be a very interesting exercise for the uninitiated - or the Rutger who comes back to his own code after a week or two.

> For now I've enabled the warning that will result if you try to do this wrong!

I'd say that is indeed the best thing to do for now.

include/effects/matrix/PatternPongClock.h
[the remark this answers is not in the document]
rbergen: Are you proposing a reimplementation? ;)

platformio.ini
rbergen: In CI we do clear default_envs and we could clear the ports, yes. For CI we don't even have to clear the ports, because they're not used in the CI flow. Any changes CI makes to the code it pulls from the repo are not persisted in the codebase though, so the code in the repo is and remains what you (the contributor) push to it.

I'm sure it would be possible to put a GitHub workflow in place that follows up a push to the code with an additional commit that removes or replaces certain lines, but that's a form of automation I'm not yet familiar with. There are some obvious pitfalls there; for one you wouldn't want the automated commit to trigger an endless loop of commits. I'm sure there are ways to navigate around that though.
However, it also just feels weird to outdate the local tree of a contributor the same second they push to the repo - which is what an automated commit like this would do.

Another option would be to use the same approach with platformio.ini as we did with secrets.h. That is, add platformio.ini to .gitignore and include a platformio.ini "template" instead. The issue with that is that the local platformio.ini gets out of sync "without warning" whenever the template is updated.

rbergen (approved):
Gave it another look and concluded I'm happy to approve, duly considering all the discussions that have taken place. As has already been said: good clean-up!

PR #384, opened 2023-07-29
Cleanup excessive virtual function decoration.

robertlipe:
Since C++11, Ex0actly one of virtual or overide should be used.
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override0x

## Contributing requirements

* [ ] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [ ] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [ ] I selected `main` as the target branch.
* [ ] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -112,7 +112,7 @@ class DeviceConfig : public IJSONSerializable
 
     DeviceConfig();
 
-    virtual bool SerializeToJSON(JsonObject& jsonObject) override
+    bool SerializeToJSON(JsonObject& jsonObject) override
     {
         return SerializeToJSON(jsonObject, true);
     }
@@ -138,7 +138,7 @@ class DeviceConfig : public IJSONSerializable
         return jsonObject.set(jsonDoc.as<JsonObjectConst>());
     }
 
-    virtual bool DeserializeFromJSON(const JsonObjectConst& jsonObject) override
+    bool DeserializeFromJSON(const JsonObjectConst& jsonObject) override
     {
         return DeserializeFromJSON(jsonObject, false);
     }
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -183,7 +183,7 @@ public:
     //
     // Lastly, the function calls the construct() method, indicating successful deserialization.
 
-    virtual bool DeserializeFromJSON(const JsonObjectConst& jsonObject) override
+    bool DeserializeFromJSON(const JsonObjectConst& jsonObject) override
     {
         ClearEffects();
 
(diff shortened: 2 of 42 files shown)


include/deviceconfig.h
```diff
@@ -112,7 +112,7 @@ class DeviceConfig : public IJSONSerializable
 
     DeviceConfig();
 
-    virtual bool SerializeToJSON(JsonObject& jsonObject) override
```
davepl, line 115: I did not know it was preferred to remove the virtual decoration in overrides. As you can tell! Now I know. :-)

davepl (approved):
LGTM!  This was either a tagged regular expression or a lot of work :-). I wonder if there's any binary difference or if it's just compiler semantics?

include/deviceconfig.h
GPSBabelDeveloper: Oh, I know totally how it happens as I've been there. If you have code (or a programmer) that predates C++11, you have the "virtual Foo" that slots into the over the abstract code. Then C++ 11 comes along and you spray an "override" on it because you've bitten by that ONE virtual method that had a slightly mismatched signature and wouldn't get called by the abstract class and took you a day to fine, so you condition yourself for the override. But we all (observed in a large group of very smart people) forget to take the virtual's out.

To your question, I'm very good with one-liners in vim, perl, sed, and beating stones together. The process of doing THIS one admittedly frustrated me enough to capture the pio build output into a 'compile_commands.json' that I then used to build a CMakeLists.txt that worked just well enough to let me load it into CLion (free for open source maintainer) which happens to be very good at this kind of that. That spooled me up for the CL that followed this one (It does spelling and grammar check, too - and no, I didn't grind the boot on every nit) so while that CLion build isn't good enough to build/run/debug, it's good enough to EDIT in a way that conveniently brings clang-tidy (whose modernizer could have been used for THIS one, but wasn't) output right into the editor with convenient little "fix" light bulbs next to the code under the microscope.

So i started off by mining my own ore and refining it into a toolsmith and end the end had a decent blacksmithing shop.

My former employer had Very Good tools for doing this kind of thing.

GPSBabelDeveloper:
Aha! It has been discussed publicly, and not in a leak. I miss Rosie.

https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext

I'm sure you were using better stuff than MS VSS by the end, right? I don't know why our hobbyist projects can't have ALL the nice toys of the multi-hundred billion companies...but honestly, the infrastructure for free/open software is pretty amazing these days with the ability to spin up VMs for each commit and run a full regression under simulation & valgrind or whatever.

rbergen:
LGTM too!

PR #386, opened 2023-07-29
Fix linedraw bug

davepl:
I found a case where PatternCube was getting bit by the watchdog, and it turned out BresenhamLine was never returning.  That turns out to be a bug in the base library we inherited.  Instead of accumulating the error over several passes, it was resetting it every time, and that doesn't always work.  Now it does.

Also fixes an issue with SpectrumBarEffect where it would display all-same-height bars.  I believe this was related to the decay rates not being set.  If it recurs, I'll reopen the issue.

I'm also enabling a heap check after each thread is created.  We can put this in ifdef DEBUG if preferred.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -179,6 +179,7 @@ DRAM_ATTR std::mutex g_buffer_mutex;
 // The one and only instance of ImprovSerial.  We instantiate it as the type needed
 // for the serial port on this module.  That's usually HardwareSerial but can be
 // other types on the S2, etc... which is why it's a template class.
+
 ImprovSerial<typeof(Serial)> g_ImprovSerial;
 
 // If an insulator or tree or fan has multiple rings, this table defines how those rings are laid out such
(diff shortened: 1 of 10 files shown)


src/main.cpp
```diff
@@ -555,7 +540,7 @@ void loop()
             strOutput += str_sprintf("LED FPS: %d ", g_Values.FPS);
 
             #if USE_STRIP
-                strOutput += str_sprintf("LED Bright: %03.0f%%, LED Watts: %d, ", g_Values.Brite, g_Values.Watts);
+                strOutput += str_sprintf("LED Bright: %d, LED Watts: %d, ", g_Values.Watts, g_Values.Brite);
```
rbergen, line 543: I see you merged this PR while I was writing this, but you seem to be undoing a change here that you implemented yesterday. I wonder if that was intentional, as the change of yesterday looked like a fix to me.

src/main.cpp
```diff
@@ -574,8 +559,8 @@ void loop()
                 auto& bufferManager = g_ptrSystem->BufferManagers()[0];
                 strOutput += str_sprintf("Buffer: %d/%d, ", bufferManager.Depth(), bufferManager.BufferCount());
             #endif
-            
-            const auto & taskManager = g_ptrSystem->TaskManager();
+
+            auto& taskManager = g_ptrSystem->TaskManager();
```
rbergen, line 563: Same here.

PR #388, opened 2023-07-30
Eliminate duplicated test in PatternSpiro

robertlipe:
## Description
This caught my eye while I was running over nearby code. The left and the
right of that || test were identical.
I chased the code through that gist through its origin git history through
git blame and this change went in with a description that was something
like "make more awesomer".

Removing it changes no behaviour. It just removes a nail that's poking
through the floor surface.

Tested: successful compilation on mesmerizer. Should compile everywhere
without change because it was already if (something || that_same_something).

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternSpiro.h
+++ b/include/effects/matrix/PatternSpiro.h
@@ -113,7 +113,7 @@ public:
       CRGB color = graphics->ColorFromCurrentPalette(hueoffset + i * spirooffset, 128);
       graphics->leds[graphics->xy(x2, y2)] += color;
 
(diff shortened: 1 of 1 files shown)


include/effects/matrix/PatternSpiro.h
```diff
@@ -113,7 +113,7 @@ class PatternSpiro : public LEDStripEffect
       CRGB color = graphics->ColorFromCurrentPalette(hueoffset + i * spirooffset, 128);
       graphics->leds[graphics->xy(x2, y2)] += color;
 
-      if ((x2 == MATRIX_CENTER_X && y2 == MATRIX_CENTER_Y) || (x2 == MATRIX_CENTER_X && y2 == MATRIX_CENTER_Y))
```
davepl, line 116: I wonder why they wrote it that way originally?  Weird!

davepl (approved):
LGTM!

PR #392, opened 2023-08-03
Build merged images and use them in Web Installer

### LINKED ISSUE #374: BUG: Web Installer doesn't include support for 8MB images
### Problem

With the merge of #346, the Mesmerizer project on the Mesmerizer board now produces an 8MB flash file. The Web Installer currently only comes with, and only supports, a single manifest template for a 4MB image - that being the one in config/manifest_template.json. This means flashing the Mesmerizer project on the Mesmerizer board via the Web Installer can't work.

### Solution

The following changes will need to be made:

- A second manifest template will have to be created for use by the web installer. This one has to align with the partition setup included in config/partition_custom_8M.csv.
- The Web Installer config, as defined in config/web_projects.json, must be extended to support the optional specification of the manifest template to use for a particular project.
- The Python script that does the groundwork for putting the Web Installer files together (tools/bake_installer.py) needs to be updated to parse and use the new specification mentioned in the previous point to determine which manifest template to use for a particular project.

rbergen:
## Description

This:

- Adds the option to _not_ erase the device when flashing it with the Web Installer.
- Removes the SPIFFS binary from the Web Installer builder, and its manifests. This and the previous point combined makes that a device that has been flashed with NightDriverStrip once, can now be reflashed without losing device or effect settings that have been configured. 
- Extends the "pio run" execution with a step to build a merged image out of the individual binaries that remain.
- Makes the Web Installer flash the merged images by default - this is now the recommended approach according to the [ESP32 Web Tools documentation](https://esphome.github.io/esp-web-tools/).
- Adds an optional device and project level property to web_projects.json, to indicate if merged images should be used for a device/project (`"merge": true`) or not (`"merge": false`). This is in case we find out some devices don't respond well to a merged image being flashed.
- Changes the Web Installer browser application to automatically select the project for devices that only have one defined.
- Updates the README, mainly on using the Web Installer for flashing and WiFi configuration. It now covers a few scenarios I've seen happening while testing.
- Fixes #374.
- Has been tested with an M5StickC Plus/Spectrum and Mesmerizer.

## Contributing requirements


davepl (approved):
Wow, Javascript!  Now I remember why I'm a C programmer :-).  nice scoping changes though!

PR #394, opened 2023-08-05
Address issues raised in issue 390

robertlipe:
Addressed most concerns in https://github.com/PlummersSoftwareLLC/NightDriverStrip/discussions/390

- Fix issue 1 about PIN0 naming.
- Fix issue 2 about DEMO overdrawing by a factor of 8.
- Let device names reflect LED electrical types (HUB75, NeoPixel) and not physical orientation (MATRIX, STRIP). Build SPECTRUM uses NeoPixels, but in a MATRIX, for example. HEIGHT > 1 can be used for "grid-ness" of those builds, but it was never used as such.

Issue 3 on what RINGS are, how a line has five of them, and why it's ever
appropriate to be equal to NUM_LEDS remains open.

All are in separate submits if you want to cherry pick.

I tried to preserve the mixture of DOS and UNIX NL conventions.

## Contributing requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,9 @@ I recommend you do the following:
 - Copy include/secrets.example.h to include/secrets.h; Set your WiFi SSID and password in include/secrets.h.
 - Build the source code. In particular, build the `DEMO` configuration. Some pointers on what's needed to do this can be found [below](#build-pointers).
 - Upload the resultant binary to the ESP32
-- Connect PIN5 and GND and 5V of a WS2812B strip to the ESP32
+- Connect the wire associated with LED_PIN0 in your include/globals.h and GND and 5V of a WS2812B strip to the ESP32. For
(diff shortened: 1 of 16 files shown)


rbergen:
About the name to use instead of STRIP, we could choose WS281X if we want to make it explicit that we support the different (physical) variations of WS2812. That obviously works as long as at the code or dependency level, no distinction between the variations needs to be made.

I think I said myself recently that combining multiple LED types/protocols (HUB75 and any WS281X variation) in one build is asking for heaps of trouble in the current codebase. However, making the choice a hard either/or by creating one define with multiple possible values is effectively an architectural decision to not go down that route anytime soon. Because of that, I'd like @davepl's input on this one as well.

It may just be that things were set up this way because all other feature flags are defined to 1 when switched on, which makes the current approach consistent from that perspective. That said, I do agree that it's not necessarily intuitively obvious that "it's WS281X unless it's HUB75" when both can be individually defined.

davepl (approved):
The problem is we overloaded the naming, so USE_MATRIX means both "matrix effects" and "use a HUB75 matrix".  It'll take me a while to get used to these, but I'll do it!

PR #395, opened 2023-08-05
Wait for WiFi credentials if not set

### LINKED ISSUE #371: BUG: LEDSTRIP project doesn't allow WiFi credentials to be set via Improv
### Problem

The LEDSTRIP project demands WiFi to be connected (`WAIT_FOR_WIFI` is defined as 1), and bails out by rebooting if it isn't. This is handled by [lines 629 to 637 in main.cpp](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/src/main.cpp#L629-L637).
If no WiFi credentials are available at boot, the code in [lines 224 to 228 in network.cpp](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/src/network.cpp#L224-L228) immediately gives up on WiFi connectivity, leading to the device rebooting very early in the boot process.

Although this makes sense if the project is flashed using a method that requires WiFi credentials to be set in secrets.h, it creates a problem if the project is flashed using the web installer, Part of that workflow is the setting of WiFi credentials using Improv. For the LEDSTRIP project that's impossible because the device is stuck in a boot loop by the time the user has a chance to enter their WiFi credentials.

### Solution

If `WAIT_FOR_WIFI` is set to 1, the device should allow enough time for the user to enter WiFi credentials.

### Note

This problem was first reported in issue #367. I'm opening this specific issue at @davepl's request.

rbergen:
## Description

This fixes #371 by waiting for WiFi credentials indefinitely if they are not set and WiFi is required. This is to cater for the case that the LEDSTRIP project is flashed using the Web Installer, which sets the SSID to an empty string. 

If an invalid set of credentials is set, then the device will eventually reboot - as it currently does immediately if no WiFi credentials are available. Due to the 99 attempts with incremental waits, it will take about an hour and a half (5247 seconds) before the code reaches that point. Which gives sufficient opportunity to correct a typo in SSID or password if one realizes one was made.

## Contributing requirements

### DIFF
--- a/include/improvserial.h
+++ b/include/improvserial.h
@@ -48,7 +48,6 @@ extern DRAM_ATTR String WiFi_ssid;
 extern DRAM_ATTR String WiFi_password;
 bool ReadWiFiConfig();
 bool WriteWiFiConfig();
-bool ConnectToWiFi(uint cRetries);
 
 template <typename SERIALTYPE>
 class ImprovSerial
--- a/include/network.h
+++ b/include/network.h
@@ -37,7 +37,7 @@
 #if ENABLE_WIFI
     void processRemoteDebugCmd();
 
-    bool ConnectToWiFi(uint cRetries);
+    bool ConnectToWiFi(uint cRetries, bool waitForCredentials);
     void UpdateNTPTime();
     void SetupOTA(const String & strHostname);
     bool ReadWiFiConfig();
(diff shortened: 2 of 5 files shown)


rbergen:
The only flows that allow WiFi credentials to be set after flashing are those where the Web Installer is used. And when it is, then the device _is_ attached via a cable, and a facility to look at debug output _is_ provided. 

The README includes a pretty comprehensive set of instructions for using the Web Installer, which will get a meaningful update when #392 get merged. These instructions include the configuration of WiFi credentials as part of the flash procedure. 
This PR basically gives a user the time to follow those instructions, instead of the device rebooting within a few seconds after start-up, preventing the WiFi credentials from ever being set.

PR #400, opened 2023-08-07
Allow effects to be loaded in disabled state, add/update license headers

rbergen:
## Description

This makes it possible to add effects to the default effect list in a disabled state, meaning they will not be included in the visualization cycle until they are explicitly enabled.

For this, two new macros have been defined that can be used in the `LoadEffectFactories()` function in effects.cpp. These are `ADD_EFFECT_DISABLED` and `ADD_STARRY_NIGHT_EFFECT_DISABLED`. The parameters to these macros match their "regular" counterparts exactly.

It also adds and updates the license header in a number of files.

## Contributing requirements


rbergen:
I've updated the PR to make the separation of concerns between the `NumberedFactory` and `EffectManager` cleaner, in the sense that the code now guarantees that the `LoadDisabled` flag on the `NumberedFactory` is respected when the effect is created using the factory function.

rbergen:
Merging after testing on Mesmerizer

PR #402, opened 2023-08-09
[Quality of Developer Life] Add a new telnet/serial debugger command: uptime.

robertlipe:
I've had this in my tree for a while, but while soaking the new effects,
it's actually justified its bits per blinken, IMO.

Help text:
(processRemoteDebugCmd)(C1) uptime              Show system uptime, reset reason

Sample invocation:
uptime
* Debug: Command received: uptime
(I) (ShowUptime)(C1) Uptime: 0 days - 00:04:50
(I) (ShowUptime)(C1) Last boot reason: Power On
(I) (loop)(C1) Here
(I) (esp32_partition_table_read)(C1) Reading partitions
(I) (esp32_partition_table_read)(C1) nvs 36864 8192

It's quite ESP32-specific in some ways and it's definitely force-fit
into ntpclient (hey, it's gotta be somewhere, right?) but I really don't
think it's any weirder or less justified that a lot of the existing
code.

I had a version of this that displayed the uptime in an overlay, but it
destroyed the FPS. If someone is interested in pair-programming a
follow-up to this PR that adds that, I'm down.

## Contributing requirements

* [x] I make up random text and append it to the forms just to test validation.

### DIFF
--- a/src/ntptimeclient.cpp
+++ b/src/ntptimeclient.cpp
@@ -151,16 +151,11 @@ bool NTPTimeClient::UpdateClockFromWeb(WiFiUDP * pUDP)
     return true;
 }
 
-// This is underdesigned for general use. It's very specifically a
-// throwaway meant only for debugging. It's not really NTP related.
-//
-// It contains some ESP32-specific goo, but that's allowed elsewhere.
+// Not NTP related. Convenience utility for debugging. Callable from gdb
+// serial shell, or network log debugger.
 
 void NTPTimeClient::ShowUptime()
 {
-    // I hate C99 time handling so much, but C17's <chrono> isn't likely
-    // worth the lift.
-
     struct timeval timeval = { 0 };
     // Microseconds since boot. File wrap bugreport in 292 million years.
     auto uptime = esp_timer_get_time();
@@ -189,10 +184,10 @@ void NTPTimeClient::ShowUptime()
         case ESP_RST_DEEPSLEEP: reason_text = "Reset in deep sleep"; break;
         case ESP_RST_BROWNOUT: reason_text = "Brownout"; break;
         case ESP_RST_SDIO: reason_text = "Reset over SDIO"; break;
-	// Documented,  but seemingly not implemented yet.
+	// Documented,  but not defined in ESP-IDF esp_system.h (v5.1.0) yet.
         // case ESP_RST_USB: reason_text = "Reset by USB peripheral"; break;
 	default: reason_text = "Unknown"; break;
     }
-    debugI("Last boot reason: %s", reason_text);
+    debugI("Last boot reason: (%d): %s", reason, reason_text);
 
 }
--- a/src/ntptimeclient.cpp
+++ b/src/ntptimeclient.cpp
@@ -150,3 +150,49 @@ bool NTPTimeClient::UpdateClockFromWeb(WiFiUDP * pUDP)
 
     return true;
 }
+
+// This is underdesigned for general use. It's very specifically a
+// throwaway meant only for debugging. It's not really NTP related.
+//
+// It contains some ESP32-specific goo, but that's allowed elsewhere.
+
+void NTPTimeClient::ShowUptime()
+{
+    // I hate C99 time handling so much, but C17's <chrono> isn't likely
+    // worth the lift.
+
+    struct timeval timeval = { 0 };
+    // Microseconds since boot. File wrap bugreport in 292 million years.
+    auto uptime = esp_timer_get_time();
+
+    timeval.tv_sec = uptime / MICROS_PER_SECOND;
+    // timeval.tv_nsec = uptime % NANOS_PER_SECOND;
+
+    char buf[128];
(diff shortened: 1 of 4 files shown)


src/ntptimeclient.cpp
```diff
@@ -150,3 +150,49 @@ bool NTPTimeClient::UpdateClockFromWeb(WiFiUDP * pUDP)
 
     return true;
 }
+
+// This is underdesigned for general use. It's very specifically a
+// throwaway meant only for debugging. It's not really NTP related.
+//
+// It contains some ESP32-specific goo, but that's allowed elsewhere.
```
rbergen, line 157: I don't think you need to label this as a throwaway. We've recently concluded that this is a developer-oriented project "with one exception", so a developer-oriented feature is a feature.

I'm also not sure it's necessary to include "defensive ponderings" about implementation decisions in the source code comments. Where appropriate, comments about implementation decisions can be added to the PR when it's opened.
(I'm aware there may be other places in the code where these comments exist, but up to this point they've been grandfathered as artifacts from before the project was made publicly available.)

src/ntptimeclient.cpp
```diff
@@ -150,3 +150,49 @@ bool NTPTimeClient::UpdateClockFromWeb(WiFiUDP * pUDP)
...
+
+// This is underdesigned for general use. It's very specifically a
+// throwaway meant only for debugging. It's not really NTP related.
+//
+// It contains some ESP32-specific goo, but that's allowed elsewhere.
+
+void NTPTimeClient::ShowUptime()
+{
+    // I hate C99 time handling so much, but C17's <chrono> isn't likely
+    // worth the lift.
```
rbergen, line 162: See second part of previous comment.

src/ntptimeclient.cpp
```diff
@@ -150,3 +150,49 @@ bool NTPTimeClient::UpdateClockFromWeb(WiFiUDP * pUDP)
...
+        case ESP_RST_EXT: reason_text = "External Pin"; break;
+        case ESP_RST_SW: reason_text = "Software Restart"; break;
+        case ESP_RST_PANIC: reason_text = "Panic"; break;
+        case ESP_RST_INT_WDT: reason_text = "Watchdog barked"; break;
+        case ESP_RST_TASK_WDT: reason_text = "Task Watchdog barked"; break;
+        case ESP_RST_WDT: reason_text = "Other Watchdog barked"; break;
+        case ESP_RST_DEEPSLEEP: reason_text = "Reset in deep sleep"; break;
+        case ESP_RST_BROWNOUT: reason_text = "Brownout"; break;
+        case ESP_RST_SDIO: reason_text = "Reset over SDIO"; break;
+	// Documented,  but seemingly not implemented yet.
```
rbergen, line 192: That's interesting! How did you establish it's not implemented yet?

While writing this, the question comes up if it would make sense to still include the case anyway? If it's not implemented it won't break anything, and this will "magically" start working if and when it is.

rbergen (commented):
Nice, I like it! 

@robertlipe I have a few thoughts about the comments, but certainly nothing substantial enough to block a merge. Let me know what you think; again, I'm happy to approve regardless.

rbergen:
CI fails because the PlatformIO Core installer script, as provided by PlatformIO, seems to have gone missing. I opened https://github.com/platformio/platformio-core/issues/4715 to report this.

ivankravets:
>         python -c "$(curl -fsSL https://raw.githubusercontent.com/platformio/platformio/master/scripts/get-platformio.py)"
>        echo "/home/runner/.platformio/penv/bin" >> $GITHUB_PATH

You don't need to use the PlatformIO Core installer in the CI instances. 

See docs https://docs.platformio.org/en/latest/integration/ci/github-actions.html

Just use 

```
      - name: Install PlatformIO Core
        run: pip install --upgrade platformio
```

rbergen (approved):
@robertlipe Thanks for your changes. LGTM and with CI back in working order, I'll proceed to merge.

ivankravets:
@rbergen, thanks! I've just updated the docs https://github.com/platformio/platformio-docs/compare/6b2af96f74d1c99f934a6975d9c26a4325abd172...eca1ad5259b39553bbbab4d7c71902089fd9a87e#diff-7472577066642c8cf999095c9b575e362a6315d0d69c50685dd175752fc5f42b

The installer script is recommended for developers on personal machines where `pip` is system dependent/global. Installing `platformio` to the global/system scope is not a good idea. This is why the installer script is the right solution that installs PIO Core into the isolated environment.

Nevertheless, the CI machines are stateless. So, using `pip` is a good idea and saves the configuration time. However, I agree with you. Both options provide the same result. So, it is up to you which method to use. In our company, we use `pip` on the CI instances.

rbergen:
@ivankravets Great, that's a very quick turnaround time! :)

Thanks for your explanation, I understand the distinction between CI and local. After bumping our noses against a number of issues related to differences between environments, we try to keep CI as close to local as possible. So, as long as the installer script is the recommended approach for local, then we'll also use that for CI.

PR #406, opened 2023-08-15
Document various 'demo' builds

robertlipe:
## Description

Make it easier for a first-time contributor to recognize that the best 'demo' for their hardware may not actually be named 'demo' .

## Contributing requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ That simplest configuration, called here simply 'DEMO', is provided by a board s
 - lilygo-tdisplay-s3-demo
 
 These [build types](#build-pointers) may be chosen by the '-e' argument
-to pio or in a menu option inside PlatformIO/Visual Studio.
+to pio or in a menu option inside the PlatformIO IDE/VS Code.
 
 Concerning JSON peristence: the effects table is persisted to a JSON file on SPIFFS at regular intervals, to retain the state of effects (and in fact the whole effect list) across reboots. This is largely in preparation for future updates to NightDriverStrip, where the composition of the effect list configuration of individual effects can be changed using the device web application. The API endpoints to facilitate this are already available and ready for use (see [Device web UI and API](#device-web-ui-and-api), below.)
 
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ Each channel of LEDs has an `LEDStripGfx` instance associated with it. `_GFX[0]`
 
 The simplest configuration, `DEMO`, assumes you have a single meter strip of 144 LEDs and a power supply connected to your ESP32. It boots up, finds a single `RainbowFillEffect` in the `LoadEffectFactories()` function, and repeatedly calls its `Draw()` method to update the CRGB array before sending it out to the LEDs. If working correctly it should draw a scrolling rainbow palette on your LED strip.
 
-That simplest configuration, called here simply 'DEMO', is provided by a board specific buid environment. The list of environments can be seen by running 'python3 tools/show_envs.py', which would tell the reader that of this writing, hardware specific vriations of 'DEMO' include:
+That simplest configuration, called here simply 'DEMO', is provided by a board specific buid environment. The list of such environments can be seen by running 'python3 tools/show_envs.py', which would tell the reader that, as of this writing, hardware specific variations of 'DEMO' include:
 
 - demo
 - m5demo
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -197,6 +197,19 @@ Each channel of LEDs has an `LEDStripGfx` instance associated with it. `_GFX[0]`
 
 The simplest configuration, `DEMO`, assumes you have a single meter strip of 144 LEDs and a power supply connected to your ESP32. It boots up, finds a single `RainbowFillEffect` in the `LoadEffectFactories()` function, and repeatedly calls its `Draw()` method to update the CRGB array before sending it out to the LEDs. If working correctly it should draw a scrolling rainbow palette on your LED strip.
 
+That simplest configuration, called here simply 'DEMO', is provided by a board specific buid environment. The list of such environments can be seen by running 'python3 tools/show_envs.py', which would tell the reader that, as of this writing, hardware specific variations of 'DEMO' include:
```
rbergen, line 200: I think "build" misses its lower-case L.

```suggestion
That simplest configuration, called here simply 'DEMO', is provided by a board specific build environment. The list of such environments can be seen by running 'python3 tools/show_envs.py', which would tell the reader that, as of this writing, hardware specific variations of 'DEMO' include:
```

README.md
```diff
@@ -197,6 +197,19 @@ Each channel of LEDs has an `LEDStripGfx` instance associated with it. `_GFX[0]`
...
+- demo
+- m5demo
+- m5plusdemo
+- heltecdemo
+- heltecv2demo
+- heltecv3demo
+- lilygo-tdisplay-s3-demo
+
+These [build types](#build-pointers) may be chosen by the '-e' argument
+to pio or in a menu option inside PlatformIO/Visual Studio.
```
rbergen, line 211: I'd rephrase "PlatformIO/Visual Studio" to "the PlatformIO IDE/VS Code". The former is what the PlatformIO project mentions in their own documentation. The latter change reflects that Visual Studio Code really is something different than Visual Studio, and adopts the colloquial name most people use for it.

rbergen (commented):
Thanks for this! Documentation improvements are always welcome.

I do have two small suggestions, see below. These are just to make this documentation upgrade is as good as it can be. :)

rbergen:
> Commit at will. (And, in general, if you have a suggestion that you don't think I'm likely to argue about, if it's less work for you, please always feel free to Just Fix It.)

I do make a case-by-case decision on what to suggest and what to Just Fix, although often implicitly. It's not just a matter of speed/ease though. For one, I don't want to make too many assumptions about what you, or others, are likely to argue about. For another, it relates to something that really matters to me, which is sense of ownership. If you consciously embrace a change I suggest then we own the result together.

As it stands this change LGTM now, so I'll kick off CI just because that's how we roll and merge after. Thanks again!

rbergen:
> Oh, and presubmit checking is being naughty, unable to find 'pio' so they're all failing. I looked and it seems to be resetting the path, but it's outside my real area of knowledge.

I think you may have to sync up (some of) your branches with the latest changes in PlummersSoftwareLLC:drag-race, or otherwise pull in its version of [.github/workflows/CI.yml](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/.github/workflows/CI.yml), _after this PR has been merged_. I fixed the "PR CI workflow" yesterday in the context of #402, but missed the one that builds and publishes the web installer. I just pushed a commit to this PR to fix that one as well.

rbergen:
Looks like "PR CI" is back in working order in your [robertlipe:robertlipe-patch-1-2](https://github.com/robertlipe/NightDriverStrip/tree/robertlipe-patch-1-2) branch after my latest commit and push to it. That .github/workflows/CI.yml version is the one that should get things back in working order for other branches also.

The web installer flow will run after I merge this. I'll let you know if that is now also fixed, but it does take well over an hour to complete - so it'll be at least that long before I can notify you. 

And this is not your fault, nor mine. Sometimes other people break stuff for us - in this case it was the PlatformIO project itself. (That's not an accusation by the way, just an observation. :) )

rbergen:
> I worked on a tippy-top node of one of the world's largest "live at $HEAD"
projects. This is all comparatively pretty sleepy. :-) The equivalent of PR
numbers used to go up by a thousand or more a minute.

A thousand or more PR equivalents a minute I never experienced. But one of the other projects in this GitHub organization was the second-most-trending repo for a short time, and at that time a fellow maintainer and I were reviewing and merging up to a dozen PRs a day. There, CI more or less depended on other people getting things right for whatever programming language they submitted for/in, and managing that was a fun challenge as well. Even more so when one language kept randomly but consistently breaking its own builds (I won't mention any names because it was V).

> I don't think I have another hour left in me right now (It's 04:44.)

I know. That was me saying "please don't wait up". :) 
Turns out you don't have to either. The web installer flow already passed the point where it previously failed, so I'm confident the breakage we had is behind us.

rbergen:
Final confirmation: post-merge CI completed successfully. I'd say this case is closed.

PR #414, opened 2023-08-21
Make USE_REMOTE and USE_BUTTONS_ more atomic

robertlipe:
- Make USE_REMOTE and USE_BUTTON more self-contained.
- Missing file from previous commit.

## Description
    IMO, this change stands on its own as just good style of constraining
    the code that uses remote into the files and blocks that use remote.

    More motivationally, the remote code is Really Annoying to build. It's
    a lot of files to compile in the hope that -ffunction-sections will
    actually throw them away. Better to simply not compile them. This makes
    that possible.

    With this change, it's possble to lib_ignore Bounce2 and IRremoteESP8266
    (and U8g2 and QRcode...) these modules, though I haven't reconciled how
    to do so against our recommended document and style. As such, the change
    that REALLY takes advantage of this,
    +lib_ignore  = U8g2
    +       Bounce2
    +       IRremoteESP8266
    +       QRCode
    +board = esp32-s3-devkitc-1-n16r8v

    is not included here. I think this change stands on its own and lays
    bricks for that later change.

    Tested: successful buddybuild.

## Contributing requirements

Same blinken. Fewer bits. Positive improvement!

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -1570,7 +1570,7 @@ inline bool SetSocketBlockingEnabled(int fd, bool blocking)
 #include "ledviewer.h"                          // For the LEDViewer task and object
 #include "network.h"                            // Networking
 #include "ledbuffer.h"                          // Buffer manager for strip
-#if TOGGLE_BUTTON_1 || TOGGLE_BUTTON_2
+#if defined(TOGGLE_BUTTON_1) || defined(TOGGLE_BUTTON_2)
 #include "Bounce2.h"                            // For Bounce button class
 #endif
 #include "colordata.h"                          // color palettes
--- a/include/globals.h
+++ b/include/globals.h
@@ -1570,7 +1570,9 @@ inline bool SetSocketBlockingEnabled(int fd, bool blocking)
 #include "ledviewer.h"                          // For the LEDViewer task and object
 #include "network.h"                            // Networking
 #include "ledbuffer.h"                          // Buffer manager for strip
+#if TOGGLE_BUTTON_1 || TOGGLE_BUTTON_2
 #include "Bounce2.h"                            // For Bounce button class
+#endif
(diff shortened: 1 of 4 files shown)


rbergen:
I did a quick first review, and I think we'll have to stick to the use of `#ifdef` in those places where the PR aims to convert a single `#ifdef` to `#if`. In those places where two defines are tested, we'll have to use `#if defined(TOGGLE_BUTTON_1) || defined(TOGGLE_BUTTON_2)`.
(For consistency, we could choose to use `#if defined()` everywhere - I don't have a strong opinion either way.)

The reason is that 0 (zero) is a valid value for both defines, and either being set to it _should_ include the code after the `#if`.

davepl:
I’m not a fan of defined() because I set the precedent early on of either defining things to ONE or ZERO rather than “defined or not”.  

Remember that:

#if defined (FOO) || defined(BAR)

Is still true if both or either are zero, so it’s potentially confusing to readers.

That’s why I went with 

#if FOO || BAR

…and made sure that both FOO and BAR had to be set to 1 for the condition to be true, not merely defined to any value.

So, where possible, I like things to be 0 or 1 and not rely on whether its been defined at all to any value.

I’ve said all this without closely investigating the context of the specific issue, just speaking in generalities and preferences.

- Dave

rbergen:
I'm with you all the way - if it concerns feature flags. In this case the defines are effectively pin numbers, and they can be 0. That 0 should not translate to "not set".

davepl:
For LED_PINx that makes sense.  You’re right, I was speaking of the feature flags.

rbergen:
I don't really care if the current 0 values for the TOGGLE_BUTTON defines are actually sent to the Bounce2 library. The point is they can be, and the first argument to the button.attach() function is the button pin number, and that can be 0.

If a define is a feature toggle,

PR #441, opened 2023-09-24
Fix bouncing ball, fix AtomLamp

davepl:
Bouncing ball had two bugs - a float that broke when the time got set from NTP, and then a static ballcount locking it to 3 that didn't match the real ballcount.
AtomLamp needed a buffer limitation to have enough RAM to run well.


include/effects/strip/faneffects.h
```diff
@@ -1007,7 +1009,8 @@ class FireFanEffect : public LEDStripEffect
         Sparking(jsonObject[PTY_SPARKING]),
         bReversed(jsonObject[PTY_REVERSED]),
         bMirrored(jsonObject[PTY_MIRORRED]),
-        Order((PixelOrder)jsonObject[PTY_ORDER])
+        Order((PixelOrder)jsonObject[PTY_ORDER]),
+        bMulticolor(jsonObject[PTY_MULTICOLOR])
```
rbergen, line 1013: I think we'll have to serialize this in `SerializeToJSON()` as well, if we plan to deserialize it here.

src/effects.cpp
```diff
@@ -560,18 +560,19 @@ void LoadEffectFactories()
...
-        //  make_shared_psram<HueFireFanEffect>(NUM_LEDS, 2, 3, 100, 1, 1, Sequential, true, false, false, HUE_GREEN),
+
+        ADD_EFFECT(EFFECT_STRIP_FIRE_FAN, FireFanEffect, HeatColors_p, NUM_LEDS, 1, 12, 400, 2, NUM_LEDS / 2, Sequential, true, false);
+        ADD_EFFECT(EFFECT_STRIP_FIRE_FAN, FireFanEffect, GreenHeatColors_p, NUM_LEDS, 1, 10, 400, 2, NUM_LEDS / 2, Sequential, true, false);
+        ADD_EFFECT(EFFECT_STRIP_FIRE_FAN, FireFanEffect, BlueHeatColors_p, NUM_LEDS, 1, 10, 400, 2, NUM_LEDS / 2, Sequential, true, false);
+        ADD_EFFECT(EFFECT_STRIP_FIRE_FAN, FireFanEffect, RainbowColors_p, NUM_LEDS, 1, 10, 400, 2, NUM_LEDS / 2, Sequential, true, false);
+        ADD_EFFECT(EFFECT_STRIP_FIRE_FAN, FireFanEffect, HeatColors_p, NUM_LEDS, 1, 10, 400, 2, NUM_LEDS / 2, Sequential, true, false, true);
+
+        ADD_EFFECT(EFFECT_STRIP_BOUNCING_BALL, BouncingBallEffect, 3, true, true, 1);
+
```
rbergen, line 571: Do you think this is a sufficiently significant change to the ATOMLIGHT effect set to force any device already running ATOMLIGHT to reload the standard effect set? In that case we could bump the EFFECT_SET_VERSION for this project to 2.

rbergen (commented):
Looks good to me in general. I think there's one thing that needs to be added, and there's one thing to consider before we merge; see my comments below.

davepl:
Don’t know for sure - but I do know that the code had issues with whatever it found in the non-volatile memory.   I did an erase_flash and then all was well.  I did add a constructor argument, so that could be an issue!

BTW, I’ve seen a case where, under low memory, the serialization code fails, grows its buffer size attempt, and repeats until it crashes.  Likely if two attempts fail, its should bail.  Of course it may not be appropriate to proceed anyway…

- Dave

davepl:
Aligning hsv.  Style choice, not one I’m passionate about.  One the cache locality, maybe.  But you’d be hard pressed to prove there’s a difference, and this effect can do about 1000 frames a second anyway, so…

- Dave

rbergen:
> Don’t know for sure - but I do know that the code had issues with whatever it found in the non-volatile memory.   I did an erase_flash and then all was well.  I did add a constructor argument, so that could be an issue!

I've updated the PR to bump the effect set version for ATOM_LIGHT to 2, so a reset to the default set will happen for anyone who installs this version on a device that already had ATOM_LIGHT installed.

I've also added a line to `FireFanEffect`'s `SerializeToJSON()` to serialize `bMulticolor`.

> BTW, I’ve seen a case where, under low memory, the serialization code fails, grows its buffer size attempt, and repeats until it crashes.  Likely if two attempts fail, its should bail.  Of course it may not be appropriate to proceed anyway…

Any "n" we choose for "bail out after n attempts" is arbitrary. The problem is that there's no way to guess how much memory ArduinoJson needs for a (de)serialization, other than trying. If the serialization code is correct (e.g., doesn't return "false" when there isn't an out-of-memory condition) then the serialization logic will eventually only claim the memory it actually needs. Which may of course just be more than what's available.  
I'd say that if there are low memory conditions/projects we want to support, we could add a build flag to simply disable JSON (de)serialization all together.

rbergen:
It just occurred to me that another scenario that can cause deserialization problems is if a device is flashed with one project, persists its effect set, and then gets flashed with another project using the same partition scheme. I've therefore added the PROJECT_NAME as a property that is serialized with the effects set and checked on deserialization. If the persisted project name and the one of the image don't match, the persisted JSON is discarded.

Issue #411, opened 2023-08-18
Web UI Bugs: reset doesn't reset + other cosmetic glitches


robertlipe:
## Bug report
The reset/restart button in the the web UI doesn't restart remaining count.

**Problem**
Prev and Next work as expected. I'd expec the 'reset' button to reset the remaining time on this effect. It just doesn't.

**Steps**
1.  Click <home> (Why is this step needed? Why isn't 'home' 'home'?
2.  Click on next and prev a few times. Watch current effect cycle. This verifies that your clicking finger works. :-)
3.  Click on reset. Nothing visible changes. I'd expect counter to reset to the cumber on the left.

**Example**
It's not intermittent. It just doesn't seem to be hooked up.

Bonus points to the assignee of this bug for reconsidering if milliseconds really is the appropriate unit of measure to show (without units or a decimal place) and to accept for input on 'interval'. If you type '120' for interval, is it really because you wanted 8 changes per second?

Bonus Bonus bug (maybe this should be three reports...) if you set the interval to '1000', the 'time remaining' field flashes for a cycle or two, but decided it can't keep up and just displays a constant number. Seems it should blank or something. Or maybe we don't allow setting times below a second.

Bonus Bonus Bonue (OK, four reports) if you type a negative number (!) it sets the time to 2^32 - 10000. Time remaining then overlaps the "Previojs" button.

**Notes**
<img width="437" alt="image" src="https://github.com/PlummersSoftwareLLC/NightDriverStrip/assets/8724975/bc2b2829-df38-4ddf-8a38-49774ac7e485">

<img width="321" alt="image" src="https://github.com/PlummersSoftwareLLC/NightDriverStrip/assets/8724975/8b96fbd4-5d9e-492d-9e76-0f75ab87bcc5">

I aborted nothing. What are the other five errors?

rbergen:
TL;DR regarding the main report: the "reset" button behavior is the expected behavior.

The reset button is actually a "refresh" button. The interaction between the UI and the API is currently pull-only, which means that the interval timer updates _and the tracking of what effect is playing_ are largely done in the browser. The actual state is pulled from the device API at some meaningful interval, i.e. seconds. 
Specifically when the web UI browser tab is not shown, this means the device and the UI can/will get out of sync. The refresh button forces the UI to pull the current state of affairs from the device API, to resync what the UI is showing. 

[...]

PR #442, opened 2023-09-25
Feature/config

KeiranHines:
## Description
Added Device and effect level configuration dialog.
* Moved the stats config from the sidebar menu to the top left of the stats bar.
* Added Device level settings button to the left-hand sidebar.
* Changed effect level settings from an options input text box to a generated dialog
* Placeholder support exists for color and palette settings if effects make use of these settings in the future.
* Placeholder support for an "Apply and reboot" option on the effect configuration dialog if or when and API endpoint is added for it. 

#### Device level Configuration
Device level config
#### Effect level Configuration
Effect level config
#### Home screen showing all settings buttons
Home screen showing all settings buttons

## Contributing requirements


rbergen:
@KeiranHines Small initial review point based on the screenshots: I think it would be good if the HTML (links) in the setting descriptions is/are rendered. If malicious HTML is a concern: the text/HTML in question is part of the NDS codebase, so will always be reviewed before being added or changed.

Also, a small heads-up: partly based on #411, it's likely that the data type for the effect interval setting will change from an integer/milliseconds at some point in the future, to a float/seconds with a minimum value (likely to be 1). If you'd like to align on the timing of this (or maybe even incorporate it into this PR) then I'm happy to do so.

KeiranHines:
Thanks. Currently I am just using the description provided from the /spec API as the helper text. I can look at providing some form of parsing to allow the links to be rendered. 

As for the interval. That's no worries at all. The dialog is rendered based on the specType provided by the /spec API. So long as that is updated to float or anything the UI will render it. 

Currently the only type with validation is the bigInteger type. If we were to make it a float and add a min max via the /spec API it would be a small change to use that min/max.

rbergen:
Ok, then I'll isolate the interval retyping in a separate PR, to keep things manageable. There are still some nuts to crack about certain details anyway, so it's probably better to make that a separate change.

Note that the min/max values are a C++ `double` in the actual `SettingSpec` class, even if they are only applied to integer settings right now. That is to say that at the back-end side, things are already lined up to also apply min/max values to floats.

KeiranHines:
Yeah I figured this was a good enough first pass with the settings I had available to get community feedback and ideally encourage more adoption of using the settings API. Of course that will mean bug and improvement requests for the UI which I am happy to support.

site/src/components/home/config/configDialog.jsx
```diff
@@ -0,0 +1,238 @@
...
+
+        return <TextField
+            {...baseProps}
+            {...textFieldProps}
+            error={error}
+            onError={(e) => updateError(setting.name, e)}
+            helperText={helper}
+            onChange={(e) => {
+                const v = e.target.value;
+                if( v === '' || v< 0 || v> 2**32) {
```
rbergen, line 67: Note that "regular" Integers (i.e., not being PositiveBigIntegers) can be negative, and the min/max values the Integer data type can support are correspondingly different (-2,147,483,648 and 2,147,483,647 to be precise).

site/src/components/home/config/configDialog.jsx
```diff
@@ -0,0 +1,238 @@
...
+                    setHelper(setting.description);
+                }
+                if(!v || isNaN(v)) {
+                    setValue('');
+                } else {
+                    setValue(Math.floor(e.target.value));
+                }
+            }}
+        />;
+    case settingType.Float:
```
rbergen, line 83: The Float type represents a floating-point numeric type. If I'm reading this code right you're currently using a checkbox to edit it?

rbergen (commented):
I've gone over this in more detail now and I have one comment and one question. Also, I still think it would be really nice if the HTML in the setting descriptions would render before we merge this. :)

site/src/components/home/config/configDialog.jsx
KeiranHines: I forgot to come back over this and add in an entry, initially I just started with the switch statement in the order of the type specification. I have added in integer to be a standard text field now.

site/src/components/home/config/configDialog.jsx
KeiranHines: Same as the other comment. That was a mistake, its now a textfield.

site/src/components/home/config/configDialog.jsx
rbergen: Great, thanks! For the numeric types we may at some point want to implement client-side "bounds checking" based on the minimumValue and/or maximumValue that can be embedded in the setting specs.

KeiranHines: I agree. At this stage without any known settings to test with I thought it was best just to get something up for people to start using before implementing validation client side too much. 

Ideally long term is like validation on all numerics, and string length either specified in effects or as default values

rbergen:
> I recently explained many settings in the new Mesmerizer effects...and dozens more that are on a branch. Those *SM* effects have LOTS of settings. Currently that have all kind of weird rules as they were hoisted from a system that apparently had 0-99 controls of a limited number, so there are effects that do one thing from 1-32, anther from 32-65, and so on. They exist.

Yes, I saw that when reviewing them. The solution here may have to be sanitizing the effects in question if/when we want to make the settings modifiable at runtime, instead of promoting the weirdness to the front-end.

KeiranHines:
@robertlipe I'd be happy to drive the UI side. As for the effects with integer configuration. I must not have it enabled on my mesmerizer. I will try to find time tomorrow and reflesh it with it enabled to test that config. 

As for sliders if that is a feature wanted. I could add a slider for integer values with a range <= 100, or some other range. I feel sliders over a certain granularity get a bit annoying.

rbergen:
@KeiranHines There's a definite risk in me talking on @robertlipe's behalf, but because I did review the effects in question I'm going to take it: I think Robert is talking about _potential settings_  in the context of a conversation about actually published effect settings. That is to say: I don't think any of the "integer setting" effects publish SettingSpec instances, nor do they publish effect setting values or contain logic to retrieve them.

KeiranHines:
@rbergen I had a look at intergrating https://www.npmjs.com/package/react-jsx-parser into the UI to enable html tag passing. While it works, I am running into an issue with pulling that pacakge from a CDN. I will have limited access to work on this project for the next ~3 weeks so as such I would complete that request as part of a future improvement in another PR.

rbergen:
@KeiranHines That's great news, I really appreciate it! I'll run a test of the updated UI when I have a chance (hopefully later today), and I'll merge this if that's successful.

rbergen:
Ok, I've tested this on my Mesmerizer, and I found the following:

- The text of links in the setting descriptions are basically illegible in dark mode because of the "default link blue" on the dialog background. I think we need to make the link text color match the color of regular text.
- If the effect interval is changed in the device settings dialog, it is not updated in the top bar on the home screen. 
- I can confirm that the countdown timer very easily gets out of sync with what's actually happening on the device.

@KeiranHines How involved would it be to fix each of the first two points?

KeiranHines:
The first one shouldn't be too bad. It will just need some CSS. I'll just have to find the best way to add that to a dynamic Dom element.

Second one may be a little more involved as those two components in the UI are disconnected however I should be able to trigger and update between them.

rbergen:
@KeiranHines Ok, then I'd like to fix the first one before we merge. The other two can be corrected by hitting the refresh button anyway, until actual fixes are in place.

KeiranHines:
@rbergen first one is fixed. 
there is now an a.link and a.visited variable in the theme file. I have selected colors from the existing colors that I found reasonable on light and dark themes. Please feel free to adjust the hex-codes if other colors are preferred. 

I have had some thoughts with the second and third issue. I will look at adding a context/provider to the front end and see if I like that for managing the state of the interval and timeout.

KeiranHines:
@rbergen I have started updating the timer management [here](https://github.com/KeiranHines/NightDriverStrip/tree/feature/globalEffectState). Its still early and will need some good testing on my part. 
That branch includes a context that updates the `/effects` response into a global state. From there, I have added an interval sync every 30s to keep the front and backends semi-sync'd. I have also updated the buttons on the home page to be refreshed based on an updated API call. 

This should fix bugs 2 once I add a `sync` call to the device config save. Bug 3 should already be fixed to some degree as the update every 30s will make sure they get back in sync. That timer can be easily adjusted to a smaller time if we want.

rbergen:
@KeiranHines 

> first one is fixed.

Tested and confirmed! Which I'm happy to merge this after CI completes.

Concerning the other branch you're now working on: nice! I look forward to that becoming a PR.

I will be opening one myself soon (probably today) to tidy up the default and effect-specific intervals. I'll describe the details in the PR, but I'm giving you a heads-up that I'm adding a new property to the /effects response that we may well want to use in the web UI.

KeiranHines:
Thanks for the heads up. I will be travelling for the next two weeks. I'll keep an eye out for the pr and update my branch while waiting at an airport or when I get back to normal life.

PR #448, opened 2023-09-27
Include (closer to) only what you actually use

robertlipe:
## Description
Everything includes globals.h. (It's not great, but I get it...)
Globals.h includes most everything.

This just isn't great for modularity and dependency tracking.

Tired of watching audio.cpp and network.cpp get recompiled
when I touched "remotecontrol.h" (wat?), I tore into globals.h. IMO,
I didn't trim aggressively, but if you read each line removed and ask
"does every file in the source code have a WHATEVER?", the review should
go down pretty smoothly.  Hopefully, the remote example is self-evident.

Of course, some #includes did get moved to the source files where they're
actually needed. You need a spiffs object or a WiFi password or a JPEG
decoder? Let me hook you up with some of those. But audio doesn't need
improv and improv doesn't need soundanalyzer and so on.

The code could still use a good bathing in "include-what-you-use", but
this gets rid of most of the lower-hanging and silly fruit.

Tested: Successful buildbuddy. 

Now I can go fix work on remote and not keep rebuilding ntpclient.

## Contributing requirements


rbergen:
@robertlipe With the merge conflict resolved and the subsequent build errors fixed, I think this is good to go! 

Having suffered circular include hell a few times myself on this project, you have my thanks for this clean up.

PR #453, opened 2023-10-06
Remove code setting common GPIOS to inputs when ATOMLIGHT ledstripgfx.h

robertlipe:
If a developer put LED_PINn on pins 4, 12, 13, 14, 15, this would have ensured that no data would have ever left the chip. PIN5 would have been OK only because this code "accidientally"? missed it.

I'm highly suspicious of this code and doubt that any of us can test it and prove it's correct, we can only inspect it and reason that it's wrong.

Do you agree that it's wrong?

## Contributing requirements

### DIFF
--- a/include/ledstripgfx.h
+++ b/include/ledstripgfx.h
@@ -124,13 +124,6 @@ public:
 
         AddLEDsToFastLED(devices);
 
-        #if ATOMLIGHT                                                   // BUGBUG Why input?  Shouldn't they be output?
-            pinMode(4, INPUT);
-            pinMode(12, INPUT);
-            pinMode(13, INPUT);
-            pinMode(14, INPUT);
-            pinMode(15, INPUT);
-        #endif
     }
 
     // PostProcessFrame


rbergen:
The BUGBUG comment in the code (which was added by @davepl) effectively says it's weird, but the code is still there.

With the relatively limited experience I have with "hardware" projects, I learned one lesson the hard way: don't change code target at a specific hardware setup that you think is bogus unless you can test the result of your change. I can't test this change, so I'm not going to adopt it (yet).

Dave, can you confirm your atom light still works with these ATOMLIGHT pinMode(n, INPUT) lines removed?

rbergen:
> Apologies if you took it that way. I was trying to show my work on how I arrived at that conclusion. It just sounded like an argument. I do that a lot.

If it came across different than you intended it, then it may well be that the problem is at my end in the sense that I just interpreted it the wrong way. You obviously don't have to apologize for that. :)

With that said, back to waiting for an Atom Light owner who's willing and able to test!

prschguy:
Had played with those lines early on before the code was changed to have devs. Back then, making changes there didn't do anything positive. Wasnt sure what they were for, so I forgot about them. That may be what is keeping Atom light from using other displays. Will revisit those lines on hopefully a few boards this weekend and see what happens. Hopefully something good comes from it.

davepl:
Sorry the context is missing from this email thread, but wrt the Atom light, I just checked in some fixes a couple of weeks ago to bring it back up to working on the current codebase.  Reprogrammed my own lamp.

I don’t know much for sure in life, but I’m pretty confident that code (a) should read OUTPUT, and (b) isn’t needed.  It’s not needed in the other configs, and FastLED does it for you anyway.

It’s one of those things I’d almost change without testing.  Almost!  If I get a chance today I’ll try it.

- Dave

prschguy1:
I can confirm that removing 
 //#if ATOMLIGHT                                                   // BUGBUG Why input?  Shouldn't they be output?
           // pinMode(4, INPUT);
            //pinMode(12, INPUT);
           // pinMode(13, INPUT);
            //pinMode(14, INPUT);
            //pinMode(15, INPUT);
        //#endif
Causes no ill effects to atom light. In addition to that, allows at least atom light stack to have a big beautiful display that runs completely stabile . Have run it for hours with no crashes. Have added some sound reactive effects to that build that are also working well. I am kicking myself now for always knowing that code was there, but never did anything about it. Honestly, just forgot it was there. Thanks for all that who helped in this. In working on this, I pulled up the code from a year ago. My, how much the code has changed since then!

Thanks again, Craig

prschguy1:
To allow anyone following along, this is what I did to make it work:
[env:atomlight]
extends         = dev_m5stack
build_flags     = -DATOMLIGHT=1
                  ;-DSHOW_VU_METER=0  ; if set to one, displays vu meter on 1 spoke ???
                  -DM5STACKCORE2=1
                  -DEFFECT_SET_VERSION=0   ;should effect persist, or restart list after re-start                
                  -DLOAD_GFXFF
                  ${psram_flags.build_flags}
                  ${remote_flags.build_flags}
                  ${unity_double_flags.build_flags}
                  ${dev_m5stack.build_flags}
board_build.partitions = config/partitions_custom_noota.csv



 #elif ATOMLIGHT
   #ifndef PROJECT_NAME
    #define PROJECT_NAME            "Atom Light"
    #endif


    #define ENABLE_WEBSERVER            1   // Turn on the internal webserver
    #define ENABLE_WIFI                 1   // Connect to WiFi
    #define INCOMING_WIFI_ENABLED       0   // Accepting incoming color data and commands
    #define WAIT_FOR_WIFI               0   // Hold in setup until we have WiFi - for strips without effects
    #define TIME_BEFORE_LOCAL           5   // How many seconds before the lamp times out and shows local content
    #define COLORDATA_SERVER_ENABLED    0   // Also provides a response packet
    #define NUM_CHANNELS    4
    #define MATRIX_WIDTH    53     // My maximum run, and about all you can do at 30fps
    #define MATRIX_HEIGHT   1
    #define NUM_LEDS        (MATRIX_WIDTH * MATRIX_HEIGHT)
    #define ENABLE_REMOTE   0                     // IR Remote Control
    #define ENABLE_AUDIO    1 
     
    #define LED_PIN0 33
    #define LED_PIN1 32
    #define LED_PIN2 27
    #define LED_PIN3 19
    #if USE_PSRAM
        #define MAX_BUFFERS         500
    #else
        #define MAX_BUFFERS         10
    #endif
    
    //#define IR_REMOTE_PIN 
    #define NUM_INFO_PAGES 2
    #define NUM_FANS        1
    #define FAN_SIZE        NUM_LEDS
    #define LED_FAN_OFFSET_BU       0
    #define FAN_LEN                 (NUM_FANS * FAN_SIZE)
    #define POWER_LIMIT_MW (INT_MAX)              // Unlimited power for long strips, up to you to limit here or supply enough!

    #define DEFAULT_EFFECT_INTERVAL     (1000*20)
    #define NUM_RINGS   1    
    #define RING_SIZE_0   NUM_LEDS
Please feel free to critique as necessary, may be making some mistakes along the way. Any help along the way is surely appreciated.

rbergen:
@prschguy1 and @robertlipe It seems to me we have confirmation that Atom Light still works with the lines removed that are the subject of this PR's change. That's good enough for me to merge this.

@davepl This also means you no longer have to find the time to confirm Atom Light working with the lines in question removed.

PR #454, opened 2023-10-06
Make brightness a device setting

rbergen:
## Description

The meat of this PR makes the g_Values.Brightness global brightness value a JSON-persisted device setting by moving it to DeviceConfig. Beyond that it includes a few tiny refactors/clean-ups that are very closely related to the main change.

## Contributing requirements


include/deviceconfig.h
```diff
@@ -338,6 +356,34 @@ class DeviceConfig : public IJSONSerializable
...
+        if (newNumericBrightness < BRIGHTNESS_MIN)
+            return { false, String("brightness is below minimum value of ") + BRIGHTNESS_MIN };
+
+        if (newNumericBrightness > BRIGHTNESS_MAX)
+            return { false, String("brightness is above maximum value of ") + BRIGHTNESS_MAX };
+
+        return { true, "" };
+    }
+
+    void SetBrightness(int newBrightness)
```
rbergen, line 377: While the brightness is actually a unit8_t, this function accepts an int. The main benefit of that is that we are friendlier towards callers, and we sort out the matter of keeping the actual brightness within defined boundaries. An additional benefit is that we don't have to add another specialization for CWebServer::PushPostParamIfPresent (and if we're pedantic, a new data type to SettingSpec).

include/globals.h
```diff
@@ -1075,9 +1075,6 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 
 #define STACK_SIZE (ESP_TASK_MAIN_STACK) // Stack size for each new thread
 #define TIME_CHECK_INTERVAL_MS (1000 * 60 * 5)   // How often in ms we resync the clock from NTP
-#define MIN_BRIGHTNESS  10
-#define MAX_BRIGHTNESS  255
-#define BRIGHTNESS_STEP 20          // Amount to step brightness on each remote control repeat
```
rbergen, line 1080: These were moved closer to where they are actually used. The first to (now) in DeviceConfig.h, the latter in remotecontrol.cpp.

include/ledstripgfx.h
```diff
@@ -89,8 +89,6 @@ class LEDStripGFX : public GFXBase
         #ifdef POWER_LIMIT_MW
             set_max_power_in_milliwatts(POWER_LIMIT_MW);                // Set brightness limit
         #endif
-
-        g_Values.Brightness = 255;
```
rbergen, line 93: Removed because LEDStripGFX has no business deciding what the global brightness value should be.

src/deviceconfig.cpp
```diff
@@ -60,18 +60,7 @@ DeviceConfig::DeviceConfig()
...
-        location = cszLocation;
-        locationIsZip = bLocationIsZip;
-        countryCode = cszCountryCode;
-        openWeatherApiKey = cszOpenWeatherAPIKey;
-        use24HourClock = false;
-        useCelsius = false;
-        ntpServer = NTP_SERVER_DEFAULT;
-        rememberCurrentEffect = true;
-        powerLimit = POWER_LIMIT_DEFAULT;
-
```
rbergen, line 73: All this was largely a duplication of defaults set in the DeviceConfig declaration in deviceconfig.h. Whatever was missing there has now been added there, so this can go.

src/remotecontrol.cpp
```diff
@@ -77,12 +79,13 @@ void RemoteControl::handle()
...
         effectManager.StartEffect();
-        g_Values.Brightness = 255;
+        g_ptrSystem->DeviceConfig().SetBrightness(BRIGHTNESS_MAX);
         return;
     }
     else if (IR_OFF == result)
     {
-        g_Values.Brightness = std::max(MIN_BRIGHTNESS, (int) g_Values.Brightness - BRIGHTNESS_STEP);
+        auto& deviceConfig = g_ptrSystem->DeviceConfig();
+        deviceConfig.SetBrightness((int)deviceConfig.GetBrightness() - BRIGHTNESS_STEP);
```
rbergen, line 88: This invocation uses the fact that DeviceConfig::SetBrightness accepts a plain int, and keeps the actual brightness within limits.

KeiranHines:
@rbergen is there a reason why min brightness is defaulted to 10 not 0. I compile with min 0 so I have the option to turn the leds off if need be.

rbergen:
@KeiranHines I just adopted the value that was in globals.h. That value was set by @davepl, so you'd have to ask him for the authoritative answer. If I were to venture a guess it would be that people might think something is wrong with their matrix or strip if it's completely dimmed to black - certainly now that the controlling board could boot with that setting.

revision29:
This is going to make reworking the remote code much easier.

revision29:
Is the plan to go through the whole codebase and change all references to g_Values.Brightness to query deviceConfig instead? I know it's referenced in ledmatrixgfx.cpp and in the code that draws stats to a built in OLED display, if enabled. I can't remember where the latter happens.

src/remotecontrol.cpp
revision29: I would suggest setting the reference variable to device config earlier in the code. I'd suggest before the if logic, such as line 41 and move "decode_results results;" down a line.

I would also suggest not setting to full power when the power button is pressed. The drawing system should just reference the last value set.

rbergen: Ok, duly noted. :)

> I would suggest setting the reference variable to device config earlier in the code. I'd suggest before the if logic, such as line 41 and move "decode_results results;" down a line.

Both of these concern personal preferences in coding (style). You're totally entitled to yours, I code things this way. In the case of deviceConfig, there is a specific reason I declare a variable (because I otherwise have to pull it from `*g_ptrSystem` twice), and I like my variables to have the smallest scope possible - in this case the `if (IR_OFF == result)` code block. the `decode_results` thing I didn't touch, i.e.: I left that where it is.

> I would also suggest not setting to full power when the power button is pressed. The drawing system should just reference the last value set.

That's how the remote control logic in the current codebase works. I'm not modifying that; I'm merely making the brightness a JSON-persisted device setting.

rbergen:
@KeiranHines I forgot to mention in the PR description that it may now be opportune to add logic in the front-end that does boundary checking on settings that have minimum and/or maximum values set in their setting specifications. So far the settings that have them check the boundaries when a "set" is attempted and act accordingly - as they should - but it might offer better UX if the UI shows them that they're out of bounds before hitting Apply.

include/deviceconfig.h
```diff
@@ -338,6 +356,34 @@ class DeviceConfig : public IJSONSerializable
...
+    {
+        if (newBrightness < BRIGHTNESS_MIN)
+            SetAndSave(brightness, BRIGHTNESS_MIN);
+        else if (newBrightness > BRIGHTNESS_MAX)
```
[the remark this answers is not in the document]
rbergen: Nice, better. Will do! Thanks!

include/deviceconfig.h
```diff
@@ -376,12 +377,7 @@ class DeviceConfig : public IJSONSerializable
 
     void SetBrightness(int newBrightness)
     {
-        if (newBrightness < BRIGHTNESS_MIN)
-            SetAndSave(brightness, BRIGHTNESS_MIN);
-        else if (newBrightness > BRIGHTNESS_MAX)
-            SetAndSave(brightness, BRIGHTNESS_MAX);
-        else
-            SetAndSave(brightness, uint8_t(newBrightness));
+        SetAndSave(brightness, std::clamp(uint8_t(newBrightness), BRIGHTNESS_MIN, BRIGHTNESS_MAX));
```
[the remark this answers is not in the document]
rbergen: You're right in the sense that I'm casting the wrong way around. I need to cast the MIN and MAX values up to int for the whole thing to work (they're set to uint8_t specifically in part to remind people fiddling with them that the _type_ boundaries are 0 and 255). Casting _is_ necessary because the parameters to clamp all have to be of the same type.

rbergen:
Confirmed functioning on Mesmerizer, both concerning persistence and configurability via the Web UI.

PR #455, opened 2023-10-07
Fix SetGlobalColor method in effectmanager.h to actually fill color

revision29:
## Description
There is an error in the efeftmanager.h file. The original version would not fill a solid color with the method SetGlobalColor. It would instead set a custom fire effect. It should do as advertised and set a global color. This minor change does just that.

## Contributing requirements


rbergen:
@robertlipe It may be that @revision29 is not yet aware that after you open a PR, any changes you push to the branch you opened the PR _from_ automatically get promoted to the PR as well. The PR doesn't take a snapshot of the branch, it tracks it (a reason for that is that this allows review comments etc. to be processed and applied before a PR is merged).

Joe, that being the case: if you want to continue development on a branch after opening a PR that otherwise reflects what you _intend_ to change, you'll have to do so in a new/other branch.

rbergen:
And yes, @revision29, in its current state this PR is unmergeable. Besides what @robertlipe mentioned, it incorporates the changes in #454, which has yet to be merged, and it commits files in the directory `site/dist`, which should/would be covered by .gitignore since that directory was introduced.

About the _intended_ change: looking at the if/else block you're aiming to modify, I think @davepl chose the flame effect to apply the selected color to a strip effect that responds to audio, and he may have an opinion about changing that. I'll check with him.

revision29:
@robertlipe and @rbergen  I was not aware that all changes made to my branch get promoted to the pull request. That is not my intention at all. I am just trying to change a couple lines of code in a single file. I'll cancel the pul request, create a branch for this change and try again. And yes, I was manually incorporating the changes from 454 but also trying to fix syncing problems with GitHub and clean up my code. Thanks for your patience as I learn the ropes with the process.

PR #459, opened 2023-10-08
Updated ledstripgfx.cpp to apply system brightness to pixels.

revision29:
## Description
Recently a device brightness level was added. This code takes advantage of that by applying it to led strip pixels just before they are displayed by FastLED. 

After the pixels are added to FastLED, we iterate each one and apply the system level brightness. This only applies to regular strip effects, and not to matrices. If you notice it is actually calling the fadeLightBy method of the CRGB color object. So we are actually passing an inverse of the brightness level.

The awesomeness of this is that a person can set the global brightness and not cause eye pain to themselves or loved ones, especially when working with lights at night.

## Contributing requirements

### DIFF
--- a/src/ledstripgfx.cpp
+++ b/src/ledstripgfx.cpp
@@ -53,11 +53,10 @@ void LEDStripGFX::PostProcessFrame(uint16_t wifiPixelsDrawn, uint16_t localPixel
 
     auto& effectManager = g_ptrSystem->EffectManager();
 
-    for (int i = 0; i < NUM_CHANNELS; i++) {
+    for (int i = 0; i < NUM_CHANNELS; i++) 
+    {
         FastLED[i].setLeds(effectManager.g(i)->leds, pixelsDrawn);
-        for (int j = 0; j < FastLED[i].size(); j++){
-            FastLED[i][j].fadeLightBy(255-g_ptrSystem->DeviceConfig().GetBrightness());
-        }
+        fadeLightBy(FastLED[i].leds(), FastLED[i].size(),255-g_ptrSystem->DeviceConfig().GetBrightness());
     }
     FastLED.show(g_Values.Fader); //Shows the pixels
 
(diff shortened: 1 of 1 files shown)


revision29:
That is a much better solution! I will try that and see if it works. I saw the reference to style standards in the making contributions documentation.  Having never learned any coding styles, I just looked and kept moving. Putting curly braces on their own line makes my eye twitch, but I’ll go ahead and do it anyway.  :) Thanks for showing me the way. On Oct 8, 2023, at 5:59 PM, Robert Lipe ***@***.***> wrote:﻿
I'm out of order right now, but can you save a loop? Most of the fastled

methods have overloads that accept a count



fadeLightBy( leds, 34, 128);  fades the first 34 LEDs by half



It may still have to loop internally, but it can potentially parallelize,

unroll in a way individual calls can't and it amortizes the overhead of the

function (specifically) from O(nleds) to O(1).



Loops inside loops can sneak up on you badly for performance.





Also, the style guide that the project requires is the Microsoft one, for

some reason 😉. Put the curlies in lines of their own so you don't get

confused by seeing more code at the same time without scrolling.

revision29:
@robertlipe the inner for loop has been removed. Now the single function call adjusts the output on all of the pixels. I also put the curly braces back where they go. The end result is that strip effects are drawn at the device's brightness level.

rbergen:
> Operators like - usually have three operands surrounded by spaces, but this line is so long it's not offensive. I just mention it to help shape future changes.

I second that. With the spaces missing things look cramped after the second comma in the call to fadeLightBy(), but like @robertlipe I'm not going to withhold approval over it either.

revision29:
@rbergen @robertlipe spaces added. I appreciate the feedback. Minor things like spaces do impact how readable the code is.

rbergen (approved):
LGTM!

PR #460, opened 2023-10-09
Bump smallest effect JSON buffer size to 192 bytes

### LINKED ISSUE #457: Problem with storing / recalling effects?
## Bug report
I have noticed that when I erase the flash on a device and re-upload a fresh compile which loads the rainbow effect it stretches a rainbow over the whole LED strip (depending on the speed one sets the effect to). The rainbow shifts across the length of the strip and wraps around and pulses as it moves. In other words, most pixels are different colors and you can see the rainbow shift linearly.  If I power cycle the device, the effect is not the same. It draws all pixels in the strip as the same color. As it fades to the next color in the spectrum, all pixels are the same color.

This is not the desired behavior.

I don't have the knowledge / familiarity with the codebase to figure this out. It seems there is either a problem in how the effect is stored as json blobs or after loaded from json on boot it is not being triggered properly. I suspect this is happening to more than just the rainbow effect. The rainbow effect is more readily noticeable. Since the effect is drawn as expected when it's loaded directly from code, the problem seems to be in the json processing code. 

**Problem**

The rainbow effect breaks after a reboot.

**Steps**

1.  Install a fresh compile of the demo environment on a freshly erased device.
2.  Watch the beautiful rainbow glide across the strip.
3.  Reboot the device.
4. Watch in horror as the rainbow no longer shifts. All pixels show the same color as they all shift from hue to hue. 
5. Spend hours testing different boards, checking wiring, double checking code.

rbergen:
## Description

This increases the smallest JSON buffer size used for effect serialization from 128 to 192 bytes. It seems that the introduction of the "mi" property in #454 makes that some effects that serialize additional properties (like RainbowFillEffect) now run out of buffer space, meaning properties get lost while serializing. 

As such, this PR fixes #457. 

Thanks to @revision29 for reporting this.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternSMNoise.h
+++ b/include/effects/matrix/PatternSMNoise.h
@@ -383,7 +383,7 @@ class PatternSMNoise : public LEDStripEffect
 
     virtual bool SerializeToJSON(JsonObject& jsonObject) override
     {
-        StaticJsonDocument<128> jsonDoc;
+        StaticJsonDocument<192> jsonDoc;
 
         JsonObject root = jsonDoc.to<JsonObject>();
         LEDStripEffect::SerializeToJSON(root);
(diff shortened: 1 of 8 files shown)


rbergen:
Reviewed this myself a few times after testing on my own devices. Even though they don't actually use the majority of the effects that were changed, I concluded it's a change that's so small it's safe to merge.

PR #466, opened 2023-10-21
Check for JSON buffer overflows

rbergen:
## Description

While looking into the documentation concerning the upgrade from ArduinoJson version 6 to 7 in follow up to #463, I found that in ArduinoJson version 6.17, a function `JsonDocument::overflowed()` was introduced that indicates if the JSON buffer that was reserved for the document ran out of space during serialization. 

This PR checks for this condition, and:

- aborts through `assert()` in cases an incomplete JSON serialization impacts the core functionality of the project, or
- logs an error in cases where it doesn't. This is the case if it concerns objects that are returned by the web API.

Although this doesn't take the guesswork out of the buffer size to allocate, it does flag when that size turns out to be too small.

Finally, I've replaced JSON buffer size literals with a `_jsonSize` constexpr where appropriate.

## Contributing requirements

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -142,6 +142,8 @@ class DeviceConfig : public IJSONSerializable
         if (includeSensitive)
             jsonDoc[OpenWeatherApiKeyTag] = openWeatherApiKey;
 
+        assert(!jsonDoc.overflowed());
+
         return jsonObject.set(jsonDoc.as<JsonObjectConst>());
     }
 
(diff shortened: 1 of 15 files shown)


rbergen:
Confirmed working/non-destructive on Mesmerizer and Spectrum. I will leave this PR open for another day in case someone wants to comment on it, and merge after.

rbergen:
@robertlipe Thanks, I always do appreciate the second pair of eyes. Now that four eyes have looked at it, I'll proceed to merge.

mggates39:
I updated the Stock Ticket effect on my repo with the new Json overflow
assert.  So far so good.

PR #474, opened 2023-10-27
More small fixes

rbergen:
## Description

This:

- Applies JSON buffer overflow checking to more places, in follow up to #466 - and actually complete that implementation across all effects.
- Makes the link between `EffectManager`'s declaration and certain effects cleaner, thereby preventing circular dependency problems that have otherwise known to occur (as illustrated in the discussion of #456). It also makes the inclusion of "soundanalyzer.h" more explicit in source files that need it.

## Contributing requirements

### DIFF
--- a/include/effectmanager.h
+++ b/include/effectmanager.h
@@ -43,8 +43,6 @@
 #include <math.h>
 
 #include "effectfactories.h"
-#include "effects/strip/misceffects.h"
-#include "effects/strip/fireeffect.h"
 
 #define JSON_FORMAT_VERSION         1
 #define CURRENT_EFFECT_CONFIG_FILE  "/current.cfg"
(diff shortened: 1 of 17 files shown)


rbergen:
Thanks! I always tell PlatformIO to build release builds and according to VSCode `assert(!jsonDoc.overflowed())` expands to:

```cpp
(__builtin_expect(!!(!jsonDoc.overflowed()), 1) ? (void)0 : __assert_func ((__builtin_strrchr( "/" "<long path>\\NightDriverStrip\\include\\effects\\strip\\paletteeffect.h", '/') + 1), 109, __PRETTY_FUNCTION__, "!jsonDoc.overflowed()"))
```

Which looks good to me. Maybe this changes if you explicitly define NODEBUG, but that's something I never do.

PR #477, opened 2023-10-29
Feature/global effect state

KeiranHines:
## Description
Added a global effect state context with all the data from the `/effects` endpoint. This enables an easy sync method to update the `/effects` state data in one location and enables timers to all be consistent. 

As part of this feature I have added support for `eternalInterval` (#451), showing an `infinity` symbol on the Interval, Time remaining and active effect countdown if the global timer is set to 0. 

## Contributing requirements


rbergen:
@KeiranHines Thanks! Considering the bigger magnitude of this change I'll spend some time testing this on one of my devices. It may therefore take a little longer before I submit my review.

KeiranHines:
No worries at all. Any and all feedback welcome.

rbergen:
The settings change sync seems to work well. Very nice!

I did notice that the /effects endpoint now seems to be polled twice at more-or-less 30 second intervals. This is a screen snippet of my Chrome Dev Tools running while the web UI is loaded to illustrate what I mean:
image

Is this the intended behaviour?

KeiranHines:
@rbergen  there is a few things here with the endpoint being called. Firstly I have noticed there is a bug where the timer is being created twice. I will try and work out the cause of that and eliminate it. I think its something in the build pipeline. Secondly there is a number of other times the effects endpoint is called, anywhere you see `sync()` will call `/effects` and restart the timer. This could be something as simple as the current effect expiring and a sync being done to update to the new effect and time. The timer changing etc. The 30s tick is meant more as a maximum time between updates. If your effect interval was say 10s it would be called every 10s.

KeiranHines:
@rbergen just to give an update. The issue with the effect firing twice was due to how react 18 treats strict mode, and webpack/babel injecting strict mode in places. As a solution I have tested migrating from Webpack/babel to vite. I will keep the process the same for developers. I am to spend some time tomorrow testing and validating the all features from the vite implementation work the same as Webpack.

rbergen:
@KeiranHines Thanks for the updates, both in text and code. Do the latter mean that this PR is ready for a second review, or are you still working on/testing things?

KeiranHines:
It's good for a second review. I noticed a one line bug when working on the updates to the list view and drag and drop support. I will push it to this branch tomorrow morning for me.

rbergen:
Then I'll wait with my review until my tomorrow morning. Sometimes time zones _do_ work out well. :)

rbergen:
@KeiranHines Thanks for the updates! I've retested, and it looks to me like the overly eager fetching of the effects list has been fixed. I did notice that after setting the interval to 0 on the main page and then editing the interval again, the edit box shows the "good old" 4294967295 as the interval. I'm guessing it pulls that from the /effects endpoint, which currently still returns the _effective_ effect interval, that being the maximum value of uint32_t.

With the infinity symbols now in place, I'll update the /effects endpoint to return the real effect interval (0) instead. I'll add that to #490 after merging this.

rbergen (approved):

PR #478, opened 2023-10-29
Config dialog improvements

### LINKED ISSUE #472: Web UI doesn't save empty string values for settings
## Bug report

**Problem**

The web UI doesn't save empty string values for string settings. More or less coincidentally this made sense for all available settings until now, but with the upcoming merge of #469 a string setting (hostname) will be introduced that can be empty.

**Steps**

1. Load the web UI on a device with firmware that includes the hostname device setting. 
2. Open the device settings dialog.
3. Enter and apply a hostname if one is not already set. Then reopen the device settings dialog.
4. Remove the hostname and apply again.
5. Reopen the device settings dialog, and observe that the hostname that was configured before step 3 is still set.

**Notes**

To support the web UI knowing what settings are optional (i.e. may be empty), an optional "optional" property has been added to the output of the /settings/specs endpoint. That will also be merged with #469.

### LINKED ISSUE #476: Web UI: implement minimumValue and maximumValue boundary checking
The web UI currently does not apply `minimumValue` and/or `maximumValue` properties that may occur in the specification JSON objects for certain (numeric) settings. 

Ideally, the web UI should provide the user visual feedback when a value they enter is out of bounds from the perspective of `minimumValue` or `maximumValue`. Values that are out of bounds can also be excluded from POSTs to the /settings endpoints, as they will be ignored by the back-end anyway.

KeiranHines:
## Description
Fixed #472 Added support for `emptyAllowed` configurations to be saved.
Fixed #476 Added support for configuration to limit between min and max values for numeric input types. If no min/max is set the dataTypes min/max is used as a proxy.  

## Contributing requirements

### DIFF
--- a/site/src/components/home/config/configDialog.jsx
+++ b/site/src/components/home/config/configDialog.jsx
@@ -56,7 +56,7 @@ const ConfigInput = ({setting, updateData, updateError}) => {
     }, [setting, updateError, error]);
     useEffect(() => {
         if(value !== setting.value) {
-            if(value === '') {
+            if(value === '' && !setting.optional) {
                 updateData(setting.name, undefined);
             } else {
                 updateData(setting.name, value);
@@ -247,7 +247,8 @@ const settingProps = PropTypes.shape({
     typeName: PropTypes.string.isRequired, 
     value: PropTypes.any,
     readOnly: PropTypes.bool,
-    writeOnly: PropTypes.bool
+    writeOnly: PropTypes.bool,
+    optional: PropTypes.bool
 });
 
 
--- a/site/src/components/home/config/configDialog.jsx
+++ b/site/src/components/home/config/configDialog.jsx
@@ -49,7 +49,8 @@ const ConfigInput = ({setting, updateData, updateError}) => {
         label: setting.friendlyName,
         id: setting.name,
         onChange: (e) => setValue(e.target.value),
-        value: value
+        value: value,
+        error: error
     };
     useEffect(() => {
         updateError(setting.name, error);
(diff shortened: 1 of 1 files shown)


site/src/components/home/config/configDialog.jsx
```diff
@@ -49,59 +49,61 @@ const ConfigInput = ({setting, updateData, updateError}) => {
...
             }
         }
     }, [value, updateData, setting]);
     const readOnly = !!setting.readOnly;
+    let min;
+    let max;
     switch(setting.type) {
     case settingType.Integer:
+        min = setting.minimumValue ? setting.minimumValue : (-2)**31;
+        max = setting.maximumValue ? setting.maximumValue : 2**31 -1;
```
rbergen, line 73: Won't `setting.minimumValue` or `setting.maximumValue` also evaluate to `false` if they are 0 (which can very well be the case)? 

Same question obviously also applies to the `PositiveBigInteger` and `Float` cases.

rbergen: @KeiranHines I had to look it up (my JS is a bit rusty), but 0 is indeed falsy. Which makes that I think we should apply an explicit strict inequality with `undefined` in these conditional expressions.

KeiranHines: You are correct. That was a mistake. I have corrected it now.

rbergen:
I have tested both the range checking and the empty string value submittal, and both seem to work well. 

@KeiranHines Combining the two, I was now thinking we should maybe also add an error condition when `emptyAllowed` is explicitly defined as `false` for a String setting, and its value is empty? Then we can choose to:

- Apply an empty String value (`emptyAllowed === true`)
- Reject it as an error (`emptyAllowed === false`), or 
- Silently ignore it, as we currently do for all but one (`emptyAllowed === undefined`). 

I'd have to make some changes in the backend to distinguish between "not set" and `false`, but I can do that independently of the implementation of this feature in the frontend. 

What are your thoughts?

rbergen:
@KeiranHines I was wondering if you could (just) comment on the idea I posted concerning empty strings? It's no problem at all if you can't (yet) work on the front-end implementation of it, but if you agree it's worth implementing at some point then I can already take care of the back-end part of it. I can do that because the current front-end implementation is compatible with both the current setup and the one I'm proposing.

KeiranHines:
I could see a use case for things like YouTube channel in the subscriber count effect. If you add the backend it's a single if statement to add it into the front-end and I'll add it when available.

rbergen:
@KeiranHines I've just merged #490, so you can add the check for `emptyAllowed === false` whenever you're ready.

PR #489, opened 2023-11-03
Supernova: dont merge pixels from VU meter at top of screen to bottom.

### LINKED ISSUE #446: Supernova effect draws VU in bottom row
I assume this effect moves stuff on the screen, and for some reason remnants of the VU meter are getting copied to the bottom line of the matrix.  It might be walking one line more than it needs to.

robertlipe:
drawPixelXYF() was merging pixels from the VU meter t othe bottom of the
line, so if there was sound in the room, red and green smears appared.

Fixed by making it better obey the HEIGHT values and compensating for
VU.

Follow through on my whining to make it look (more) like it was written
by a programmer.  Generated assembly code is now beautiful with
references carrying values across methods. Data is now per-Blob,
sequentially accessed, and in a std::array.

More importantly, memory is accessed sequentially and locally, just
like read-ahead and paging systems like.  Most accesses should be
cache hits now.

These various minor optimizations resultign in 12 FPS improvement
according to self-reporting. It _looks_ fast.

Fixes #446.

Tested with asserts() inserted for out-of-bounds pixel accesses (now
remove) and 30+ minutes of running on stock mesmerizer.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternSMSupernova.h
+++ b/include/effects/matrix/PatternSMSupernova.h
@@ -4,60 +4,80 @@
 
 // Inspired by https://editor.soulmatelights.com/gallery/1923-supernova
 
-class PatternSMSupernova : public LEDStripEffect 
-{
-    void Init()
-    {
-        memset(enlargedObjectTime, 0, sizeof(enlargedObjectTime));
-        memset(PosX, 0, sizeof(PosX));
-        memset(PosY, 0, sizeof(PosY));
-        memset(Hue, 0, sizeof(Hue));
-        memset(State, 0, sizeof(State));
-        memset(SpeedX, 0, sizeof(SpeedX));
-        memset(SpeedY, 0, sizeof(SpeedY));
-        memset(IsShift, 0, sizeof(IsShift));
+// A more cache-friendly version of the 8 independent arrays that were
+// used. This allows better locality per member, especially since we
+// access them sequentially.
+// Nothing special here; just default-zero initialized struct.
+class Object {
+  public:
+    void Objects() {
+        PosX = 0.0f;
+        PosY = 0.0f;
+        State = 0.0f;
+        SpeedX = 0.0f;
+        SpeedY = 0.0f;
+        Hue = 0;
+        IsShift = false;
     }
 
+    float PosX;
+    float PosY;
+    float SpeedX;
+    float SpeedY;
+    uint8_t Hue;
+    uint8_t State;
+    bool IsShift;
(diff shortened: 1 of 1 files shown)


include/effects/matrix/PatternSMSupernova.h
```diff
@@ -4,60 +4,80 @@
...
-        memset(Hue, 0, sizeof(Hue));
-        memset(State, 0, sizeof(State));
-        memset(SpeedX, 0, sizeof(SpeedX));
-        memset(SpeedY, 0, sizeof(SpeedY));
-        memset(IsShift, 0, sizeof(IsShift));
+// A more cache-friendly version of the 8 independent arrays that were
+// used. This allows better locality per member, especially since we
+// access them sequentially.
+// Nothing special here; just default-zero initialized struct.
+class Object {
```
rbergen, line 11: For the record, I'll comment on the brace style used not being "Microsoft-like". I will explicitly not make it an issue beyond that.

Also, I find "Object" a bit of a confusing name for a class, but that may just be me.

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -4,60 +4,80 @@
...
+  public:
+    void Objects() {
+        PosX = 0.0f;
+        PosY = 0.0f;
+        State = 0.0f;
+        SpeedX = 0.0f;
+        SpeedY = 0.0f;
+        Hue = 0;
+        IsShift = false;
     }
```
rbergen, line 21: What exactly is the purpose of this function? It looks like a constructor, except the function name Object**s** doesn't match the class name Object. As far as I can tell, it also isn't used anywhere.

rbergen (commented):
Thanks for picking this up! 

I have a couple of tiny points for the record and one genuine question.

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -4,86 +4,96 @@
...
     void Draw() override {
         step = -1;
         g()->DimAll(200);
-        for (unsigned i = 0; i < enlargedObjectNUM; i++) {
-            if (!IsShift[i] && step) {
-                starfield2Emit(i);
+
+        for (auto& debris_item : _debris_items) {
+            if (!debris_item._is_shift && step) {
+                StarfieldEmit(debris_item);
```
rbergen, line 39: (etc.) Shall I just ignore the bracing beyond what you changed in your most recent commit? :)

Issue #479, opened 2023-10-29
Web UI: Global Brightness and Color1/Color2 selection


davepl:
I believe we need a global brightness slider and a way to set the global color/palette.  So something like a slider and two color panels with color pickers attached?

I'd imagine it to be either at the top of the main page/effects list, perhaps even in a non-scrolling region below which the list scrolls.

You might imagine it differently, and that's fine too, as long as you imagine it better :-)

rbergen:
Hahaha, that last comment is rather cheeky, Mr. Plummer. :D

We already have a device setting for the global brightness that can be set via the web UI, but the slider is currently not yet in place. I'll think of a way to "suggest" to the UI to use a slider instead of a numeric input field.

I think the color picker control is something that still needs to be developed as well. Once that's in, we can easily apply that to the global color/palette once it has been "promoted" to a device setting itself.

Specifically picking the brightness and color/palette as settings that are lifted out of the device settings dialog (and then exactly where they are put on the main screen) is a matter of personal preference. I don't have a strong opinion about it, so I'll go with whatever those who do think is best.

rbergen:
So I played with some ideas and then decided to apply the KISS principle: I think it's best to just add a "Slider" data type (next to "String", "Integer", "PositiveBigInteger", "Boolean", etc.). It will translate to an `int32_t` at the C++ level, and it will be mandatory to provide both a minimum and a maximum value when the corresponding `SettingSpec` is created.

This does mean that I can't apply this change to the existing global brightness setting until the slider control has actually been implemented in the frontend.

KeiranHines:
@rbergen is color1/color2 exposed via the API similar to the global settings or effectSettings? If its exposed so I have something to work with, I am happy to build out a colorPicker or similar Config.

rbergen:
@KeiranHines Not yet. For one, I didn't want to expose colours or palettes until there was an editor to pick/compose them - but the JSON data type encodings for both can be found in /effectsConfig.

Concerning color1/color2 I'm not sure I fully understand those properties themselves. I do know we have a global color (which may well be color1) and a global palette, but a palette is usually a set of _16_ colours, not 2. 

@davepl Could you maybe help out by clarifying why we're talking about 2 colors in the context of a discussion about 16 color palettes?

KeiranHines:
Fair enough with respect to exposing the pallette. I will take a look at the backend and have a go at getting the UI working at some point soon.

[...]

PR #490, opened 2023-11-04
Extend SettingSpec

rbergen:
## Description

This extends the `SettingSpec` struct for future use by the web UI. Specifically, it:

- Makes the `EmptyString` property `std::optional`, thereby allowing for a distinction between explicitly not allowed, explicitly allowed, and "we'll silently ignore empty" - as per [this comment](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/478#issuecomment-1785353947) on #478.
- Adds a Slider type, which can be used to ask the front-end to use a slider to set a numerical value. This is discussed in [this comment](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/479#issuecomment-1790376007) on #479.
- Adds some rudimentary (and optional, because `struct`) logic to finish setting defaults and perform sanity checks on the consistency of properties. If at some point this reaches a higher level of complexity than it currently does then we may need to decide to make the `struct` a `class`; partly because errors in the specs will quickly show up as weird behaviour in the web UI that's easy to relate to oddness in the spec contents, I don't think that's necessary yet.

## Contributing requirements


KeiranHines:
@rbergen for the slider type I have a few thoughts. Firstly and probably the most encompassing. Instead of a full new type. Would it make sense to add a metadata property. Similar to the 'emptyString' property for numeric config. Something like a `preferredDisplay` property. On the frontend, I could ingest that and render out different inputs based on that, if its set e.g. 

```
{
   name: 'brightness',
   type: 'integer',
   preferredDisplay: 'slider' 
} 
```
That would clearly say, I want this field to be a slider display not just an numeric input box. Alternatively you could use a `tickSize` property. If the frontend sees that on a numeric field it renders a slider with ticks at the tickSize frequency. This could be useful in cases where you wanted to a numeric input but didn't need the fidelity of every single integer value. 

To add the questions that got me to this point.
1. With the slider input type, will it support integer and floating point at some point. 
2. In the idea for the UI to just render a slider between min/max and allow the user to adjust to any point along that slider, or should there be a config for tickSize or majorTick/minorTick? 
3. Would it be easier to add a `preferredDisplay` `uiDisplay` or similar property as optional for existing datatypes so the input can be configured as say slider, dropdown or other options in the future without having to have a unique type for each one?

rbergen:
I thought about this before making these changes, and I decided against adding a metadata property of the "preferredDisplay" type. The reason is that we would effectively add another dimension to the UI control specification and front-end implementation side, and I think that could make things quite a bit more complicated in the end, for very limited gain if any.

Coming from a different angle one could argue that the setting spec "type" property already _is_ a control selector. At the JSON level, almost all properties are actually a "number". That's even true for the type representing a colour, as you'll soon find. As such, it already is the type that indicates what editor control to apply.

About the underlying type at the C++ side, we could indeed make that a float (or actually, double) instead of an int. I chose an int because as a user I've hardly ever seen sliders that actually slide over a floating point value - where they possibly would, they tend to be sliding over a more discrete derived scale (like a percentage) instead. In fact, IIRC then all OS-provided slider controls I ever worked with back in my desktop app development days worked with integer ranges and values as well.

Concerning ticks and hints for those, I think we can choose to chew that food when we have to, if we ever have to. In the case of the brightness I think we can do without them, for one.

KeiranHines:
Sounds good to me. I will add support for the new slider type in the Ui sometime this week. I assume with the new slider type min/max are required otherwise a slider wouldn't be very functional.

rbergen:
@KeiranHines That's correct. I've even implemented a (rudimentary) check on the presence of both min and max values when a Slider `SettingSpec` is constructed at the back-end.

PR #494, opened 2023-11-06
Feature/config dialog

### LINKED ISSUE #493: Web UI: improve config Option Support
* Add emptyAllowed === false on strings [comment](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/478#issuecomment-1793716121)
* Add support for a color picker [comment](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/479#issuecomment-1793684011)
* Add support for a color palette (Array of colors) [comment](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/479#issuecomment-1793684011)

KeiranHines:
## Description
Config Dialog now support Sliders, Color, Color Palettes and emptyAllowed === false. Fixes #493.

See attached images and gifs for examples. Only color config that has been added is the subscriber count effect. 
Empty_Strings_allowed_false_example
Palette_config
Slider Example
Color Picker

### DIFF
--- a/site/src/components/home/config/configDialog.jsx
+++ b/site/src/components/home/config/configDialog.jsx
@@ -39,64 +39,70 @@ const RGBToInt = ({r, g, b}) => {
     return parseInt(r.toString(2).padStart(8, 0) + g.toString(2).padStart(8, 0) + b.toString(2).padStart(8, 0), 2);
 }
 
-const ColorPickerDialog = ({title, color, value, setValue, open, closeFn}) => {
+const ColorPickerDialog = ({title, initialColor, settingValue, setValue, open, closeFn}) => {
+    const [color, setColor] = useState(initialColor)
+            
     return <Dialog open={open} onClose={closeFn}>
         <DialogTitle>{title}</DialogTitle>
         <DialogContent sx={{maxWidth: '250px'}}>
-            <RgbColorPicker color={color} onChange={(color) => setValue(RGBToInt(color))}></RgbColorPicker>
+            <RgbColorPicker color={color} onChange={(color) => setColor(color)}></RgbColorPicker>
             <Box sx={{minHeight: '15px'}}></Box>
             <Box>
-                <TextField label="r" value={color.r} sx={{width: '33%'}} onChange={(e) =>{
-                    const input = e.target.value
-                    let newR = Math.min(255, Math.max(0, parseInt(input)))
-                    setValue( v => {
-                        let {r, g,b} = intToRGB(v);
-                        if (isNaN(newR)) {
-                            r = input === '' ? 0 : r;  
-                        } else {
-                            r = newR
-                        }
-                        return RGBToInt({r, g, b})
-                    })}
-                }
-                InputProps={{ inputProps: { min: 0, max: 255 } }}
+                <TextField label="r" value={color.r} sx={{width: '33%'}}
+                    onChange={(e) =>{
+                        const input = e.target.value
+                        let newR = Math.min(255, Math.max(0, parseInt(input)))
+                        setColor( c => {
+                            let {r, g,b} = c;
+                            if (isNaN(newR)) {
+                                r = input === '' ? 0 : r;  
+                            } else {
+                                r = newR
+                            }
+                            return {r, g, b}
+                        })
+                    }
+                    }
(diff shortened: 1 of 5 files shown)


KeiranHines:
@rbergen I look forward to more effects having color config now :P

site/src/components/home/config/configDialog.jsx
```diff
@@ -118,21 +147,73 @@ const ConfigInput = ({setting, updateData, updateError}) => {
             {...baseProps}
             {...textFieldProps}
             helperText={helper}
+            onChange={(e) => {
+                if (value === '' && (setting.emptyAllowed === undefined && !setting.emptyAllowed)) {
```
rbergen, line 151: Shouldn't this check for `setting.emptyAllowed !== undefined` instead of `setting.emptyAllowed === undefined`? The error should (only) be shown when `emptyAllowed` is explicitly set to `false`. If it's `undefined` then the empty value should be silently ignored, as it is now.

rbergen: Thinking about it again, I think this `(setting.emptyAllowed === undefined && !setting.emptyAllowed)` can be replaced by `setting.emptyAllowed === false`, for simplicity.

KeiranHines: You are correct. Updated.

rbergen:
@KeiranHines I tested this, and found the following:

- When I open the config dialog for the Subs effect then the color picker for the background color opens immediately, without me selecting it. That's not how it should work.
- I think it would be better if the color selection dialog also allows the R, G and B values to be entered in numeric fields. A user may want a specific combination of values to be set, and offering only the hue slider and saturation/lightness field makes one have to "hunt" the right values using those.

rbergen:
@KeiranHines Thanks, that fixes the issues I mentioned in my previous comment. I've done some further testing, and I think there's one issue remaining. Consider the following scenario:

1. Open the settings dialog for the Subs effect.
2. Open the color picker for the background color. Change the color and click OK.
3. Open the color picker for the background color again. Either change the color or leave it untouched. Click Cancel.

What happens now is that the background color reverts to _the original color when the settings dialog was first opened_, so the color before the first change (step 2). I don't think it should do that. I'd expect the Cancel on the color picker to cancel whatever change you make in that interaction with it, not any (unsaved) changes that were made before. If someone wants to do that, they can use the Reset button or Cancel the overall change at the settings dialog level.

rbergen:
In other news, I've just pushed a small change to make the global Brightness setting use a Slider.

KeiranHines:
That shouldn't be hard to fix. It would be caused by where and how I am managing the state of the colour picker. I will try to rectify that in the next few days.

rbergen:
Perfect! I have to say, watching the background on the Subs effect shift colour the second I hit Apply is pretty darn cool!

rbergen:
@KeiranHines I think we've taken one step forward and one step back now. When I open the Subs effect settings dialog, I now see three colour squares for each of the two colours:
image

Each of those seems to work well - including the cancel behaviour -, but there really should be one colour selector for each colour, as there was before your most recent changes. And I double checked that the `/effect` endpoint returns one colour for both `backgroundColor` and `borderColor`, not an array of three or something like that.

KeiranHines:
@rbergen woops, my bad. I had an if statement in that was treating color inputs like a palette with hard coded colors so I could test them given I didn't have one setup in the backend. I forgot to remove it before I committed. Its removed now.

rbergen:
Alright, I'll give it another try! Thanks for the quick response!

rbergen:
@KeiranHines Tested and confirmed working now! I'll run CI on this because that's how we roll, and merge after. And when that's done I'll open a PR to add the global colors that Dave effectively asked for in #479.

rbergen:
@KeiranHines My apologies, I (almost) jumped the gun there. It seems the empty strings error doesn't trigger correctly when `emptyAllowed` is `false`. That is to say, it doesn't trigger when the value is empty, but _when the first character is then entered into the field_. 

You can try this with the Location setting for instance. If you clear that out nothing happens, when you then enter an "a" the error shows, and when you then enter another character or remove the "a" again, the error disappears.

rbergen:
@KeiranHines Thanks, that fixed it! Running CI one more time, will merge after.

PR #499, opened 2023-11-11
Update jsonserializer.cpp to break infinite fail

### LINKED ISSUE #473: Programmatically changing the effect interval causes an infinite recursive json buffer increase
## Bug report

**Problem**

When I set the effect interval from a key on my remote control, the interval is set. After it writes the value to json storage it then starts a loop where the json buffer size continues to increase until the board is power cycled. This does not happen when other events trigger json to store

What I see in the console:

(W) (NotifyJSONWriterThread)(C1) >> Notifying JSON Writer Thread
(W) (NotifyJSONWriterThread)(C1) >> Notifying JSON Writer Thread
(W) (SerializeWithBufferSize)(C0) Out of memory serializing object - increasing buffer to 6144 bytes
(W) (SerializeWithBufferSize)(C0) Out of memory serializing object - increasing buffer to 8192 bytes
(W) (SerializeWithBufferSize)(C0) Out of memory serializing object - increasing buffer to 10240 bytes

**Steps**

1. Have the debugger console running in VS Code.
2. Somewhere in code trigger: effectManager.SetInterval(30000);   
3.  Watch the json buffer size increase by 2048 bytes (as defined in jsonbase.h).
4. Watch the effects being drawn on the leds lag more and more.
5. Power cycle the board and the effect interval now reflects the value triggered by the remote. The buffer increase has stopped.
6. Press the button that changes the interval. Watch the console fill up with SerializeWithBufferSize notices.

**Example**

effectManager.SetInterval(30000);   

**Notes**

Until today, I was running on an outdated codebase (last changes from main branch merged into the branch I'm working on are from October 9). I merged the changes to the main branch from the pull requests dating from October 17th onward. I don't think the error showed up until I did that.

I suspect there are other instances where this error is popping up, but I am having a hard time isolating when / where. I just pushed a fresh build after erasing a device and it is giving the out of memory notice as well. But that appears to be related to a couple of audio effects I added to effects.cpp. If I disable those and disable audio, the out of memory warning goes away.

I will verify on a different board later and see if the error is present there as well.

revision29:
Added code to catch instances where SerializeWithBufferSize gets into an *infinite fail loop. Adjusted SaveToJSONFile to handle SerializeWithBufferSize failing.

## Description

There are times when a serializer fails to allocate memory for pJsonDoc. The SerializeWithBufferSize function interprets this as the JSON document needing more buffer. That might be the case sometimes, but it does not handle cases when the device runs out of usable memory. This code change keeps the buffer size adjustment, but adds code to handle the case where the device can't allocate memory for the JSON document.

The code allows for the allocation in SerializeWithBufferSize to fail 5 times before it stops trying. Once it stops trying it exits the loop. When SaveToJSONFile continues, it will have to handle having either a null pJsonDoc variable or pJsonDoc with a capacity of 0. If pJsonDoc is null is the result of the fail code in SerializeWithBufferSize. If SerializeWithBufferSize failed, it will skip trying to save the file. That is because if it tries to save the file with a null variable or pJsonDoc->capacity() of 0, it will cause an uncaught exception and crash the whole program and the device will reboot. So, this just returns false, causing it to skip the file writing operation.

Note: it is my assumption that the device is running out of memory. There might be some bug somewhere else in the code causing the memory allocation failure.

Fixes #473. 

## Contributing requirements

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -439,42 +439,42 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
     // known to work with at least one other type of device as well.
 
     #ifndef PROJECT_NAME
-    #define PROJECT_NAME                    "Mesmerizer"
+    #define PROJECT_NAME                "Mesmerizer"
     #endif
 
-    #define SHOW_FPS_ON_MATRIX              0
-    #define ENABLE_AUDIOSERIAL              0   // Report peaks at 2400baud on serial port for PETRock consumption
-    #define ENABLE_WIFI                     1   // Connect to WiFi
-    #define INCOMING_WIFI_ENABLED           1   // Accepting incoming color data and commands
-    #define WAIT_FOR_WIFI                   0   // Hold in setup until we have WiFi - for strips without effects
-    #define TIME_BEFORE_LOCAL               2   // How many seconds before the lamp times out and shows local content
-    #define ENABLE_WEBSERVER                1  // Turn on the internal webserver
-    #define ENABLE_NTP                      1   // Set the clock from the web
-    #define ENABLE_OTA                      1   // Accept over the air flash updates
-    #define ENABLE_REMOTE                   1   // IR Remote Control
-    #define ENABLE_AUDIO                    1   // Listen for audio from the microphone and process it
-    #define SCALE_AUDIO_EXPONENTIAL         0
-    #define ENABLE_AUDIO_SMOOTHING          1
-    #define REQUIRE_EFFECTS_SERIALIZATION   1   // Require effects serialization to succeed
+    #define SHOW_FPS_ON_MATRIX          0
+    #define ENABLE_AUDIOSERIAL          0   // Report peaks at 2400baud on serial port for PETRock consumption
+    #define ENABLE_WIFI                 1   // Connect to WiFi
+    #define INCOMING_WIFI_ENABLED       1   // Accepting incoming color data and commands
+    #define WAIT_FOR_WIFI               0   // Hold in setup until we have WiFi - for strips without effects
+    #define TIME_BEFORE_LOCAL           2   // How many seconds before the lamp times out and shows local content
+    #define ENABLE_WEBSERVER            1  // Turn on the internal webserver
+    #define ENABLE_NTP                  1   // Set the clock from the web
+    #define ENABLE_OTA                  1   // Accept over the air flash updates
+    #define ENABLE_REMOTE               1   // IR Remote Control
+    #define ENABLE_AUDIO                1   // Listen for audio from the microphone and process it
+    #define SCALE_AUDIO_EXPONENTIAL     0
+    #define ENABLE_AUDIO_SMOOTHING      1
+    #define EFFECT_PERSISTENCE_CRITICAL 1   // Require effects serialization to succeed
 
-    #define DEFAULT_EFFECT_INTERVAL         (MILLIS_PER_SECOND * 60 * 2)
-    #define MILLIS_PER_FRAME                0
+    #define DEFAULT_EFFECT_INTERVAL     (MILLIS_PER_SECOND * 60 * 2)
+    #define MILLIS_PER_FRAME            0
 
-    #define NUM_CHANNELS                    1
-    #define RING_SIZE_0                     24
-    #define BONUS_PIXELS                    0
-    #define MATRIX_WIDTH                    64
-    #define MATRIX_HEIGHT                   32
-    #define NUM_FANS                        128
-    #define FAN_SIZE                        16
-    #define NUM_BANDS                       16
-    #define NUM_LEDS                        (MATRIX_WIDTH*MATRIX_HEIGHT)
-    #define IR_REMOTE_PIN                   39
-    #define INPUT_PIN                       36
-    #define LED_FAN_OFFSET_BU               6
-    #define POWER_LIMIT_MW                  (5 * 8 * 1000)         // Expects at least a 5V, 8A supply
+    #define NUM_CHANNELS                1
+    #define RING_SIZE_0                 24
+    #define BONUS_PIXELS                0
+    #define MATRIX_WIDTH                64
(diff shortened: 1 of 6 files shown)


rbergen:
@revision29 First off, this check is a great catch, and obviously something I should have added from the very beginning. However, I do think we need to give the way we handle the out-of-memory scenario some (more) thought.

The fact of the matter is that we serialize stuff to the FS for a reason, and in certain cases we should probably consider the inability to do so a fatal condition - making a crash and reboot actually appropriate. I think that always applies to the device settings, and to the effect list in certain build configurations.  
I therefore think we should do the following:

- Add a global `REQUIRE_EFFECT_SERIALIZATION  define that defaults to 0 but can be set to 1 for certain build configs (`SPECTRUM` and `MESMERIZER` come to mind).
- Report back the success or failure of serialization to the caller of `SerializeWithBufferSize` (a bool return value seems logical), and propagate that back to the caller of `SaveToJSONFile`.
- Let the caller (usually a lambda of some sort) decide how to handle the failure: log and continue or crash and reboot.
- Update the callers to crash when saving device settings, or when saving the effects list with `REQUIRE_EFFECT_SERIALIZATION` defined as 1.

At a more granular level:

- I don't think there's a point to retrying allocation 5 times in an out of memory condition. If the RAM isn't there, that's it. Again, it's up to the caller to figure out what to do next.
- I don't see the purpose of removing the log line that reports the increase in buffer size. Let's keep that in.
- An out of memory condition while serializing is definitely worthy of a warning-level debug line at least.

rbergen:
I've taken the liberty of updating your PR with the suggestion I made earlier today. Please let me know if this works for you.

revision29:
@rbergen Having the option to reboot after failure is a good idea. On my device rebooting won't help. But, another device might end up having a memory leak and a reboot would be helpful. Regarding the fail 5 times before failing: there was no need for that. Just trying to preserve the original idea of retrying a few times before failing. I also deleted the debug line by accident. My working version had debug lines everywhere, and I was over-aggressive in removing them when moving the changes to a new fork.

Your changes to the PR are great. The code works as expected on the device that is running out of ram. REQUIRE_EFFECTS_SERIALIZATION set to 1 does reboot the device, and 0 just gives outputs the debug statement.

While I have your attention: I am 90% done with getting global color into a device setting. There are a couple of more small issues to work out first, but I expect to submit a pull request in about 2 weeks. Just wanting to let you know my intention and progress.

revision29:
@robertlipe 
> I like the idea of making this case fatal, but let's just do it for all

It makes sense to apply it to all configurations, but I want to make sure we keep the flag. We can default it to 1, but I’d really like the ability to set it to 0 for my use case.

> Can you please look at the diffs as shown on github and see why it's identifying so many "unchanged" lines as changed?

I noticed it too. I copied the whole contents of each whole file into VSCode and it all looked normal. I thought the days of DOS vs Linux vs Mac line breaks were behind us.

> Also, what does the MS style guide say about whitespace on lambdas?

Is there a link to the style guide? I’ve seen it referenced but I don’t recall seeing a link in the project documentation. Forgive me if you have shared the link before.

Regarding the memory leak concern. It would be nice to build in a system wide reboot trigger when a certain threshold is met for low memory.  Determining what that threshold should be would be rather complicated. I’d also want the ability to disable the trigger. A lot of the cheaper boards have little memory to begin with.

rbergen:
So a lot of text was produced while I was sleeping, and first off I have to take ownership of a few of the things that were discussed - they were included in the commit I added to this PR:

- The addition of `REQUIRE_EFFECT_SERIALIZATION` was triggered by this situation, but I was actually reasoning from a different perspective - low memory is not the only reason JSON serialization _to the FS_ can fail. I was thinking if writing the effect list to SPIFFS should be considered a critical necessity or not, and I figured that in many cases it isn't. If you're only running one or a couple of effects on a low-spec device, then maybe one doesn't actually care if that "list" is written to the FS or not, regardless of the reason of failure. I renamed the thing to `EFFECT_PERSISTENCE_CRITICAL` to better express the meaning of the define.
- The whitespace diffs in globals.h are caused by me realigning the values after the various defines in those configuration blocks where I introduced `REQUIRE_EFFECT_SERIALIZATION`.
- The lambda format is also mine. I'm basically following the format for a function in those cases where the lambda code is more than a "one-liner" (i.e. more complex than a single statement, possibly followed by a return of a static value). In that, the function name is replaced by the square brackets. In one-liner lambdas I remove the argument parentheses if there are no arguments to pass to remove the clutter in the single line, but maybe I should keep them in multiline lambdas to make it more clear that we're looking at a lambda definition. 
  Anyway, I've scanned a few other places where multiline lambdas are defined, and there the parentheses that enclose the lambda are on the same line as the `[]()`, and the closing `}`, so I've adapted that style here (and applied it where it wasn't used exactly like that) now as well. So, two lines of vertical space saved. And yes, this way of formatting takes more vertical space than others, but personally I do find it the most clear. Also, I really, really don't want to adopt the "opening brace on end of line before the first line of the block" style here - we don't do it anywhere else either.   
  I know we can argue about this until the cows come home, but I'm not going to. :)

The low memory handling thing I already responded to in a comment on another PR.

rbergen:
@revision29 I just noticed this:

> While I have your attention: I am 90% done with getting global color into a device setting. There are a couple of more small issues to work out first, but I expect to submit a pull request in about 2 weeks. Just wanting to let you know my intention and progress.

This is a bit awkward, but I am about 30 minutes away from opening a PR to add _two_ global color settings myself... This is triggered by issue #479, and implements an approach that I've discussed with Dave in some out-of-band communication. I was waiting for #494 to be ready and merged before opening it, otherwise I would have already done so some time ago.

I apologize for the fact that our efforts are crossing here, and particularly for any time and energy you may have invested in something I have now conceptually duplicated. Considering the fact you have also been working on this, I _am_ interested in your feedback on my PR when I've opened it; I'll post another comment here with the PR number.

rbergen:
@revision29 PR #503 is what it's become.

PR #500, opened 2023-11-11
Restructure flag inheritance in platformio.

robertlipe:
I'm tired of watching huuuuuge lists of files build that I know won't be
used. I've not shaved everything to the bone, but I've cut the cases
that are big enough to annoy me out of the builds I care about. For
example, waiting for it to check out and build that ridiculous
olikraus/U8g2 library build up to 37 times on a full build is a drag.

Anything that used to set USE_SCREEN now uses ${use_screen.build_flags}
to get -DUSE_SCREEN and ${parent-device.lib_flags} to get the
dependencies and the -L ... -l stuff. I similarly tried to tighten up
the remote_flags case, upon which this was inspired.

There's still a bit of silliness, such as everything checking out and
building BUTTONS when almost none of our builds HAVE buttons...but that
lib builds quickly enough to not annoy me. Now checking it out and
storing it 38 times isn't awesome.

The way we have configuration split between globals.h and platform.ini
is definitely sub-awesome. I wish I could figure out how to make
Platformio's extra_script.py to look at the -D flags in CPPFLAGS and
then add_libs based on that. Then when globals.h consitionally adds
a -DFoo, we could add_libs Foo without it having to appear in globals.h
It's pretty silly to wait and checkout and build AsyncTCP or WebServer
if the configuration didn't even turn on ENABLE_WIFI, but we do

If your build is broken, it's probably this PR's fault. Buzz me with a
repro case for help if needed. Notably, this tightens many environments
to use the "correct" hardware lib_deps (which may include extra
libraries as it might have been extended) instead of those of base_libs.

Tested:
Successful python3 tools/build-all.py

## Contributing requirements


rbergen:
@robertlipe Thanks for submitting this. The segmented inheritance within platformio.ini is totally in line with the pattern I implemented in a massive overhaul of platformio.ini some time ago, so I obviously can't disagree with that. (That exercise aimed to significantly decrease repetition in the env blocks, not to decrease the inclusion of unnecessary dependencies.)

I've given the specific change some thought, and I'd like to propose a few small changes:

1. Rename the `USE_SCREEN` define to `HAS_SCREEN` or `SUPPORTS_SCREEN`; the flag now indicates if we have the bits on board to be able to use it, so I think it's good to reflect that in the name of the thing. Of course, the various `#if USE_SCREEN` etc. lines throughout the codebase then have to be updated accordingly.
2. Add a similar `HAS_REMOTE` or `SUPPORTS_REMOTE` define for the remote control scenario. We'd then have to add that to the `#if ENABLE_REMOTE` lines, turning them into `#if HAS_REMOTE && ENABLE_REMOTE`. Without this, it would now become possible to enable the remote control using its define when the stuff that's needed to make it actually work isn't there.

Let me know your thoughts on this.

rbergen:
> Can I push back on 2 and agree to keep thinking about it until we have a solid plan forward that lets us disentangle the platform.ini and the globals.h representations?

I'm sorry but no, not if you do want to apply selective inclusion of the remote control library dependency in this PR. The reason is this: we now _remove_ the ability to use the remote by not including the necessary library dependency for certain configurations, and that _will_ cause weird problems when people then set (or leave) `ENABLE_REMOTE` to 1 in those builds. Unless they have their head around the entire dependency/define/... stack and know the library may just not be there because of this new thing in platformio.ini, they won't have a clue why the thing isn't working when `ENABLE_REMOTE` says it should.

My statement is basically this: if we _selectively_ include library dependencies for certain capabilities in platformio.ini, we **have** to tell the code that's being built what we have, and thereby what we don't. That's what the HAS/SUPPORTS (I understand you prefer the latter, which is fine with me) defines are for. That's also where they (should) differ from the USE variety - those should indicate what is _chosen_ to be used. I can definitely imagine that other defines (USE or otherwise) should also be converted to HAS/SUPPORTS, but that can indeed be a follow-up.

In a sense, selectively including (or excluding) lib_deps adds another layer of complexity to the multiple levels that are already embedded in the platformio.ini/globals.h combo. I could have raised that as an objection to this change as a whole, but I didn't as I do agree this is a sensible step to take. We just can't abandon the rest of the setup until we fix it, regardless of how icky we think it now looks.

So, as far as I see it, we can do one of three things: 

1. Add the `SUPPORTS_REMOTE` flag and check for it in places that matter, on top of whatever else is already there, or
2. Clean up the `ENABLE_REMOTE` and other IR remote defines throughout the project now - possibly replacing `ENABLE_REMOTE` with `SUPPORTS_REMOTE` and killing the former with fire altogether -, or 
3. Postpone the selective inclusion of the relevant library dependency in platformio.ini and leave it in the global list for now. We can then pull the selective inclusion back in when we get around to 2.

revision29:
I'll add my two cents. Thanks for cleaning this up. I can guarantee this change will break my setups, but not sure how. I needed to restart with a fresh fork anyway.

I do have reservations about changing ENABLE_REMOTE to SUPPORTS_REMOTE. It is an input feature I want to enable, just like the web server. It is not natively supported by most boards since it requires adding an additional piece of hardware (IR diode). Logically, supports_remote is not true for most boards. ENABLE_REMOTE makes more sense to me. It is a small point in the midst of the greater issue being discussed above. But, it should not be enabled by default.

I can appreciate the need for these modifications. Wrapping my brain around the interplay between globals.h and platformio.ini took a while. There is a lot of redundancy and inconsistency in what features are enabled where. Sometimes web server is enabled in plat*.ini and sometimes in globals.h.

rbergen:
Bluntly put: no.

If we want to clean up the compile/build of the overall compile/build of the project, then that clean-up needs to be clean. That means at least:

- "Selected dependencies" are built when they are needed, and not built when they are not. Stubs (which I guess we would be creating ourselves?) to me are proof we just haven't figured out how to get that right.
- The code itself should be at least as or more straightforward about what it supports than it is now. We cannot rely on run-time console messages to tell a build-time user of the code why something isn't working if that decision is made by the code in one way or another; that should be obvious by looking at the code itself. Hence the previous exchange about the name of defines.

In my view, the aim should not be to save time on building dependencies. Instead, the aim should be to improve the structure of the project's build-time configuration _including dependencies_ so that it is obvious to a reasonable extent that a) certain dependencies don't need to be built and therefore aren't, and b) they're not there when they could be.

PR #501, opened 2023-11-12
Reconfigure Spectrum builds

rbergen:
## Description

This changes the feature flags of the Spectrum build configuration to allow the web server to be started and working - this currently doesn't work due to low memory conditions on some devices at least.

This is mainly for the purposes of the Web Installer; I consider it more likely that Web Installer users want to use the on-chip webserver than the services/features that have now been disabled. People who make and flash local builds can easily customize the build configuration in globals.h to their own liking, anyway.

## Contributing requirements

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -716,7 +716,7 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 
     #define ENABLE_WIFI                 1   // Connect to WiFi
     #define INCOMING_WIFI_ENABLED       1   // Accepting incoming color data and commands
-    
+
     #define WAIT_FOR_WIFI               1   // Hold in setup until we have WiFi - for strips without effects
     #define TIME_BEFORE_LOCAL           5   // How many seconds before the lamp times out and shows local content
     #define COLORDATA_SERVER_ENABLED    0   // Also provides a response packet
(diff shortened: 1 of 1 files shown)


rbergen:
> Looks like good tradeoffs. I'd have ranked remote above ntp, but it's likely that remote is way more expensive than ntp in memory taken.

You're right. The remote logic lives in its own task/thread, with its own stack etc., whereas the NTP updates are dealt with in the loop of the main network thread. Memory-wise, NTP is therefore almost "for free".

rbergen:
> On the excessive buffer copies in the Arduino stack, part of that is just Arduino's style of being flippant about such thing

I think we're now talking about the "webserver won't respond with larger files thing"? I didn't think that could be done "zero copy"; I'd imagine buffer space for a couple network packets to be needed to be able to slice, dice and push it through the driver stack. I just don't get why transfer would work for a smaller file in ROM, but not for a bigger one. But again, I'm only applying my version of common sense.

> Should we test if memory < some percentage or floor and include some scary words or maybe stick/flash one pixel to red so people know they're in a danger zone?

I'll think about the message. The tricky bit is getting the threshold right - maybe it shouldn't be a percentage, but an absolute number of KB? Otherwise we may end up crying wolf for no reason, which will make the whole thing meaningless.  
Promoting this sort of thing to the actual effect visualizations I'm not a fan of - nor is someone else, I know.

rbergen:
I agree that flagging a low memory condition makes sense, but I don't want to do that on the actual effects unless it is 100% certain we have a problem. So, unless and until we figure that out I just don't want to go there. 

Concerning accessibility of logging - where we can easily be a little less "careful", that may actually be better than you think. Whatever logging is output to serial can even be read using the Web Installer - ironically that's the one Web Installer feature that _always_ works, even when the whole Improv exchange has collapsed into a smoking heap of rubble.

I'll give it some more thought, as I mentioned. This PR I'll merge, thanks again for your feedback.

PR #511, opened 2023-11-14
Feature/designer improvements

### LINKED ISSUE #300: Web UI: Drag and Drop to reorder effects
In the effect UI, we need the ability to reorder the effects by dragging and dropping them to their new position.  As far as I'm aware Rutger has already added the underlying API for this, but check with him to make sure its checked into the code before relying on it being there!

### LINKED ISSUE #301: Web UI:  Change from TILES to LIST with checkboxes
Rather than small tiles for each effect, we should use a "list" format where each effect gets its own row tile. There should be a checkbox on the left side to enable/disable the effect.

Other functionality, such as Activate, can be a link within the tile, perhaps right-aligned.

image

KeiranHines:
## Description
Fixes #301: Adding a list view as well as the existing grid view for effects
Fixes #300: Drag and Drop to reorder effects is now possible on grid and list layout
Added a "show disabled" button to hide effects you have disabled. 
image

## Contributing requirements


rbergen (approved):
Tested, confirmed working and LGTM! 

@KeiranHines Thanks for submitting this, UI-wise this is a very nice number of steps forward!

PR #512, opened 2023-11-14
Fix Supernova effect

### LINKED ISSUE #510: Bug: Supernova effect dies
## Bug report

**Problem**

The Supernova effect dies after a few runs/minutes, leaving a permanently black panel. The reason is that the number of elements/debris items seems to gradually decrease to 0.

**Steps**
1. Start the Supernova effect and let it run for a while, restarting it if necessary. 

**Notes**

Without being able to exactly pinpoint it, my guess would be that this behavior got introduced with the merge of #489.

rbergen:
## Description

This fixes #510. The main change is that the effect makes a clean start every run, as it used to before #489 was merged. I've also:

- Cleaned up a few (more) things that didn't seem to serve any purpose.
- Reinstated the use of `g()->isValidPixel()` to check if we're using sane x and y values, and skipping the draw otherwise.
- Added a check that only avoids sampling VU meter pixels if we're actually showing it.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternSMSupernova.h
+++ b/include/effects/matrix/PatternSMSupernova.h
@@ -24,9 +24,7 @@ public:
 
     void Start() override
     {
-        for (auto& debris_item : _debris_items)
-            debris_item.Clear();
-
+        std::for_each(_debris_items.begin(), _debris_items.end(), [](DebrisItem& debris_item) { debris_item.Clear(); });
         g()->Clear();
     }
 
--- a/include/effects/matrix/PatternSMSupernova.h
+++ b/include/effects/matrix/PatternSMSupernova.h
@@ -121,11 +121,12 @@ private:
             hue2 += 1;
         debris_item._position_x = MATRIX_WIDTH * 0.5;
         debris_item._position_y = MATRIX_HEIGHT * 0.5;
+
         debris_item._speed_x = (((float)random8() - 127.) / 512.);
         debris_item._speed_y = sqrtf(0.0626f - debris_item._speed_x * debris_item._speed_y);
-        if (random8(2U)) {
+        if (random8(2U))
             debris_item._speed_y = -debris_item._speed_y;
-        }
+        
         debris_item._state = random8(1, 250);
         debris_item._hue = hue2;
         debris_item._is_shift = true;
--- a/include/effects/matrix/PatternSMSupernova.h
+++ b/include/effects/matrix/PatternSMSupernova.h
@@ -24,8 +24,8 @@ public:
 
     void Start() override
     {
-        std::transform(_debris_items.begin(), _debris_items.end(), _debris_items.begin(),
-            [](DebrisItem &item) { item.Clear(); return item; });
+        for (auto& debris_item : _debris_items)
+            debris_item.Clear();
 
         g()->Clear();
     }
--- a/include/effects/matrix/PatternSMSupernova.h
+++ b/include/effects/matrix/PatternSMSupernova.h
@@ -9,7 +9,7 @@ class PatternSMSupernova : public LEDStripEffect
 {
 public:
 
-    PatternSMSupernova() : LEDStripEffect(EFFECT_MATRIX_SMSUPERNOVA, "Supernova"), hue(0), hue2(0), step(0), enlargedDebrisItemNUM(0)
+    PatternSMSupernova() : LEDStripEffect(EFFECT_MATRIX_SMSUPERNOVA, "Supernova"), hue(0), hue2(0), step(0)
     {
     }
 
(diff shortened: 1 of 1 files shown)


include/effects/matrix/PatternSMSupernova.h
```diff
@@ -112,6 +108,8 @@ class PatternSMSupernova : public LEDStripEffect
         if (debris_item._state == 0 || x < 0 || x >= MATRIX_WIDTH ||
```
davepl, line 108: Good opportunity to reflow this line!

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -137,17 +135,23 @@ class PatternSMSupernova : public LEDStripEffect
...
-            // but we just clamp instead.
-            const int xn = std::clamp((int)x + (i & 1), 0, MATRIX_WIDTH - 1);
-            const int yn = std::clamp((int)y + ((i >> 1) & 1), 0, MATRIX_HEIGHT - 1);
-
-            CRGB clr = g()->leds[XY(xn, MATRIX_HEIGHT - yn)];
+            const int xn = x + (i & 1);
+            const int yn = y + ((i >> 1) & 1);
+
+            // Make sure we're on the panel and leave the VU meter pixels alone, if we're showing it
+            if (!g()->isValidPixel(xn, yn) || (showingVUMeter && yn == (MATRIX_HEIGHT - 1)))
```
davepl, line 151: Why the extra level of labels?  Why not (SHOW_VU_METER) directly in the expression, since it will always be defined?

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -24,10 +24,8 @@ class PatternSMSupernova : public LEDStripEffect
 
     void Start() override
     {
+        std::for_each(_debris_items.begin(), _debris_items.end(), [](DebrisItem& debris_item) { debris_item.Clear(); });
         g()->Clear();
-        enlargedDebrisItemNUM = 200;
-        if (enlargedDebrisItemNUM > MAX_COUNT)
-            enlargedDebrisItemNUM = 200U;
```
davepl, line 30: Is this the same 200 as the debris count define below?  I don't think they're liinked?

include/effects/matrix/PatternSMSupernova.h
rbergen: I actually didn't touch this in past because I'm not sure how to restructure it in a way to make it better. It's a whole bunch of terms no matter how you slice and dice it.  
I'm happy to take on a concrete proposal though! :)

include/effects/matrix/PatternSMSupernova.h
rbergen: It actually isn't always defined. It's only defined for matrix/HUB75 configurations, and even then not for all - env:mesmerizer_lolin doesn't, as I found when CI ran on an earlier commit. However, @robertlipe made a good alternative suggestion, which I will implement.

include/effects/matrix/PatternSMSupernova.h
rbergen: I didn't change the debris count magic to something I came up with myself. I reverted to the value it had before #489. At that time there was an index "if" that ran up to the value of `enlargedDebrisItemNUM`, thus 200.

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -24,10 +24,8 @@ class PatternSMSupernova : public LEDStripEffect
 
     void Start() override
     {
+        std::for_each(_debris_items.begin(), _debris_items.end(), [](DebrisItem& debris_item) { debris_item.Clear(); });
```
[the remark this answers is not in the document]
rbergen: No, the thing is that the original (pre-#489) logic started with a zeroed out set of arrays (now array of `DebrisItem` objects) every time the effect was started. I figured that was an important factor in the effect working correctly, as it "eats" its own output of `Draw()`, the next time `Draw()` is called. The `DebrisItem` constructor would only be called once per `DebrisItem` when the effect itself was created, now the `DebrisItem` objects are also cleared on every invocation of `Start()`.

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -137,17 +135,23 @@ class PatternSMSupernova : public LEDStripEffect
 
     void drawPixelXYF(float x, float y, CRGB color)
     {
+        #if SHOW_VU_METER
```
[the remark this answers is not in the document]
rbergen: I was looking for that thing, but couldn't find it. After looking again, now I have `¯\_(ツ)_/¯`. It's `EffectManager::IsVUVisible()`. Which I will promptly employ instead.

include/effects/matrix/PatternSMSupernova.h
rbergen: Had another look and came up with an idea after all. Thanks for the tickle! :)

PR #516, opened 2023-11-16
Fix crash on "large" configurations, especially on ESP32-S3.

### LINKED ISSUE #515: Our use of FastLED reliably crashes on "large" ESP32S3 configurations
This was originally reported in https://github.com/PlummersSoftwareLLC/NightDriverStrip/discussions/195#discussioncomment-7557796 , but as that's a discussion item that was already marked closed, I'm opening a bugreport so that I can have one of my lovely wall of text murder-mystery conversations with myself before providing a one-line fix ... that took about ten hours to research and test.

## Bug report

**Problem**
The original report came in on a board without PSRAM. The same weekend, we were battling two other issues with low memory causing bizarre symptoms. As this worked well with CHANNELS=1, worked for a long time with CHANNELS=2, and would reliably crash in a minute with CHANNELS=3, all signs pointed to memory exhaustion.

**Steps**
It turns out that it's not related to EITHER being low on memory or even that specific board. Its a bug thats been in FastLED forever that just happens to be more likely to exhibit on ESP32-S3. A generalized repro case follows:
1.  The problem was reported on heltecv3demo. Turns out, it can be reproduced on YD-ESP32-S3 with any of yd-esp32-s3-demo,   lilygo-tdisplay-s3-demo, and lilygo-tdisplay-s3-demo. This exonerates low memory issues as this is a 16MB board with working PSRAM.
2.  The original problem report uses pin numbers that are a bit nonsensical. I thought we might be configuring inputs with bad values and getting interrupts on non-configured sources as some of our exceptions indicated unknown exceptions. This was a dead-end. Reported error:
Guru Meditation Error: Core 1 panic'ed (InstrFetchProhibited). Exception was unhandled.
0 0x7ad00baa:0x3fc97230 in ?? ??:0
1 0x403781e5:0x3fc97280 in panicHandler at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/port/panic_handler.c:213
2 0x40377b74:0x3fc972a0 in xt_highint4 at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/port/soc/esp32s3/highint_hdl.S:119
3 0x00040022:0x3fc97360 in ?? ??:0

Additional errors:
Guru Meditation Error: Core 1 panic'ed (StoreProhibited). Exception was unhandled.

0 0x40382c58:0x3fc97100 in mwdt_ll_write_protect_disable at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/hal/esp32s3/include/hal/mwdt_ll.h:259
(inlined by) wdt_hal_write_protect_disable at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/hal/wdt_hal_iram.c:132
1 0x4202c335:0x3fc97120 in esp_panic_handler_reconfigure_wdts at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/panic.c:190
2 0x4202c724:0x3fc97150 in panic_handler at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/port/panic_handler.c:155
3 0x403781f9:0x3fc971a0 in xt_unhandled_exception at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/port/panic_handler.c:219
4 0x403792b8:0x3fc971c0 in _xt_user_exc at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/freertos/port/xtensa/xtensa_vectors.S:703
5 0x4202c6bf:0x3fc97280 in start_cpu0_default at startup.c:?
6 0x40377b74:0x3fc972a0 in xt_highint4 at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/port/soc/esp32s3/highint_hdl.S:119

3.  Build a config in include/globals with three channels for GPIOs known to be free.  To further let memory usage off the hook, it's reproducible with small number of LEDs. The key is the number of channels.
+ #define MATRIX_WIDTH            14
+    #define NUM_CHANNELS            3
+#define LED_PIN0 12
+#define LED_PIN1 13
+#define LED_PIN2 14

**Example**
With a single channel, it'll run for at least tens of minutes.
With two channels, it fails within a few minutes.
With three channels, it doesn't fail immediately, but it will usually fail within 30 seconds.

**Notes**

Being able to use the debugger was super helpful, but the failure wasn't terribly deterministic; it often failed in different places and the platformio debugger lies to you just often enough to not be a super trustworthy partner.  Still, some patterns formed. MOST of the crashes were chased into the dark alley of FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp. 

ESP32RMTController::doneOnChannel is called one per channel (LED strip in our case) when the DMA tx chain is complete. Thus, more channels == more interrupts == more time in this function. It contains both some #ifdef specific ESP32 device voodoo because the RMT DMACs are different between models and some semaphore signaling (that's abstracted into FreeRTOS thread handling in our case)  in our case. Seems a likely problem area. The end of that function is approximately:
    if (gNumDone == gNumControllers) {
        // -- If this is the last controller, signal that we are all done
    } else {
        // -- Otherwise, if there are still controllers waiting, then
        //    start the next one on this channel
        if (gNext < gNumControllers) {
            startNext(channel);

Quite often, our crash was inside startNext (deep inside...) but most importantly, that's the case where we're handling multiples of completion interrupts from the RMT devices from a single interrupt event.

Notably, this function is a performance hotpath and thus lives inside instruction RAM:
void **IRAM_ATTR** ESP32RMTController::doneOnChannel(rmt_channel_t channel, void * arg)

Unexpectedly, disabling the thread that wrote the .5Hz messages fixed the crash. (!)

The next to craziest thing is that I was able to find the root cause of this online.

The craziest thing is that this is a behaviour of the super well-known FastLED project, that there's a known fix for it available in the tree, that it's been there for four years, and that it's not turned on by [default](https://www.reddit.com/r/FastLED/comments/f1cmjo/comment/fhbaqxt/?utm_source=reddit&utm_medium=web2x&context=3). Crashing devices seems to be OK with them. It turns out this actually isn't ESP32-S3 specific; it's just more likely to show up there.

From https://github.com/FastLED/FastLED/issues/1220#issuecomment-829537474 : 

> there's a peculiarity of the ESP32 that access to the flash memory locks the regular memory bus. As a result, ESP-IDF enforces a fairly draconian policy: when flash access is going on, all code on both cores is disabled except interrupt handlers residing in IRAM. Here's where it gets interesting. All of our code is in IRAM (which was a huge pain, btw), BUT the RMT library code is not. So, what you're seeing is a bus violation: our interrupt handler is calling an RMT function that is not in IRAM while a flash access operation is going on.
> 
> OK, that's probably way more detail than you wanted! ;-) There are two ways to address this problem. One is to add your own synchronization to make sure that SPIFFs access is not going on while FastLED.show() is running. That's a pain. The other option is to add the following #define directive before you include FastLED.h:
> 
> #define FASTLED_ESP32_FLASH_LOCK 1

Sure enough. Every crash I studied was in or near code that was in IRAM. The interaction with the code writing JSON to "disk" triggered the SPI traffic on one core while we were likely on the other. MOST of the crashes I studied had a bus violation of instruction fetches or data stores or loads. So the shoe fit.

In my own testing, admittedly without LEDs, working up to 8 channels and up to 200 LEDs/channel (I didn't test further...) now ran without issue.

I worked with the poster to coach them on this change.  He also confirmed success.

After that, I attached a couple of strips to ESP32-S3 and plain ESP32 modules, including the one in Mesmerizer by dangling a strip from it. I observed no regression. Sure, in some world somewhere, adding a lock slows things down, but if it makes the code RIGHT, it's hard to argue with success.

This wall of text exists to serve as the permanent record of my upcoming change to turn on that flag.

See related discussions at, e.g.:
https://www.reddit.com/r/FastLED/comments/n7pi6a/fastledshow_not_compatible_with_ota_updates_on/
https://www.reddit.com/r/FastLED/comments/gea52t/would_fastled_block_a_webserver_running_on_esp32/
https://github.com/FastLED/FastLED/issues/967
And the actual fix:
https://github.com/FastLED/FastLED/commit/fec0f0300e14c7dd1fa4fb7eb08ff1fd862f0dae

robertlipe:
- Fix crash on "large" configurations, especially on ESP32-S3.

## Description
See  https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/515 for the (painful) details.

Fixes #515.

## Contributing requirements

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -102,6 +102,8 @@
 
 #include <Arduino.h>
 
+//  See https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/515
+#define FASTLED_ESP32_FLASH_LOCK 1
 #define FASTLED_INTERNAL 1               // Suppresses build banners
 #include <FastLED.h>
 


rbergen (approved):
@robertlipe Very nice catch! Thanks for doing the investigative work on this one; I could imagine a "going down the rabbit hole" feeling at some points.

It's nice that the flag delays (or, should) the flash access, not the drawing.

rbergen:
You're welcome!

The reason for not enabling it by default may just be that there is a performance/lead time hit, at least in theory, that not everybody wants to take. And as far as I can tell, I've never run into the issue the flag fixes on any of my devices since I got involved with this project.

If people start reporting issues here that may be related to this change we'll look into it. If stuff breaks in the future, we'll investigate and deal with it then. For all kinds of reasons the stability of the foundations we use is just not such that I'd like to be in an airplane that has its flight control systems built on top of them. One could say that with projects like this one, it's both a source of frustration _and_ part of its charm (insert comment about humanity's perpetual state of discomfort, here).

PR #518, opened 2023-11-16
Feature/device control

### LINKED ISSUE #482: Web UI: allow user to reset the device config, reset the effect list and/or reboot the device
The on-board API's `/reset` endpoint allows three types of "reset" to be triggered:

- A reboot of the device (`board=true`)
- A reset-to-defaults of the device configuration (`deviceConfig=true`), which should always be combined with a device reboot
- A reset-to-default of the effect list (`effectsConfig=true`), which should always be combined with a device reboot

So far, the UI does not offer a facility to trigger the latter two. A board reset can be triggered by opening the global settings dialog and clicking "Apply and reboot device" without changing any settings, but I think that's not very intuitive.

I think the web UI should allow a user to trigger each of the three resets, as an independent action. Because of the impact of the latter two, a confirmation flow should probably be triggered before actually applying the resets in question. When it comes to location, the global settings dialog does seem to be a logical place to me. Exactly how to embed it into that dialog's design is something I'm not exactly sure of, which also means I don't have a strong opinion about it, up-front.

KeiranHines:
## Description
Fixes #482 added a sidebar button for reboot, reset device and reset effect options. 
image

## Contributing requirements

### DIFF
--- a/site/src/components/home/home.jsx
+++ b/site/src/components/home/home.jsx
@@ -108,9 +108,9 @@ const AppPannel = (props) => {
                 </ListItemIcon>
                 <ListItemText primary="Settings"></ListItemText>
             </ListItem>
-            <ListItem id={'deviceControl'}>
+            <ListItem onClick={(e) => setDeviceControlOpen(a => a ? null : e.currentTarget)}>
                 <ListItemIcon>
-                    <IconButton onClick={(e) => setDeviceControlOpen(a => a ? null : e.currentTarget)}>
+                    <IconButton>
                         <Icon>settings_power</Icon>
                     </IconButton>
                 </ListItemIcon>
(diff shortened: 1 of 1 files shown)


rbergen:
@KeiranHines Very nice! This is going to save me more time than you could ever imagine :)

One tiny visual thing: it seems the popup menu now floats behind the sidebar, which makes the "reset device configuration" look a little cramped at the left. I think it would be better to either move the popup menu block a little to the right so its left border aligns with the right border of the sidebar, or have it above the sidebar in the visual stack.

KeiranHines:
Corrected 
image

rbergen (approved):
Tested, confirmed fixed. LGTM!

PR #519, opened 2023-11-16
Fix Powder/PatternSMBlurringColors effect

rbergen:
## Description

This fixes the Powder/`PatternSMBlurringColors` effect, which could get stuck with one color-fluctuating pixel in the lower-left corner of the matrix (technically the one pixel would "radiate" a little outwards, but who's looking that closely?)  
As with `PatternSMSupernova` before, the main issue seems to be that the effect needs to always start with an empty set of, in this case, `trackingObject...` arrays to ensure it has a usable working set of pixels to evolve. 

I note for the record that this effect would definitely benefit from an overall clean-up and OOP overhaul like the one that `PatternSMSupernova` received in #489. For the benefit of getting the effect to work the way it should as quickly as possible, in this PR I decided to focus purely on setting up the start-up conditions right.

## Contributing requirements


rbergen:
@robertlipe Thanks for your quick review!

PR #450, opened 2023-09-28
Visual fixes in SMSmoke

robertlipe:
@davepl, how about this?

## Contributing requirements

* [x] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [x] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [x] I selected `main` as the target branch.
* [x] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #525, opened 2023-11-18
Remove tear/pepper in fire from SMSmoke

### LINKED ISSUE #486: Smoke Broken!
Looks like a recent change has broken the noise effect of Smoke.  It's now all sparkly on the leading edges, and looks broken compared to what it used to, where it was a smooth smoke edge.

rbergen:
## Description

This updates the SMSmoke effect to remove the tear/"pepper in fire" that was added in #450. As such, fixes #486.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternSMSmoke.h
+++ b/include/effects/matrix/PatternSMSmoke.h
@@ -64,18 +64,12 @@ public:
     // Two diagonal (note Y is used for height AND X offset)
     // "wipers", of different colors. Each leaves a
     // sky-written trailer of color.
-    // slew is used to make the lines look less mathematically diagonal
-    // and introduce some delay between color1 and color2. This gives a
-    // "tear" in the refresh and as a bonus, gives some 'pepper in fire'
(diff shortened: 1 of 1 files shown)


rbergen:
Well, one of us must be missing something. I left a comment on #486 on November 7th, in which I indicated that I'd had a chat with Dave about his report, and that he'd like the tearing removed. I tagged you in it, explicitly asking you, and I quote "is this something you can look into?"

That was thus a week and a half ago. As you'd already indicated you weren't too happy with the tearing part yourself in an earlier comment (I quote once more: "I'll admit I'm not totally in love with that wiper effect at all anyway"), So, in my mind you had already indicated you'd be okay with the revert, and I figured you just didn't care enough about the issue to pick it up - which is fine, nobody's getting paid.

I've taken note of your labeling of my way of handling this. I don't think it applies given what I just wrote, so I'm not going to apologize for it. I'm also not going to spend more energy on it, because frankly I have better things to do.

PR #532, opened 2023-11-20
Interval input fixes

KeiranHines:
## Description
* Added press enter to submit new time interval.
* Changed interval display in designer from ms to a time format.
  * If interval Config < 1000ms it will show in ms and can be configured as 0.xxx seconds. anything 1s and above can be configured to second precision.  
intervalConf

A couple of other code cleanups. Fixed promises in chipRequests and fixed bad propTypes in effects.jsx

## Contributing requirements

### DIFF
--- a/site/src/components/home/designer/countdown/countdown.jsx
+++ b/site/src/components/home/designer/countdown/countdown.jsx
@@ -3,6 +3,7 @@ import { Box, Icon, Typography } from "@mui/material";
 import countdownStyle from "./style";
 import { EffectsContext } from "../../../../context/effectsContext";
 import PropTypes from 'prop-types';
+import { msToTimeDisp } from "../../../../util/time";
 
 const Countdown = props => {
     const { remainingInterval, pinnedEffect, sync} = useContext(EffectsContext);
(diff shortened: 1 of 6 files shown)


KeiranHines:
Feedback welcome if people would prefer a different display design.

rbergen:
@KeiranHines I will review the visual design when I test this. As already indicated, I want to make it impossible to configure an effect interval between 0 and 1 seconds (0 itself is allowed to configure "infinity"), so it's not necessary to cater for that scenario. 

Is it okay with you I'll add the updated to EffectManager::SetInterval() to ignore values between 1 and 999 to this PR? I will then test my change in combination with yours.

KeiranHines:
That's perfectly fine. I wasn't sure the best way to handle devices already set less than 1s. I guess I can just round it up so next set will be to >1s. 

It's also worth noting the setting dialog render the raw value, i.e. the milliseconds value due to just blinding parsing the settings json.

rbergen:
> That's perfectly fine. I wasn't sure the best way to handle devices already set less than 1s. I guess I can just round it up so next set will be to >1s.

If I recall correctly, @robertlipe mentioned in #411 that setting the value to less than a second effectively locks up the device, so the chances that anyone has such a value configured seem small. Also, if I check the interval value when it's set, I'll also do so when deserializing the effect manager from JSON. That way the interval will never be between 1 and 999 on a device that's running a firmware version with my update in place.

> It's also worth noting the setting dialog render the raw value, i.e. the milliseconds value due to just blinding parsing the settings json.

Yes, I'm aware. Which means that people can still configure the interval with ms precision if they really want to - as long as it's 1,000ms or more.

KeiranHines:
I was able to run successful at 600ms although not really with much display. However I agree with the backend adjustments. I'll remove support for setting <1s.

rbergen:
@KeiranHines I think the UI itself looks a lot better this way. In my casual testing I have noticed a few times that the timer countdown got out of sync with what was happening on the device (as in, which effect was showing) in weird ways. It would:

- count down to a point, say 10 seconds
- retrack the countdown a number of seconds, say jump back to 20 seconds
- count down all the way to 0, while the effect on the device would actually already change _during_ this countdown - say at 10 seconds 
- start the countdown on the next effect - the one the device already changed to - already some seconds (say 10 seconds) in

and keep repeating that cycle.

I am sure I've seen this happen when the effect interval was set to 30 seconds.

I was wondering if this could be related to you no longer using Promises when interacting with the API? Closely related, why did you make that change?

KeiranHines:
There was one call to sync() removed that might be the cause. I'll take a closer look on the weekend. Fetch actually returns a promise so the old logic was just wrapping a promise in a promise and wasn't always waiting for the fetch promise before doing other work.

rbergen:
Clear, thanks for responding so quickly. I'll await further updates about the sync thing.

KeiranHines:
Corrected all feedback @rbergen The bug mentioned above was caused by the way the timer was managed. Previously the remaining time would trigger a move to the next effect based on what the last activeTimer it had available. Now it triggers a resync, so if the information is out of date it will simply update and schedule a new timer. Please let me know if you have any further issues.

rbergen:
@KeiranHines Thanks for the update. I have not been able to reproduce the issue with this change in.

rbergen (approved):
LGTM!

PR #520, opened 2023-11-16
Feature/new weather icons

mggates39:
## Description
Adding a more colorful set of icons for the weather.  Implemented day and night versions.  Fixed an issue in tomorrow's minimum and maximum temperatures.

I have kept this branch up to date with all changes that have been merged into main to ensure no issues with the current code base.

This should close issue [#256](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/256)

## Contributing requirements
* [x] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [x] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [x] I selected `main` as the target branch.
* [x] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #537, opened 2023-11-21
Just Better Weather Icons

mggates39:
## Description
This pull request is to begin to address issue [256](https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/256) for better weather icons.
- Add a more colorful set of icons for the weather. 
- Implemented day and night icons for the current weather.
- Clean  up initial default settings for the weather.

This pull request supersedes pull requests [520](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/520) and [530](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/530) to better address the issues that came up during the code review.

Remember, code reviews are about the code, not the person.  Leave you ego at the door.

## Contributing requirements


mggates39:
I think this PR is the first minimal set and is ready for review.

rbergen:
@mggates39 Noted. It's just a case of unfortunate timing: you happened to open the PR just before I have a few busier days ahead (work, other things). I'll look at it when I can.

include/secrets.example.h
```diff
@@ -37,5 +37,7 @@
 #define cszOpenWeatherAPIKey ""                     // Your OpenWeatherMap API key goes Here
 #define cszLocation          "98074"
 #define bLocationIsZip       true
+#define bUse24HourClock      false
+#define bUseCelsius          false
```
rbergen, line 41: I don't think we need to, or should, add these. These settings are only used on devices/projects that have the web UI enabled, and that is the way to change the default. Coming from a different angle, there's a whole heap of other settings for which secrets(.example).h doesn't contain a define. If I'm totally honest I considered taking defines _out_ of here while I was working on the first iteration of `DeviceConfig`, but I figured grandfathering what was already there anyway was ok.

include/deviceconfig.h
```diff
@@ -65,12 +65,12 @@ class DeviceConfig : public IJSONSerializable
...
     String  location = cszLocation;
-    bool    locationIsZip = false;
+    bool    locationIsZip = bLocationIsZip;
     String  countryCode = cszCountryCode;
     String  timeZone = cszTimeZone;
     String  openWeatherApiKey = cszOpenWeatherAPIKey;
-    bool    use24HourClock = false;
-    bool    useCelsius = false;
+    bool    use24HourClock = bUse24HourClock;
+    bool    useCelsius = bUseCelsius;
```
rbergen, line 73: In line with my comment about secrets.example.h, I think it's better to leave this as it is now.

include/effects/matrix/PatternWeather.h
```diff
@@ -227,10 +255,9 @@ class PatternWeather : public LEDStripEffect
                     if (main["temp_min"] > 0)
                         lowTemp         = KelvinToLocal(main["temp_min"]);
 
-                    String iconIdTomorrow = entry["weather"][0]["icon"];
-                    iconTomorrow = iconIdTomorrow.toInt();
+                    iconTomorrow = String(entry["weather"][0]["icon"]);
```
rbergen, line 258: The explicit construction of the String is not necessary; ArduinoJson is pretty clever when it comes to figuring out what C++ type/class to convert to.

include/effects/matrix/PatternWeather.h
```diff
@@ -271,9 +298,8 @@ class PatternWeather : public LEDStripEffect
             highToday   = KelvinToLocal(jsonDoc["main"]["temp_max"]);
             loToday     = KelvinToLocal(jsonDoc["main"]["temp_min"]);
 
-            String iconIndex = jsonDoc["weather"][0]["icon"];
-            iconToday = iconIndex.toInt();
-            debugI("Got today's temps: Now %d Lo %d, Hi %d, Icon %d", (int)temperature, (int)loToday, (int)highToday, iconToday);
+            iconToday = String(jsonDoc["weather"][0]["icon"]);
```
rbergen, line 301: Same comment.

rbergen (commented):
@mggates39 I found the time now. I have a few comments, but much smaller than before. Thanks again for changing your approach on this!

include/secrets.example.h
mggates39: The Celsius setting is used in weather and I promoted it to secrets so that I could change it without affecting my commit set as I have to run it without the web UI to get it to fit in my paltry 4MB board.  I can certainly revert this.

include/deviceconfig.h
mggates39: Same comment as with secrets.example.h

include/effects/matrix/PatternWeather.h
mggates39: That's what I thought too, but I get this error without it.
include/effects/matrix/PatternWeather.h:258:62: error: ambiguous overload for 'operator=' 
Turns out it is my lack of familiarity with the ArduinoJson library. turns out you just need to use .as<String>() to do it directly with the library.

include/effects/matrix/PatternWeather.h
mggates39: Same comment. I found the correct way to do the type coercion.

mggates39:
@rbergen Thank you for the feedback.  I now have a better understanding of the ArduinoJson library.

rbergen:
You're welcome. And indeed, in the case of addressing ambiguity between `const char *` and `String` the right approach is indeed `as<`_type_`>()`; technically the `String(`_JsonVariant_`)` approach is still ambiguous, even though the compiler suddenly feels entitled to choose on your behalf when you use it.

PR #541, opened 2023-11-23
Tweaked ColorFillEffect to optionally ignore global color

revision29:
Tweaked ColorFillEffect to optionally ignore global color

## Description

Instead of creating a new color fill effect that ignores the global color, I tweaked the existing effect to optionally ignore.
I also updated the label comment for the effect since it was wrong.

## Contributing requirements

### DIFF
--- a/include/effects/strip/misceffects.h
+++ b/include/effects/strip/misceffects.h
@@ -214,7 +214,7 @@ protected:
 //
 // Fills the pixels with a single color.
 // everyNth can be used to light some pixels with specified color, leaving the others unlit.
-// Unless a user chooses to ignor the global color, the glboal color will be used instead when
+// Unless a user chooses to ignore the global color, the global color will be used instead when
 // DeviceConfig().ApplyGlobalColors() returns true.
 
 class ColorFillEffect : public LEDStripEffect
@@ -267,7 +267,7 @@ protected:
     {
         if (_everyNth != 1)
           fillSolidOnAllChannels(CRGB::Black);
-        if (_ignoreGlobalColor == false && g_ptrSystem->DeviceConfig().ApplyGlobalColors() == true)
+        if (!_ignoreGlobalColor && g_ptrSystem->DeviceConfig().ApplyGlobalColors() == true)
           fillSolidOnAllChannels(g_ptrSystem->DeviceConfig().GlobalColor(), 0, NUM_LEDS, _everyNth);
         else
           fillSolidOnAllChannels(_color, 0, NUM_LEDS, _everyNth);
--- a/include/effects/strip/misceffects.h
+++ b/include/effects/strip/misceffects.h
@@ -35,6 +35,7 @@
 #include "TJpg_Decoder.h"
 #endif
 #include "effects.h"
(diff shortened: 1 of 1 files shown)


revision29:
I had to cancel my last pull request because the commits got messy and GitHub did not ignore entries in .gitignore. What use is .gitignore if it is ignored?

rbergen:
@revision29 This _may_ not be what happened in this case, but adding a file to .gitignore after it's already been committed doesn't work. You then also have to run `git rm --cached <filename>` to make git stop tracking the file in question, or run `git rm -r --cached .` if you want to "untrack" all files in .gitignore. I'm sure there are equivalent actions in GitHub Desktop, but I can't tell you what those are.

include/effects/strip/misceffects.h
```diff
@@ -260,7 +267,10 @@ class ColorFillEffect : public LEDStripEffect
     {
         if (_everyNth != 1)
           fillSolidOnAllChannels(CRGB::Black);
-        fillSolidOnAllChannels(_color, 0, NUM_LEDS, _everyNth);
+        if (_ignoreGlobalColor == false && g_ptrSystem->DeviceConfig().ApplyGlobalColors() == true)
```
rbergen, line 270: I won't hold back approval over this, but I'd consider using the shorter version because the line is already pretty long:

```suggestion
        if (!_ignoreGlobalColor && g_ptrSystem->DeviceConfig().ApplyGlobalColors())
```

include/effects/strip/misceffects.h
```diff
@@ -209,10 +210,12 @@ class RainbowFillEffect : public LEDStripEffect
...
 };
 
-// RainbowFillEffect
+// ColorFillEffect
 //
-// Fills the spokes with a rainbow palette
-
+// Fills the pixels with a single color.
+// everyNth can be used to light some pixels with specified color, leaving the others unlit.
+// Unless a user chooses to ignor the global color, the glboal color will be used instead when
```
rbergen, line 217:
```suggestion
// everyNth can be used to light some pixels with the specified color, leaving the others unlit.
// Unless a user chooses to ignore the global color, the global color will be used instead when
```

rbergen (commented):
Thanks, this is much cleaner. 

Feedback-wise I found a couple of typos and one small suggestion for readability.

revision29:
I have always been bad at spelling and have grown dependent on spellcheck. IDEs don't check for misspelled words.
I normally opt for the longer boolean checks because I often make stupid errors in observation. "== false" is easier for me to spot than "!". But, I fixed and pushed a new commit.

rbergen:
That's not a problem. I've always been pedantic about spelling and grammar (and still get it wrong from time to time), so that seems to be an area we can complement each other. :)

The use of `!` vs `== false` is technically only a style thing as both are correct, and I would have accepted it if you'd said "yeah, well, I'll stick with what I have." I only commented on it this time because using the shorter notation increases the chances that the reader sees the whole Boolean expression in one glance.

rbergen (approved):
LGTM!

PR #542, opened 2023-11-24
Effect manager global color and palette fire tweaks

revision29:
### Effect manager
Modified the effect manager to change how it applies global palette colors for regular light strips.
- For configurations that have enabled audio and use spectrum analyzer, it will create a spectrum analyzer based on the global colors.
- Other configurations will just show solid the global fill color.
- Also changed where `auto& deviceConfig = g_ptrSystem->DeviceConfig();` was defined.

### Palette fire effects
Modified the palette fire and musical palette fire effects to integrate the global fill color.
- Both have an optional ignore global color argument.
- If ApplyGlobalColors is true, it will use the global colors for palette indexes 1 and 2 (second and third colors) in keeping with how effect manager used to create a temporary effect.

### IR Remote
Modified IR remote code to change how it handles clearing the temporary effect.
- Pressing the buttons corresponding to next effect and previous effect will check to see if global colors are being applied. 
- If ApplyGlobalColors is true it will set applyGlobalColors to false and re-activate the effect that was active before.
- If ApplyGlobalColors is false it will either show next or previous effect.

The overall purpose is to show the global color when a button is pressed. This can be used for cases where a person just wants to have one color on, overriding any effects. It also can be used to "preview" the color being applied to an effect that integrates the global color choice. Pressing the color will display the color on all pixels. Then pressing the previous or next effect button will return back to that effect and a person can now see the new global color taking effect.

Note: even though GH claims I modified effectmanager.cpp near lines 395-411 (the part with if (USE_HUB75)), I did not modify any code there. I changed the indent level at one point but changed it back. That might be what it is reporting as a change.

## Contributing requirements


include/effects/strip/fireeffect.h
```diff
@@ -226,14 +229,16 @@ class PaletteFlameEffect : public FireEffect
...
+          _ignoreGlobalColor(ignoreGlobalColor)
     {
         construct();
     }
 
     PaletteFlameEffect(const JsonObjectConst& jsonObject)
       : FireEffect(jsonObject),
-        _palette(jsonObject[PTY_PALETTE].as<CRGBPalette16>())
+        _palette(jsonObject[PTY_PALETTE].as<CRGBPalette16>()),
+        _ignoreGlobalColor(jsonObject["igc"])
```
rbergen, line 241: Now that "igc" is used in multiple effects with a defined meaning, I think it's best to add a definition (`PTY_IGNOREGLOBALCOLOR` seems sensible) to effects.h for it, and use that in all 3 effects that now use "igc".

src/effectmanager.cpp
```diff
@@ -392,39 +392,29 @@ void EffectManager::ApplyGlobalPaletteColors()
...
+        {
+            CHSV hsv = rgb2hsv_approximate(globalColor);
+            pMatrix->setPalette(CRGBPalette16(globalColor, CRGB(CHSV(hsv.hue + 64, 255, 255))));
+        }
+        else
+        {
+            // But if we have two different colors, we create a palettte spread between them
+            pMatrix->setPalette(CRGBPalette16(secondColor, globalColor));
+        }
+        pMatrix->PausePalette(true);
```
rbergen, line 410: I understand why you're doing this, but if you change nothing in the code of a block then please don't modify the white spacing either.

src/remotecontrol.cpp
```diff
@@ -72,32 +72,45 @@ void RemoteControl::handle()
     lastResult = result;
 
     auto &effectManager = g_ptrSystem->EffectManager();
+    auto& deviceConfig = g_ptrSystem->DeviceConfig();
```
rbergen, line 75: Now that the declaration of `deviceConfig` is immediately under `effectManager`, I think it's best to align the `auto` and the `&` to it:
```suggestion
    auto &deviceConfig = g_ptrSystem->DeviceConfig();
```

src/remotecontrol.cpp
```diff
@@ -72,32 +72,45 @@ void RemoteControl::handle()
...
-        effectManager.ClearRemoteColor();
-        effectManager.NextEffect();
+        if (deviceConfig.ApplyGlobalColors())
+        {
+            effectManager.ClearRemoteColor();
+            effectManager.Update();
+        } else
+        {
+            effectManager.NextEffect();
+        }
```
rbergen, line 100: This makes the behavior of IR_BPLUS (and IR_BMINUS below) depend on whether global colors are currently applied or not. Is this what you intend to do?

Also, please:

- Don't call `effectManager.Update()` directly; it will be called soon enough by the main drawing loop in drawing.cpp, - which is what that function is intended for.
- Put `else` on a line of its own.

src/remotecontrol.cpp
```diff
@@ -72,32 +72,45 @@ void RemoteControl::handle()
...
-        effectManager.ClearRemoteColor();
-        effectManager.PreviousEffect();
+         if (deviceConfig.ApplyGlobalColors())
+        {
+            effectManager.ClearRemoteColor();
+            effectManager.Update();
+        } else
+        {
+           effectManager.PreviousEffect();
+        }
```
rbergen, line 112: See comment above.

rbergen (commented):
Thanks for submitting this. I have a few comments and one question, see below.

rbergen:
I now see that I originally missed part of your original description, which answers one question I asked in my comments.

Considering the description you gave:

- I'll run the "algorithm change" you propose in how the BPLUS and BMINUS buttons work past Dave, because of the two of us he is the avid IR remote control user - I hardly ever do. The other comments I made concerning the remotecontrol.cpp changes still stand in any case.
- The indentation in the HUB75 block in effectmanager.cpp _did_ change. In the current codebase it is indented one level deeper than it is in this PR.

src/remotecontrol.cpp
revision29: Yes, that is the desired behavior. Before I touched it, it stopped showing the temporary effect and went to the next effect in the queue. This minor change makes it so that it clears the temporary effect so that a person returns to the state before pressing the color button. It's the most logical way I have come up with yet to arrive at that result.

Regarding the call to the update function, that is good to know. I didn't know exactly when it was called, but figured it would be better to explicitly trigger. If it is not needed it will be eliminated.

revision29:
Regarding the indentation: I am still getting used to formatting around the preprocessor conditions. I remember that I changed it because it looked one intent too shallow. But, I need to think of the formatting as if the preprocessor statements are not there and format the code inside the conditions accordingly. How should the preprocessor statements be indented? Should they be one level less than the surrounding code, or should those statements have no indents?

revision29:
To summarize the PR: I am trying to tackle three related improvements with one pull.
- Make effect manager not create a fire effect and just fill the global color.
- Update palette flame effect and the musical variant to respect the global color.
- Make minor remote code tweaks to make clearing applyGlobalColors more intuitive. 

I realize that adding the variable "ignoreGlobalColor" won't be ideal for all effects that we want to respect the global color, but I think it is a good option for people that don't ever want a particular effect using the global color. Getting it back to the effect's original state might be too cumbersome with en IR remote.

For the two projects I have in the real world, and a third to be installed in the next week, the IR remote is the primary input mechanism. These are intended for people other than myself to change things without having to open a webpage. So, I am trying to think about my use cases while not breaking how Dave or others might be using the IR interface. I want things to work a certain way, but I want it to work as intuitively as possible for the people who are using the remote who have no clue what ws2812b is or that C++ exists.

include/effects/strip/fireeffect.h
revision29: Re: PTY_IGNOREGLOBALCOLOR not a bad idea. I will make the change and push a new commit.

src/remotecontrol.cpp
revision29: Hey, all I did was cut and paste someone else's line of code and didn't notice the inconsistency. Maybe the person who originally typed it should have respected the style already implemented in the file... :)  Honestly I didn't notice when I pasted the line, but now that you pointed it out, it make my eye twitch. One thing I can say is there is no consistency in where to put the `&` in this codebase. I prefer to put it by the variable name, but others like to attach it to the type declaration. I am here to learn whichever way is "right."

src/effectmanager.cpp
revision29: This was an artifact that occurred when I was deciding how to implement the change I wanted. I originally wrapped that section in an if else statement. But, I quickly realized that was a stupid idea and removed the conditional. Now that I am looking at the original upstream file, I notice there is an extra indent in that section. The block in the if section does not match the indent in the else section. Since I am already proposing change to the file is it okay to also "fix" the indent?

rbergen:
> Regarding the indentation: I am still getting used to formatting around the preprocessor conditions. I remember that I changed it because it looked one intent too shallow. 

No, it currently actually _is_ one level _too deep_, which your PR would correct. It's just that whitespace-only changes are very noisy in PRs, so unless the legibility of the code is severely impacted we just leave it alone, even if it's "wrong". If we ever do need to change the logic in the HUB75 block, we can then incorporate this whitespace fix as well.

> But, I need to think of the formatting as if the preprocessor statements are not there and format the code inside the conditions accordingly.

Generally speaking, I treat preprocessor #if/#else/#endif/... etc as "normal" if/else/... statements. That is: the #if/... preprocessor statements align with the code around them, and the code between them is indented one level deeper. So again, the indentation change your PR now proposes would constitute a correction, but we just don't do that unless there's also an actual code change to be made.

> To summarize the PR: I am trying to tackle three related improvements with one pull.

Yes, I got that now. Most of it was clear from the code, and the third got confirmed after I read the opening description. By the way, Dave responded that he's happy to try your update to how + and - work "in production", so that change we can proceed with.

> I realize that adding the variable "ignoreGlobalColor" won't be ideal for all effects that we want to respect the global color, but I think it is a good option for people that don't ever want a particular effect using the global color. 

Of course, they have to set the flag when constructing the effect (or actually, when putting its factory together in effects.cpp), so that does require a code change.

src/effectmanager.cpp
rbergen: No, please don't. The block itself is consistently indented, so it's indented to deep but not confusing. This may not make much sense to you, but I have to review a **lot** of PRs, so reducing whitespace noise really saves me a lot of energy - and sanity.

src/remotecontrol.cpp
rbergen: Both styles are used in the codebase and that's ok, until two lines are so close to each other that it looks weird to use one in one line and the other in another. Which I think is clearly the case when one line is line 74 and the other 75. :)  
It only stands out now because the deviceConfig declaration got moved a whole lot higher than it was.

include/effects/strip/fireeffect.h
```diff
@@ -283,8 +300,9 @@ class MusicalPaletteFire : public PaletteFlameEffect, protected BeatEffectBase
                        int sparkHeight = 3,
                        bool reversed = false,
                        bool mirrored = false)
-        : PaletteFlameEffect(strName, palette, ledCount, cellsPerLED, cooling, sparking, sparks, sparkHeight, reversed, mirrored),
-          BeatEffectBase(1.00, 0.01)
+        : PaletteFlameEffect(strName, palette, _ignoreGlobalColor, ledCount, cellsPerLED, cooling, sparking, sparks, sparkHeight, reversed, mirrored),
+          BeatEffectBase(1.00, 0.01),
+          _ignoreGlobalColor(ignoreGlobalColor)
```
rbergen, line 306: It just occurred to me that `MusicPaletteFire` subclasses `PaletteFlameEffect`, which also holds and persists `ignorelGlobalColor`. If you pass `ignoreGlobalColor` to the `PaletteFlameEffect` constructor (as you do), then there's no need to put it in a variable or persist it to JSON in `MusicPaletteFire` as well.

src/effectmanager.cpp
mggates39: Does your review tool not have a configuration to ignore whitespace?

rbergen: That's not the point. Whitespace _does_ matter, and changes to it _should_ be an improvement. We just don't merge big blocks of whitespace-only changes, and they do distract from other stuff.

include/effects/strip/fireeffect.h
revision29: Thanks for the suggestion. Still I am still learning a lot, I had no idea how much to explicitly declare in the subclass.

revision29: General observation: While I like the ability to converse about specific code sections, it is real hard to keep track of the conversation. So, forgive me for being redundant at times and not missing important things that are pointed out from time to time.

src/effectmanager.cpp
revision29: At this point, I don't know what the spacing was. But, the last few commits have made the spacing what it should be to conform to the spacing standards of the project. Unless I am missing something obvious.

revision29:
Okay, I have made one last commit. Removed the redundant variable for the child class. But, there is another small but very important change.

When I went to compile the code for the umbrella project, it failed. There was a type error with the palette variable. More specifically, it was defined for the classes as a reference variable. But, in the effects.cpp, some instances of the MusicalPaletteFire were created with an inline construction of a palette. Thus, it is passes an instance of CRGBPalette16 instead of a reference to one. I don't know how that ever compiled for anyone. I know this potentially causes more memory to be consumed. But, in that section 9 of the 10 MusicalPaletteFire instances use inline defined palettes. Only one uses a pre-existing palette. So, yes, a copy of a pre existing palette is made, but that is nothing compared to the 9 new palettes taking up memory.

I could be wrong, but I am not smart enough to argue with the compiler. But, someone else might have a better fix.

rbergen:
I've just pushed some commits myself to move this PR forward. I'll list them below, but first:

> While I like the ability to converse about specific code sections, it is real hard to keep track of the conversation.

I don't know if you already do this, but if not: I find it easiest to use the Files Changed tab on the GitHub webpage for the PR. It shows which files have changed, highlights changes made and also shows conversations that are attached to code, in context.

> When I went to compile the code for the umbrella project, it failed. There was a type error with the palette variable. More specifically, it was defined for the classes as a reference variable. But, in the effects.cpp, some instances of the MusicalPaletteFire were created with an inline construction of a palette. Thus, it is passes an instance of CRGBPalette16 instead of a reference to one. I don't know how that ever compiled for anyone.

I can tell you why the compiler started complaining. It's because you removed the `const` in front of the `CRGBPalette16 &palette` parameter in the constructor declarations of `PaletteFlameEffect` and `MusicalPaletteFire` in your first commit (this one: https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/542/commits/c7016b05e32675461301976db181511d67cd3864). I don't know why you did that, but putting it back made the Umbrella builds work again - as they have been for a very long time.

The changes I pushed were the following:

- I reverted the indentation change we've been going back and forth about
- I removed the `_ignoreGlobalColors` variable that was still lingering in `MusicalPaletteFire`
- I restored the `const` in front of the `CRGBPalette16 &palette` constructor parameters, as mentioned
- I changed the way the "temporary" palette is created in `PaletteFlameEffect::GetBlackBodyHeatColor()`: it now creates a palette that stretches between the two global colors, like the "global palette" is made for HUB75 builds - which is how the two global colors are intended to be used.

@revision29 Let me know if you have any objections to these updates to your PR. I'll note for the record that it's quite possible I'll have to sort out a merge conflict if #543 gets merged first (which would require one more commit/push), but I'll take care of that if and when it turns out to come up.

revision29:
I am mostly fine with the changes. I don't know why I deleted the `const` keyword. I do remember getting a compile error trying to modify the palette directly. I realized my mistake and then make the tempPalette variable. Stupid mistake. 

But, I did choose to modify the specific index numbers for a reason. I don't disagree with the change. When effect manager was using the effect to draw a custom fire effect, it was using black and white in the colors. The black allows for some pixels to be turned off while the fire jumps around. I don't know what it looks like for real since I don't have hardware to match, but generally in the world of reactive audio visualizations, you want black to be used to indicate a decrease in volume or silence. So, my suggestion is to mostly keep your change in place, but add a CRGB::Black into the palette. Otherwise, your approach is better.

rbergen:
In that case I suggest we just replicate the way `EffectManager` (currently still) puts the palette together. That is, make line 268 in fireeffect.h this:

```cpp
            auto tempPalette = CRGBPalette16(CRGB::Black, deviceConfig.GlobalColor(), CRGB::Yellow, CRGB::White);
```

That will create a palette that smears out over the four mentioned colors (one being the primary global color), while still looking "natural" from a flame temperature perspective.

Agreed?

revision29:
`EDIT: I've also made SpectrumAnalyzerEffect adopt the global colors, so we can get rid of the #if SPECTRUM && ENABLE_AUDIO block too - which means no specific "global color effects" are created any more.`
Nice catch. Everything looks great.

PR #549, opened 2023-11-28
Centralize port numbers in source

robertlipe:
Instead of scattering port numbers in source, headers, or macros, we
stuff them into one grep-able enum.

Tested:
Touch test on (out-of-tree) Demo project with webserver, debugger turned on.
Successful buddybuild
Not tested:
Commodore pet rock. (Wat!?!)

## Contributing requirements

### DIFF
--- a/include/network.h
+++ b/include/network.h
@@ -33,6 +33,18 @@
     #include "socketserver.h"
 #endif
 
+    // For now, just a centralized location for the port numbers for our
+    // various services. Someday these might be configurable.
+    // This could be an enum class, but the static_cast<int> at the
+    // callers is ugly.
+    enum NetworkPort
+    {
+      ColorServer  = 12000,
+      IncomingWiFi  = 49152,
+      VICESocketServer = 25232,
+      Webserver  = 80
+    };
+
 #if ENABLE_WIFI
     enum class WiFiConnectResult
     {
--- a/include/webserver.h
+++ b/include/webserver.h
@@ -50,8 +50,9 @@
 #include <ArduinoJson.h>
 #include <HTTPClient.h>
 #include "deviceconfig.h"
-#include "jsonbase.h"
 #include "effects.h"
+#include "jsonbase.h"
+#include "network.h"
 
 class CWebServer
 {
@@ -202,7 +203,7 @@ class CWebServer
   public:
 
     CWebServer()
-        : _server(80), _staticStats()
+        : _server(NetworkPort::Webserver), _staticStats()
     {}
 
     // begin - register page load handlers and start serving pages
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -36,6 +36,10 @@
 #include <esp_task_wdt.h>
 #include "soundanalyzer.h"
 
+#if ENABLE_VICE_SERVER
+#include "network.h"
+#endif
+
 // AudioSamplerTaskEntry
 // A background task that samples audio, computes the VU, stores it for effect use, etc.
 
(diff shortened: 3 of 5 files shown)


include/network.h
```diff
@@ -33,6 +33,18 @@
     #include "socketserver.h"
 #endif
 
+    // For now, just a centralized location for the port numbers for our
+    // various services. Someday these might be configurable.
+    // This could be an enum class, but the static_cast<int> at the
+    // callers is ugly.
+    enum NetworkPort
```
rbergen, line 40: Why aren't we making this an enum class to keep our global namespace clean? It seems you use the enum name prefix already, anyway.

rbergen (commented):
Thanks for this, I think that's a nice and tidy clean-up. I have one comment/question, see below. And then there's this:

> Commodore pet rock. (Wat!?!)

Careful there, sir. When all's said and done and I'm totally honest, that's probably still my favorite project in the entire PSL collection. :)

include/network.h
rbergen: > Besides, Look in globals.h and tell me really care about clean global namespaces. :-)

We do. That's why we've been grunting at globals.h for ages already, and I've spent quite a lot of time cleaning and bundling global variables in the past - I have my eye on a few more that may undergo the same treatment.

> We could use std::to_underlying, but that's C++23, which we're not exactly prepared to embrace yet. (I'm still working on 20...)

There actually is a "constexpr to_value()" in jsonserializer.h that does the same, which we could move to types.h - which is already included in network.h anyway.

But I agree, it's suboptimal A vs suboptimal B. I'm fine leaving it like this to reduce the hassle. 

> Now that I found the source for RemoteDebug (and files with leading dot are, by definition, hidden)

:) From a file system perspective on UN*X, yes. That's not how I interpreted your statement.

> If you nominate adding the "23" in there with an entry in here, we'll make RemoteDebug depend on our network.h (that may or may not play out well...) 

No, I don't like doing that directly. What we could do is turn the current telnet port define in RemoteDebug into an #ifndef/#define, and define it just above the RemoteDebug include.
To be totally honest on the other hand, 23 has been the telnet port for a very long time, so I'm actually also fine with just leaving it there.

rbergen:
Alright, I've tumbled it over in my head a few more times, and I think we can leave the enum an enum, instead of making it an enum class. I suddenly remembered that the port numbers are actually integers, and enum class is more appropriate for actual _identifiers_; they just "happen to have" this or that as an underlying type.

Which means we can also leave to_value() where it is, for now. Who knows, maybe another use case turns up later besides JSON deserialization.

...which is why I figured using en enum class would be a better option for GIFIdentifier in the context of PatternAnimatedGIF, and then ran into the crashes reported in #547. It's an interesting morning - but I digress.

I would like to make the telnet port number configurable (yes, telnet _is_ port 23, but http is 80 and we rightfully also make that configurable), so if you'd be willing to pick that up then I'd appreciate that.

include/network.h
rbergen: My preference for inclusion of the telnet port was based on our mutual earlier belief that doing so would be pretty straightforward. As I now understand it that conclusion no longer stands, which changes things of course. Reading your explanation we could dodge this by sticking the enum in a different header file that does not introduce circular dependencies, but we'd always still have a situation where a dependency end up depending on a project internal header - which is something I failed to realize the last time we discussed this. That's just doing things the wrong way around.

So yes, I'll review one more time with this knowledge in mind.

Thanks for digging into this and providing this level of clarity on the matter!

PR #552, opened 2023-11-30
Effect manager can now receive a temporary effect

revision29:
## Description

This allows a user to send an effect to effect manger to be used as a temporary effect. This can be done by using a remote control button to create and then populate the effect. For example: FADE3 might be used to trigger a palette effect of 3 specified colors. Another use would be for a user to be able to show a color as a temp effect when pressing a color button since effect manager no longer shows a temporary effect when setting global color. 

Added three self-explanatory methods:
- SetTempEffect
- ClearTempEffect
- HasTempEffect

## Contributing requirements


rbergen:
As [mentioned here](https://github.com/PlummersSoftwareLLC/NightDriverStrip/discussions/545#discussioncomment-7679703) I'm going to hold on merging this until a number of related points have been addressed/demonstrated, first.

revision29:
@rbergen A valid use case has been added as an example. Turns out that it is probably a decent example. I discovered that the current IR remote code assumes that audio is enabled and triggers an audio effect (ShowVU) when a certain button is pressed. When audio is disabled, I send a temp effect.

PR #563, opened 2023-12-07
Add nycancat, palette support for hypnosis

davepl:
Adds nyancat animation
Adds tesseract animation
Updtes hypnosis effect with palette support
Minor fixes


include/effects/matrix/PatternSMHypnosis.h
```diff
@@ -47,6 +52,6 @@ class PatternSMHypnosis : public LEDStripEffect
         t += 4;
         for (uint x = 0; x < MATRIX_WIDTH; x++)
             for (uint y = 0; y < MATRIX_HEIGHT; y++)
-                g()->leds[XY(x, y)] = ColorFromPalette(RainbowStripeColors_p, t / 2 + rMap[x][y].radius + rMap[x][y].angle, sin8(rMap[x][y].angle + (rMap[x][y].radius * 2) - t));
+                g()->leds[XY(x, y)] = ColorFromPalette(g()->IsPalettePaused() ? g()->GetCurrentPalette() : RainbowStripeColors_p, t / 2 + rMap[x][y].radius + rMap[x][y].angle, sin8(rMap[x][y].angle + (rMap[x][y].radius * 2) - t));
```
davepl, line 55: OK, this line is getting even more unwieldly, but...

rbergen: Yeah, it's becoming quite hard to follow with all the symbols. Maybe cut it up in smaller parts?

```suggestion
                g()->leds[XY(x, y)] = ColorFromPalette(g()->IsPalettePaused() 
                    ? g()->GetCurrentPalette() 
                    : RainbowStripeColors_p, t / 2 + rMap[x][y].radius + rMap[x][y].angle, sin8(rMap[x][y].angle + (rMap[x][y].radius * 2) - t));
```

Or go the whole way, and make a good ol' if/else out of it.

include/effects/matrix/PatternSwirl.h
```diff
@@ -73,27 +73,25 @@ class PatternSwirl : public LEDStripEffect
...
         graphics->leds[graphics->xy(i + 1, j)] += color;
         color.maximizeBrightness();
-        graphics->leds[graphics->xy(i, j)] += color;
+        graphics->leds[XY(i, j)] += color;
     }
 
     void Draw() override
     {
-        auto graphics = _GFX[0];
-
```
rbergen, line 89: I'd change this to set `graphics` to `g()`, and then keep using `graphics` further down in this function. Contrary to `drawAt()`, where I'd indeed remove `graphics` and call `g()` directly on line 83.

platformio.ini
```diff
@@ -19,7 +19,7 @@
 ; mesmerizer    HUB75 info panel with audio effects, weather, info, etc,
 
 [platformio]
-default_envs =
+default_envs =
```
rbergen, line 22: Trailing space.

platformio.ini
```diff
@@ -35,7 +35,7 @@ monitor_port    =
 build_type      = release
 upload_speed    = 1500000
 build_flags     = -std=gnu++2a
-                  -Dregister=                       ; Sinister: redefine 'register' so FastLED can use that keyword under C++17 and later
+                  -Dregister=                       ; Sinister: redefine 'register' so FastLED can use that keyword under C++17
```
rbergen, line 38: I think the "...and later" is more accurate. This could make it even clearer:

```suggestion
                  -Dregister=                       ; Sinister: redefine 'register' so FastLED can use that keyword under C++17 and later C++ versions
```

src/remotecontrol.cpp
```diff
@@ -47,7 +47,7 @@ void RemoteControl::handle()
     uint result = results.value;
     _IR_Receive.resume();
 
-    debugV("Received IR Remote Code: 0x%08X, Decode: %08X\n", result, results.decode_type);
+    debugW("Received IR Remote Code: 0x%08X, Decode: %08X\n", result, results.decode_type);
```
rbergen, line 50: Is it appropriate to make every IR button press issue a Warning level debug message? I'd say Verbose feels more suitable.

src/effects.cpp
```diff
@@ -205,18 +221,21 @@ void LoadEffectFactories()
...
         ADD_EFFECT(EFFECT_MATRIX_SMGAMMA,           PatternSMGamma);
-        ADD_EFFECT(EFFECT_MATRIX_ANIMATEDGIF,       PatternAnimatedGIF,     "Rings", GIFIdentifier::ThreeRings);
+        ADD_EFFECT(EFFECT_MATRIX_ANIMATEDGIF,       PatternAnimatedGIF,     "Rings",       GIFIdentifier::ThreeRings);
         ADD_EFFECT(EFFECT_MATRIX_ANIMATEDGIF,       PatternAnimatedGIF,     "Atomic",      GIFIdentifier::Atomic);
         ADD_EFFECT(EFFECT_MATRIX_ANIMATEDGIF,       PatternAnimatedGIF,     "Bananaman",   GIFIdentifier::Banana,        true,               CRGB::DarkBlue);
         ADD_EFFECT(EFFECT_MATRIX_SMMETA_BALLS,      PatternSMMetaBalls);
         ADD_EFFECT(EFFECT_MATRIX_SMSUPERNOVA,       PatternSMSupernova);
         ADD_EFFECT(EFFECT_MATRIX_CUBE,              PatternCube);
+        ADD_EFFECT(EFFECT_MATRIX_ANIMATEDGIF,       PatternAnimatedGIF,     "Tesseract",   GIFIdentifier::Tesseract, false);
+        ADD_EFFECT(EFFECT_MATRIX_ANIMATEDGIF,       PatternAnimatedGIF,     "Nyancat",     GIFIdentifier::Nyancat);
```
rbergen, line 238: Maybe bump `EFFECT_SET_VERSION` for this `MESMERIZER` block to 6?

rbergen (commented):
I have a few comments/suggestions that aren't blocking for approval. I would suggest bumping the effect set version for Mesmerizer, though.

src/remotecontrol.cpp
rbergen: I maintain a "W" is too strong. A warning is something the viewer should mildly worry about. If "V" is too weak then I think "I" could be the compromise; that's still included in the default debug output if I'm not mistaken.

include/effects/matrix/PatternSMSupernova.h
```diff
@@ -39,8 +40,7 @@ class PatternSMSupernova : public LEDStripEffect
             }
 
             if (debris_item._is_shift && ParticlesUpdate(debris_item)) {
```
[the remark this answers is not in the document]
rbergen: Please do. :)

samples/videoserver/VideoServer2.py
```diff
@@ -30,7 +30,7 @@
 MATRIX_HEIGHT = 32
 FUTURE_DELAY = 5
 URL = "https://youtu.be/WPdCT0l6vGs"
-ESP32_WIFI_ADDRESS = '192.168.8.86'
+ESP32_WIFI_ADDRESS = '192.168.8.127'
```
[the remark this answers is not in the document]
rbergen: Not really, but as every user of the example would have to change this IP address anyway, it's a bit of a case of six of one, half a dozen of the other.

The actual solution here is making the target IP a script argument, as I think you yourself argued some time ago. Maybe one day this will annoy me to the point I will actually take care of that.

src/effects.cpp
```diff
@@ -177,9 +177,27 @@ void LoadEffectFactories()
       #include "custom_effects.h"
 
     // Fill effect factories
-    #elif DEMO
+    #elif M5DEMO
```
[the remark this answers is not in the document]
rbergen: It's a fair question, and I think one to keep in mind when we design and implement the Great Overhaul of platformio.ini/globals.h/effects.*. For the time being we're effectively grandfathering the fact that NDS started as Dave's personal project. One of the reasons for doing that is that we do not yet have a structural solution in place for people who use NDS to build multiple projects for multiple different devices in one local clone; the custom_* setup doesn't cover this.

rbergen:
I've made a few changes based on review comments (some of them my own), which means I now need to test the current state of the code before I can responsibly merge - and that on both Mesmerizer and the M5StickC Plus because of the changes made. I'll do so when I get a chance, which will be this weekend at the latest.

rbergen:
Tested on Mesmerizer and M5StickC Plus with Spectrum.

PR #574, opened 2023-12-24
Make secrets.h more immune to changes from secrets.example.h

robertlipe:
When new fields are added in the example file, add a test in the
file that uses the real file to be sure the fields are present. This
is to solve the case where a developer has a checked-out, working tree
with a working secrets.h, syncs to head or another branch that adds
fields in secrets, and then incurs weird compilation errors when those
new fields aren't found.

This change leads the developer to the correct file and the doc
describing it for a better experience.

## Description

As discussed in #572.

## Contributing requirements

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -35,9 +35,12 @@
 #include <tuple>
 #include <algorithm>
 #include "jsonserializer.h"
+#include "types.h"
+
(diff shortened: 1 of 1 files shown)


rbergen (approved):
Made some textual updates. Otherwise, LGTM!

PR #594, opened 2024-01-19
Allow for web UI endpoints to be disabled during build

Blackburn29:
## Description
- Adds build flags and a new base configuration to _optionally_ enable the React Web UI. 

When working with Heltec V3 boards, memory pressure is tight for HTTP requests memory due to the lack of PSRAM. The webserver is disabled by default but I figured it would be useful to allow users to be able to utilize the API endpoints without the need of the dedicated UI. 

(In my use-case I am building a standalone react app that can be self-hosted to manage many NightDriverStrip devices and I need API access but do not need the embedded UI)

I welcome any feedback on the platform IO build configuration as it is still pretty new to me and I want to make sure that I do not increase the build complexity too much

## Contributing requirements

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -527,7 +527,7 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
     #define INCOMING_WIFI_ENABLED       1   // Accepting incoming color data and commands
     #define WAIT_FOR_WIFI               0   // Hold in setup until we have WiFi - for strips without effects
     #define TIME_BEFORE_LOCAL           2   // How many seconds before the lamp times out and shows local content
-    #define ENABLE_WEBSERVER            1  // Turn on the internal webserver
+    #define ENABLE_WEBSERVER            1   // Turn on the internal webserver
     #define ENABLE_NTP                  1   // Set the clock from the web
     #define ENABLE_OTA                  1   // Accept over the air flash updates
     #define ENABLE_REMOTE               1   // IR Remote Control
@@ -1171,7 +1171,15 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 #endif
 
 #ifndef ENABLE_WEBSERVER
-#define ENABLE_WEBSERVER        0   // Chip provides a web server with controls to adjust effects
+    #define ENABLE_WEBSERVER        0   // Chip provides a web server with controls to adjust effects
+
+#endif
+
+
+#if ENABLE_WEBSERVER
+    #ifndef ENABLE_WEB_UI
+    #define ENABLE_WEB_UI           1   // Enable endpoints for the web UI
+    #endif
 #endif
 
 #ifndef ENABLE_OTA
--- a/include/globals.h
+++ b/include/globals.h
@@ -1171,14 +1171,12 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 #endif
 
 #ifndef ENABLE_WEBSERVER
-    #define ENABLE_WEBSERVER        0   // Chip provides a web server with controls to adjust effects
-
+#define ENABLE_WEBSERVER        0   // Chip provides a web server with controls to adjust effects
 #endif
-
 
 #if ENABLE_WEBSERVER
     #ifndef ENABLE_WEB_UI
-    #define ENABLE_WEB_UI           1   // Enable endpoints for the web UI
+    #define ENABLE_WEB_UI           1   // Enable HTTP pathnames for the web UI
     #endif
 #endif
 
--- a/platformio.ini
+++ b/platformio.ini
@@ -30,7 +30,8 @@ extra_configs =
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     =
+#upload_protocol = 
+#upload_port     = 
 monitor_port    =
 build_type      = release
 upload_speed    = 1500000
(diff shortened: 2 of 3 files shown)


rbergen:
@Blackburn29 I agree with the premise of @robertlipe's feedback, but as far as I'm concerned it can be implemented with the `ENABLE_` approach. That could be done by defining `ENABLE_WEB_UI` to 1 at the bottom of globals.h, if it's not yet defined _and_ `ENABLE_WEBSERVER` is 1. 

So again, I do agree that "web UI enabled" needs to be the default when the webserver itself is enabled. With that in place, only configurations that do want the webserver but not the web UI need to be changed to define `ENABLE_WEB_UI` to 0.

Blackburn29:
> @Blackburn29 I agree with the premise of @robertlipe's feedback, but as far as I'm concerned it can be implemented with the `ENABLE_` approach. That could be done by defining `ENABLE_WEB_UI` to 1 at the bottom of globals.h, if it's not yet defined _and_ `ENABLE_WEBSERVER` is 1.
> 
> So again, I do agree that "web UI enabled" needs to be the default when the webserver itself is enabled. With that in place, only configurations that do want the webserver but not the web UI need to be changed to define `ENABLE_WEB_UI` to 0.

How do you feel about the configuration I added in the platform config? Are you fine with that? It does add a level to complexity to the feature flag since you would need to change the base configuration of the device you are targeting in order to add/remove the embedded web files

rbergen:
@Blackburn29 To be honest I hadn't gotten there yet, but now that you ask: no, let's not do that. We've been including the webserver files in builds without the webserver enabled the whole time, and I think we should stick to that. With the gzip compression in place we're talking about ~23K of flash memory, and in my opinion we can do without saving that. It's RAM where we run into limitations at the moment, not flash.  
Obviously, we can reconsider this if and when builds start failing over lack of flash space.

Blackburn29:
@rbergen  I’m good with that. I’ll update the PR shortly

Blackburn29:
Also, here is the standalone app if anyone is interested. 

image

https://github.com/Blackburn29/NightDriverWeb

rbergen:
@Blackburn29 Is my understanding correct that you're still intending to push an update commit to reverse the changes to platformio.ini this PR now includes?

Blackburn29:
> @Blackburn29 Is my understanding correct that you still intend to push an update commit to reverse the changes to platformio.ini that this PR now includes?

Yes, apologies. I’ve been busy at work and haven’t had a chance to make the changes. Hoping to get to it by the weekend

rbergen:
@Blackburn29 Not a problem. I'm well familiar with the cadence. :)

Blackburn29:
@rbergen The PR has been updated. This _should_ meet the requirements for what you were asking for.

include/globals.h
```diff
@@ -1171,7 +1171,15 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 #endif
 
 #ifndef ENABLE_WEBSERVER
-#define ENABLE_WEBSERVER        0   // Chip provides a web server with controls to adjust effects
+    #define ENABLE_WEBSERVER        0   // Chip provides a web server with controls to adjust effects
+
```
rbergen, line 1175: I suggest we just leave this line alone and decrease the diff size a bit. We're not a big fan of whitespace-only changes, and this one breaks the "same indentation #ifndef/#define/#endif" pattern as well.

platformio.ini
```diff
@@ -30,7 +30,8 @@ extra_configs =
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     =
+#upload_protocol = 
+#upload_port     =
```
rbergen, line 34: Please revert these changes as well.

platformio.ini
```diff
@@ -245,11 +245,15 @@ lib_extra_dirs  = ${PROJECT_DIR}/lib
...
                   post:tools/merge_image.py
+
 board_build.embed_files = site/dist/index.html.gz
                           site/dist/index.js.gz
                           site/dist/favicon.ico.gz
+
 board_build.embed_txtfiles = config/timezones.json
 
+
+
```
rbergen, line 256: Same here for the extra white lines.

src/webserver.cpp
```diff
@@ -134,19 +134,17 @@ void CWebServer::begin()
...
     // Instance handler requests
-
     _server.on("/statistics",            HTTP_GET,  [this](AsyncWebServerRequest* pRequest) { this->GetStatistics(pRequest); });
     _server.on("/getStatistics",         HTTP_GET,  [this](AsyncWebServerRequest* pRequest) { this->GetStatistics(pRequest); });
 
     // Static handler requests
-
     _server.on("/effects",               HTTP_GET,  GetEffectListText);
     _server.on("/getEffectList",         HTTP_GET,  GetEffectListText);
     _server.on("/nextEffect",            HTTP_POST, NextEffect);
```
rbergen, line 150: And again.

rbergen (commented):
@Blackburn29 Thanks for the update, but I do still have a few comments. They all boil down to one thing: please keep your changes limited to actual code changes, and leave unrelated things (like white spacing in different code areas) as they are.

rbergen:
I've taken the liberty of applying my feedback myself. If CI completes successfully I will be happy to merge.

PR #601, opened 2024-01-27
Comment fix, or the math is wrong ?

fanno:
## Description
Just a comment change, tho it could also be the math is wrong and the comment is correct.

## Contributing requirements

This is my first pull request here and a long time since i have made one. so don't know if i am doing it right.

should i do it via a separate branch on my fork ? or master fine ? its been soooo long not sure if it matters.

-Thanks

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -264,7 +264,7 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
     #define NUM_CHANNELS            1
     #define ENABLE_AUDIO            0
 
-    #define POWER_LIMIT_MW       12 * 10 * 1000   // 10 amp supply at 5 volts assumed
+    #define POWER_LIMIT_MW       12 * 10 * 1000   // 12 volts, 10 amp supply
 
     // Once you have a working project, selectively enable various additional features by setting
     // them to 1 in the list below.  This DEMO config assumes no audio (mic), or screen, etc.
@@ -332,7 +332,7 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 
     #define DEFAULT_EFFECT_INTERVAL     (60*60*24*5)
 
-    #define POWER_LIMIT_MW       12 * 10 * 1000   // 10 amp supply at 5 volts assumed
+    #define POWER_LIMIT_MW       12 * 10 * 1000   // 12 volts, 10 amp supply
 
     #if M5STICKC || M5STICKCPLUS || M5STACKCORE2
         #define LED_PIN0 32


rbergen:
I'll start with this question:

> should I do it via a separate branch on my fork ? or master fine ? its been soooo long not sure if it matters.

In a professional context there's usually a convention for this, but in the context of this project we leave it up to the contributor to select their head branch (that's the one containing the changes they want to push). Using the default branch in your fork does mean that you can't make any unrelated changes to that before this PR is merged or closed, which is why I personally always create a dedicated topic branch for a PR I intend to open. But again, it's up to you.

About the change you're proposing: I'm not comfortable with updating the comment, because it might suggest people to hook up 12V power supplies to kits that (most) often really need 5V. And as resilient as ESP32s tend to be to "software-level abuse", overvolting them that much will usually fry them.

I'll check with Dave if there's a reason the maths seem to based on 12V, while the comment suggests 5V is the correct value.

rbergen:
@fanno @robertlipe I sent Dave a message to ask the question. I'll see what he comes back with, and we'll take it from there.

rbergen:
> I dunno. This is a copy-paste mess.

That's basically what we're looking at. Some bits in the universe would be in better places if we fix this, which is why I'm making an effort to figure out where this 12 is coming from. More news will follow when I have it.

fanno:
I my self is working on a 24 V led strip project

I think you are right i should have made a separate branch for this..

This is not directly related to this BUT so so on same "toppic"

In global.h we have this as a default.
```
#ifndef POWER_LIMIT_MW
    #define POWER_LIMIT_MW 500*5                // Define for your power supply, default is a low 2500mA for USB
#endif

```

I would personally have made it "different" in stead of having to calculate the the power limit i would have something like this (giving the flexibility of being able to "overwrite it")

something like this (not tested yet at all)
having already NUM_LEDS

with 3 new variables WATS_PER_LED / LED_VOLTAGE / AMPS_PER_LED (or some other naming)

```
#ifndef POWER_LIMIT_MW
    #ifdef NUM_LEDS
        #ifdef WATS_PER_LED
            #define POWER_LIMIT_MW NUM_LEDS*WATS_PER_LED*1000
        else
            #ifdef LED_VOLTAGE
                #ifdef AMPS_PER_LED
                    #define POWER_LIMIT_MW NUM_LEDS*LED_VOLTAGE*AMPS_PER_LED*1000
                #endif
            #endif
        #endif
    #endif
#endif

#ifndef POWER_LIMIT_MW
    #define POWER_LIMIT_MW 500*5                // Define for your power supply, default is a low 2500mA for USB
#endif
```

My reason for this you are sometimes/often given info from datasheet eg "max wats/meter" (strip) and sometimes also watts per led.

So from that if length change and so on, you don't need to remember to "recalculate" the POWER_LIMIT_MW

Most often a "user" would chose the same type led strip but the length for each strip would chance "NUM_LEDS" so if you define WATS_PER_LED or (LED_VOLTAGE and AMPS_PER_LED) it POWER_LIMIT_MW would not need corrected if working with say  strips with various lengths

Not sure if the naming scheme of the variables here is correct/professionally i mostly program robots and plc's for a living.

I am aware this consent perhaps mostly works with strips and matrix ect may be different, but it could be "best of both worlds"

-Morten Hundevad

fanno:
Hello

> as to default is a low 2500mA for USB
Thats also in the source code i did not even see that one.

if NUM_LEDS is not the total number of LED you need a "TOTAL_NUM_LEDS"

POWER_LIMIT_MW should also be the total.. 

tho you could have 2 powersupply aswell not cuppoled. so you can have a 2 am supply on 1 strip and 8 amps on another.

Tho i think in reality normally one supply.

of one wanted this to be "correct"

POWER_LIMIT_MW for each strip/device

have them as array one per device would properly be more correct. (and maybe more flexible method), tho i do not know how this would work with defines.
NUM_LEDS[xx]
POWER_LIMIT_MW[xx]
....[xx]

I am willing to put in the work to make this change, just need pointers on how it should be done and i will make sure its correctly implemented and make a pull request.

is backwards capability something we should also think about?

Highlights how i see it:
Power limit "should" be PER "device"
I would like to see it calculated "automatic" but still have ability to overwrite it, where calculation via led/w led/V/A don't make sense..

Just so i don't come off a "snob" as i am a new comer her, i do recolonize this and don't expect that i just come and "change" everything.
This just looks like a "slight" over side/optimization that CAN be change but do not NEED to be.

-Morten Hundevad

P.S.
Again i just want to stress i am not trying to step on anyone's toes.

fanno:
I dig a little further.

Looks like Fast led do NOT support for separate power supply, or device. (of cause i could be wrong)

```
void set_max_power_in_milliwatts( uint32_t powerInmW)
{
    FastLED.setMaxPowerInMilliWatts(powerInmW);
}
```

there is also a function that takes amps/voltage directly it looks like
```
   // limit my draw to 1A at 5v of power draw
   FastLED.setMaxPowerInVoltsAndMilliamps(5,1000); 
```

calculate_max_brightness_for_power_mW is used as m_pPowerFunc
```	inline void setMaxPowerInMilliWatts(uint32_t milliwatts) { m_pPowerFunc = &calculate_max_brightness_for_power_mW; m_nPowerData = milliwatts; }```


As you can see unless i miss something, it IS not taking in to account power usage per channel (this IS a total over side in my opinion (in FastLED))

by this i mean it don't "protect" each "device" and don't support separate power supplies

One can argue that separate power supply is "uncommon", protection of the device should be done with a fuse.

What this is normally use for i assume is when you have led turning on/off and you have a power supply of say 5 amps and total draw in strip is sat 6 amps at full power all led it will scale IF all are on max but not if say 50 % of the led's are on.

But that would depend on the m_pPowerFunc
```
void CFastLED::show(uint8_t scale) {
	// guard against showing too rapidly
	while(m_nMinMicros && ((micros()-lastshow) < m_nMinMicros));
	lastshow = micros();

	// If we have a function for computing power, use it!
	if(m_pPowerFunc) {
		scale = (*m_pPowerFunc)(scale, m_nPowerData);
	}

	CLEDController *pCur = CLEDController::head();
	while(pCur) {
		uint8_t d = pCur->getDither();
		if(m_nFPS < 100) { pCur->setDither(0); }
		pCur->showLeds(scale);
		pCur->setDither(d);
		pCur = pCur->next();
	}
	countFPS();
}
```

In FASTled m_pPowerFunc (or a separate function that support reference to pCur ) should have been called inside each loop of pCur.

But i do not know if this is something we can "fix" i don't know cpp well enough to know if we can overwrite the (void CFastLED::show(uint8_t scale)) from within NDS to add this support. (don't know how hard it would be to have FastLED to make changes) for this support, if that is the only way.

This analysis assume that pCur is a "channel", so pehaps i am all wrong.

-Morten Hundevad

rbergen:
Ok, so an update on this one: the conclusion Dave and I came to is that we should drop the _requirement_ for defining `POWER_LIMIT_MW` altogether, and with that also actually remove all "project default" definitions of that symbol. Which is quite easy to do, except there is one place in `LEDStripGFX::PostProcessFrame()` (defined in ledstripgfx.cpp) that uses it unconditionally - but it is quite obvious how to fix that as well.

So, I'll open a PR shortly that makes those changes exactly. When I have and once we can agree that is the way to go, I'll merge that approach instead of this one.

fanno:
ok no problem

Den ons. 7. feb. 2024 kl. 20.08 skrev Rutger van Bergen <
***@***.***>:

rbergen:
Closing this in favor of #601. @fanno Thank you for bringing the greater point to our attention!

PR #642, opened 2024-07-08
Add M5 Unified support (M5 Stick C Plus 2, M5 Stack Core 2)

### LINKED ISSUE #582: Update from M5Core2 to M5Unified
It looks like all of the M5 support has been unified into a single library, and the old individual stuff for the M5 Stick / Stack/ Plus/ Core2/ etc will go away.

Someone should move us to the unified lib!

### LINKED ISSUE #618: New M5StickC Plus2 doesn't work
I tried to upload project as old M5stickc plus and did not work,
don't reach wifi config. someone has experience ?

davepl:
## Description
Moves all M5-based projects to the unified M5 stack

Fixes #582. Fixes #618.


davepl:
Can’t believe I didn’t know how to spell Cylon :-). Whoops!  I’ve only heard it on TV, not read it in books…

Std::size should be used in place of ARRAYSIZE

On the PSRAM, not sure what you’re advocating for.  If you run your own MENUCONFIG for the ESP kit, you can set it to even do static data allocations out of PSRAM, but I was under the impression that since we use the Arduino version, we’re stuffed with the one and only MENUCONFIG setup they offer, so we have to do it ourselves.

No?

_ Dave

include/effects/strip/faneffects.h
```diff
@@ -1067,8 +1067,8 @@ class FireFanEffect : public LEDStripEffect
     {
       for (int i = 0; i < CellCount(); i++)
       {
-        int coolingAmount = random(0, Cooling);
-        abHeat[i] = ::max(0.0, abHeat[i] - coolingAmount * (2.0 - g_Analyzer._VURatio));
+          float coolingAmount = random_range(0.0f, 2.0f);
```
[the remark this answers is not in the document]
rbergen: I think the "problem" with the indentation is that the file uses 2-space indentation, where the project heavily leans towards 4 spaces. As the inconsistency is applied in a consistent manner within the block, I'm now going to conclude I've just used my yearly quota of words-spent-on-spaces.

include/effects/strip/misceffects.h
```diff
@@ -503,6 +503,97 @@ class TwinkleEffect : public LEDStripEffect
...
+    SilonEffect() : LEDStripEffect(EFFECT_MATRIX_SILON, "SilonEffect")
+    {
+    }
+
+    SilonEffect(const JsonObjectConst& jsonObject)
+      : LEDStripEffect(jsonObject)
+    {
+    }
+
+    int _offset = 0;
```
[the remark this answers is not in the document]
rbergen: Would it? I think an unsigned would underflow to the maximum value of the unsigned type when 1 gets subtracted from 0, which would make the "less than 0" condition _harder_ to identify.

include/globals.h
```diff
@@ -903,14 +985,70 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
...
     #if !(ELECROW)
         #define TOGGLE_BUTTON_1         37
         #define TOGGLE_BUTTON_2         39
     #endif
 
+#elif HELMET
+
+    // This project is set up as a 48x16 matrix of 16x16 WS2812B panels such as: https://amzn.to/3ABs5DK
+    // It uses an M5StickCPlus which has a microphone and LCD built in:  https://amzn.to/3CrvCFh
+    // It displays a spectrum analyzer and music visualizer
```
rbergen, line 1000: I think this comment concerns the Spectrum project?

include/globals.h
```diff
@@ -903,14 +985,70 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
...
+    #define ENABLE_AUDIOSERIAL              0   // Report peaks at 2400baud on serial port for PETRock consumption
+    #define ENABLE_WIFI                     1   // Connect to WiFi
+    #define WAIT_FOR_WIFI                   0   // Hold in setup until we have WiFi - for strips without effects
+    #define TIME_BEFORE_LOCAL               2   // How many seconds before the lamp times out and shows local content
+    #define ENABLE_WEBSERVER                1   // Turn on the internal webserver
+    #define ENABLE_NTP                      1   // Set the clock from the web
+    #define ENABLE_OTA                      0   // Accept over the air flash updates
+    #define ENABLE_REMOTE                   1   // IR Remote Control
+    #define ENABLE_AUDIO                    1   // Listen for audio from the microphone and process it
+    #define COLORDATA_SERVER_ENABLED        0
```
rbergen, line 1017: I think we can remove this line, considering the `#if USE_PSRAM`/`#else` block that follows which also contains it for both cases.

platformio.ini
```diff
@@ -13,7 +13,7 @@
 ; default_envs to the project you wish to build.  There are project configs for various different
 ; NightDriver projects ranging from WiFi LED Strip to the simple lantern candle:
 ;
-; demo          Simple WS2812B strip controller, runs a rainbow on PIN 5
+; demo          Simple WS2812B strip contro ller, runs a rainbow on PIN 5
```
rbergen, line 16: I think we can continue spelling "controller" as one word ;)

platformio.ini
```diff
@@ -496,6 +506,23 @@ build_flags     = -DSPECTRUM=1
                   ${dev_m5stick-c-plus.build_flags}
 board_build.partitions = config/partitions_custom_noota.csv
 
+[env:spectrum2]
+extends         = dev_m5stick-c-plus2
+build_flags     = -DSPECTRUM=1
+                  -DSHOW_VU_METER=1
+                  -DM5STICKCPLUS2=1
```
rbergen, line 513: This line can be removed. It's in included in the `dev_m5stick-c-plus2` section build flags.

samples/videoserver/pcstats.py
```diff
@@ -15,7 +15,7 @@
 
 # Dependencies
 
-from PIL import Image, ImageDraw, ImageFont     # python3 -m pip install pillow
+from PIL import Image, ImageDraw, ImageFont     # python3` -m pip install pillow`
```
rbergen, line 18: I'm guessing the addition of the back ticks happened by accident? I don't think they will work in the copy/pasted command.

include/effects/strip/misceffects.h
```diff
@@ -503,6 +503,97 @@ class TwinkleEffect : public LEDStripEffect
...
+
+// PDPGridEffect
+// 
+// A Display for the front of the PDP-11/34
+
+class PDPGridEffect : public LEDStripEffect
+{
+  public:
+
+    PDPGridEffect() : LEDStripEffect(EFFECT_MATRIX_PDPGRID, "PDPGRIDEffect")
```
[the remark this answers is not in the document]
rbergen: I agree.

rbergen (commented):
Note: some of my comments are responses to @robertlipe's comments. I've been told by others that GitHub can make it difficult to follow these threads when consumed via email. So, it may be best (if not necessary?) to view the review on GitHub.

So I do have a few comments, but no deal-breakers as far as I'm concerned. I'll push an update to this PR to add an M5StickC Plus2 project to the web installer, and address some of my own (textual) comments in that. 

@davepl I'll leave up to you which of Robert's and my remaining comments you want to process before you merge.

include/effects/strip/misceffects.h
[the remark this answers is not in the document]
rbergen: No worries. You definitely made me look at that bit of code with more attention than I otherwise would have. :)

prschguy1:
Hello Dave, thanks for your work on getting us to the new library. I had tried doing that, but had failed to get it working. That being said, it seems your spectrum display is processing artifacts, rather than actual audio. I have included a couple of photos. 1 in the fork preview you provided, and 1 loaded from 11/18/2023. Both are in glass containers, under a slight vacuum. As you can see, m5 plus, plus 2, and stack 2 are processing audio that just isn't there. I don't think this is something that can be addressed with noise floor settings. Hoping you can get this figured out. Additionally, I was using the original soundanalyzer.h files for m5 stack for other esp32 boards. Perhaps that code could be kept. and re-named as misc
20240713_045324_resized
20240713_051315_resized
-digital mic or other. Just my 2 cents.

Craig

davepl:
Yup, I’m aware of it, and unfortunately do not have a good solution.  Doesn’t mean there isn’t one, but someone with a bigger brain than mine might need to look at soundanalyzer.h

Long story short, with the M5 Unified library, the levels never go down, they seem to auto-gain.  It’s the same MIC…

I’d rather not have multiple M5 impls, but if someone can get the old i2s code to work on all the M5s, that’d likely solve it and would be a good solution.

Mostly I’d like to avoid the copout solution of “it works better on old M5s” and I guess I’m using that as a forcing function to get it fixed!

- Dave

rbergen:
Merging after out-of-band alignment with @davepl.

PR #643, opened 2024-07-12
Parallelize Web Installer build

### LINKED ISSUE #644: Build instructions should document requirements
I'm making this a bugreport because it'll probably take a few of us (esp. those that are multi-OS and have ability to make blank OS images for testing...) to produce a list of requirements.

## Bug report
New developers, following our instructions, can run into errors like this:

file:///NightDriverStrip-yulc/site/node_modules/vite/bin/vite.js:7
    await import('source-map-support').then((r) => r.default.install())
    ^^^^^

SyntaxError: Unexpected reserved word
    at Loader.moduleStrategy (internal/modules/esm/translators.js:133:18)
    at async link (internal/modules/esm/module_job.js:42:21)

**Problem**

**Steps**

1.  Follow instructions.
2.  Observe build failure.
3.  Become sad. Start drinking.

**Example**
We should probably have a section of doc that lists requirements for installing
* Platformio
* (Does ESP-IDF have to be separately installed?)
* Vite
(any other npm kind of stuff?)

**Notes**
For MacOS, I know 'brew install vite' is the (a) solution to above.
Ideally, we'd have copy-pastable:
Linux: yum install platformio vite 
Windows: whatever
MacOS: recommend the excellent homebrew. $ brew install vite  Site for platformio, etc.

rbergen:
## Description

This parallelizes the Web Installer build when executed by/through GitHub Actions, thereby significantly shortening the overall build time for the installer.

While reviewing and testing #642 I found that the Web Installer build alone crossed the 1 hour duration boundary, which marks my personal psychological boundary for "ok, let's see if we can make this go quicker".

The bake_installer.py script can still be used for (sequential) local builds; it's been refactored to use the segmented Python scripts that are also used by the GitHub Actions workflow.

Also, the Build tools section in the README got updated to better clarify the requirements concerning NodeJS and NPM. This fixes #644.

## Contributing requirements


rbergen:
Small heads-up: if this PR is not reviewed by tomorrow I will apply the "they who are silent agree" principle to it then. The justification is that I've tested it on my fork on two states of the codebase (one with #642 in and another with that PR out), and the fact that if this breaks anything it will just mean that no new Web Installer will be published on the next release of NightDriverStrip - in which case I will fix that then.

rbergen:
Merging based on the principle mentioned in an earlier comment.

PR #648, opened 2024-08-21
Numerous minor changes for correctness.

robertlipe:
- **Zillions of tiny edits to remove a warning or improve correctness.**
- **Prep for push**
- **Fix typo in a symbol that should be deleted...**

## Contributing requirements

### DIFF
--- a/include/colordata.h
+++ b/include/colordata.h
@@ -23,7 +23,7 @@
 //
 // Description:
 //
-//    Palettes and other color table defnitions.  CPP daa in colordata.cpp
+//    Palettes and other color table definitions.  CPP daa in colordata.cpp
 //
 // History:     May-11-2021         Davepl      Commented
 //
@@ -34,7 +34,6 @@
 extern const CRGBPalette16 vuPaletteGreen;
 extern const CRGBPalette16 vuPaletteBlue;
 extern const CRGBPalette16 spectrumBasicColors;
-extern const CRGBPalette16 bluesky_pal;
 extern const TProgmemRGBGradientPalette_byte vu_gpGreen[];
 extern const TProgmemRGBPalette16 BlueHeatColors_p;
 extern const TProgmemRGBPalette16 HeatColors2_p;
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -30,7 +30,6 @@
 
 #pragma once
 
-// #include <memory>
 #include <vector>
 #include <tuple>
 #include "jsonserializer.h"
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -30,10 +30,9 @@
 
 #pragma once
 
-#include <memory>
+// #include <memory>
 #include <vector>
 #include <tuple>
-#include <algorithm>
 #include "jsonserializer.h"
 #include "types.h"
 
@@ -146,7 +145,7 @@ class DeviceConfig : public IJSONSerializable
 
     static constexpr int _jsonSize = 1024;
 
-    void SaveToJSON();
+    void SaveToJSON() const;
 
     template <typename T>
     void SetAndSave(T& target, const T& source)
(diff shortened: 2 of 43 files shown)


rbergen:
Hi @robertlipe, I was wondering how/to what extent you've tested these changes?

rbergen:
For me the main thing is that I **hate** breaking functionally working code to implement improvements that may not even be contested in themselves, but are "only" improvements on a non-functional level. At a 99.5% chance of over-explaining, the reason is that in my mind the only real reason for code to be there is to do the stuff it does. There are almost always multiple ways to do stuff and some are better than others, but if the code that's there works and the best-case scenario from a functional perspective is that it does exactly what it does now, then I tend to stick to "if it ain't (functionally) broken, don't try to fix it."

In a "binary world" that would put me on the track of "thanks, I'll take the comment changes, but let's just leave the code alone until we have another reason to change it". However, I do understand the value of maintainability, consistency, testability (yes, I read ahead), etc. so I'm not going to make that the conclusion just like that.

It does mean that I think we should indeed know we've tested all code paths we're touching at least once, and can't really have any known blind spots there. As responsible guardians of our users' functionality, if you will. And I know, most if not all of these changes should "just work", but you and I have both been doing this long enough to have collected a beautiful batch of campfire stories where it should have just worked... and didn't.

So, if testing blind spots do exist (like the colorserver example you mention), then let's cover those before we proceed. And colorserver is a great example, because I also don't use that - which is why I don't think I've ever touched its code in my three years on this project. If we can "casually" make the way we cover them automated and repeatable then one could successfully argue with me that we actually _have_ a reason to touch the code.

Concerning the contestability of the changes in this PR at a non-functional level: in my view many of them are clearly improvements because the English language dictates that words are spelled in a certain way, because `.empty()` is less error-prone than `.size() == 0`, adding `const` more clearly documents in code what is (not) supposed to happen in a function or with a variable/parameter, etc. In some other cases I think it's more of a coding style/standard thing. Is a raw string literal really clearer than two escaped double quotes, when that literal is still part of a string concatenation? I think one could argue either way (and in all honesty: I think not, but that's also just an opinion).

Which kind of brings me to the question about what makes changes like these more digestible for me, in decreasing order of importance:

1., 2. and 3. Make sure you've tested the new version of the old working code and have seen it working
4. Make it "obvious" that the change is an improvement, not the likely subject of a conversation that boils down to "but I really think blue is better than red!" (I know there is a bit of aptness in US politics with that statement, and it's actually a case in point in a way: I do have a strong opinion on the matter, but as I have no vote and therefore no influence, I think it's best for me to keep my mouth shut about _that_ topic...)

Also, non-functional changes that are part of another "objective" improvement are always easier for me to be comfortable with, if only because that tends to cover the first criterion.

I don't know if I've covered all points you asked my input on with this; feel free to point out any that I missed.

Cheers,
Rutger

rbergen:
@robertlipe May I suggest we reopen this? I really would like to incorporate these changes to the extent we (that includes me personally) can test them, and I that's actually the majority of them.

rbergen:
I'm reopening this myself - I noticed the PR was created from a branch in my own fork, so I feel I'm not out of line by doing so.

Compared to the original submission by @robertlipe:

- The changes to socketserver.h have been reverted, as I can't properly test them.
- I've reverted a couple of changes I felt didn't constitute an improvement to the codebase.
- I've included an update to platformio.ini, to fix builds for M5 devices - they started breaking because the M5Unified library now depends on M5GFX, which wasn't pulled in when M5Unified was pulled in from GitHub instead of as a PlatformIO package.

All other changes included in the original PR have been tested by myself using the Mesmerizer and Spectrum (for M5StickC Plus) configurations.

rbergen:
As I prepare to merge this, I'm hereby documenting for posterity that I'll restore the originally proposed changes to socketserver.h in this PR's head branch after this PR has been merged. 

That way they can still be applied after someone has been able to confirm on actual hardware that the changes don't break anything.

PR #658, opened 2024-10-10
nuke reg

robertlipe:
- **Remove outdated kludge for FastLED's use of register.**

## Description
FastLED used to use C style hints of 'register' that were deprecated
long ago and the keyword was removed in C++17. There was a kludge
that just defined register away.
FastLED was been updated (long ago) so this is no longer necessary.
We can return this token to valid namespace.

This has been in my tree for many months and has survived buddybuilds
and a variety of FastLED versions.

## Contributing requirements

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -35,7 +35,6 @@ monitor_port    =
 build_type      = release
 upload_speed    = 1500000
 build_flags     = -std=gnu++2a
-                  -Dregister=                       ; Sinister: redefine 'register' so FastLED can use that keyword under C++17 and later C++ versions
                   -g3
                   -Ofast
                   -ffunction-sections


rbergen:
Thanks!

If this "sinister" hack is no longer necessary then I'm all for removing it. So, that change I support.

However, there are two issues with the PR:

- It includes a lot more than the undefine of `register`. To me it looks like it includes at least also the changes proposed in #648, which is still open due to an unresolved discussion about testing.
- Probably (partly) due to the previous point, it now causes merge conflicts in 4 files.

From a pragmatic perspective, the easiest way forward for the `register` change may be to create a branch from current "main", apply the change to it and base the PR on that.

rbergen:
As discussed out-of-band, I don't know any developer who used git more than a week who does not have some horror story experience with it. Which means things like these are part of the dynamics, which means apologies are not necessary.

And in relation to that, just in case there's any confusion: my previous comment was an observation, not an accusation. Shit happens, pardon my French.

Issue #294, opened 2023-05-31
Feature: expose additional effect settings


rbergen:
Recently, the ability was added for effects to expose settings that can be changed via the API. A few basic settings have been exposed by `LEDStripEffect`; the `PatternSubscribers` effect adds two more for itself. 

Obviously, there are more effects for which it makes sense to expose properties as settings that can be changed via the device API and (later) the device web UI. 

When working on this:

- The [`REST_API.md`](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/REST_API.md) file contains information about the effect settings API endpoints.
- The aforementioned [`PatternSubscribers` effect](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/include/effects/matrix/PatternSubscribers.h) provides an example of an effect extending the basic settings exposed by `LEDStripEffect`.
- [`LEDStripEffect`](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/include/ledstripeffect.h) includes a number of helper functions for changing effect settings of "known types". Known types are themselves defined by the [`SettingSpec` struct in `types.h`](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/023ae7f148ac3509ddaa4f4ffe1e92d149e18134/include/types.h#L50).

Louis-Riel:
I'm POCing something to save UI effect settings in the browser storage facilities. In this POC, I'm setting an effect image URL option. This is what the effect image feature looks like:
![image](https://github.com/PlummersSoftwareLLC/NightDriverStrip/assets/53452394/a10542c6-5cf5-4dba-b13a-252effabd0e3)
The little gear opens a config dialig:
![image](https://github.com/PlummersSoftwareLLC/NightDriverStrip/assets/53452394/6f067be7-e4fa-47a0-a1d8-5ceca0292eb8)
The favicon.ico is the default, and the one in the screenshot is the overridden value:
![image](https://github.com/PlummersSoftwareLLC/NightDriverStrip/assets/53452394/d9016586-60f5-4b87-a8a0-5acd95ea6902)

rbergen:
@Louis-Riel Would you envisage using this UX for other chip-stored effect settings as well? Because that's what this issue is primarily about.

Concerning the image per effect: I can definitely see that being a chip-stored effect setting, provided we indeed make it a URL (not an actual image upload that is stored to SPIFFS, for instance.)

[...]

PR #660, opened 2024-10-28
Test of RGBW Branch

davepl:
## Description
Adds RGBW support, tweaks SpiralLamp, adds PlateCover

## Contributing requirements


config/partitions_custom.csv
```diff
@@ -1,12 +1,8 @@
...
+# Name,   Type, SubType,     Offset,      Size, Flags
 nvs,        data,   nvs,     0x00b000,    0x002000,
 otadata,    data,   ota,     0x00d000,    0x002000,
-app0,       app,    ota_0,   0x010000,    0x190000,
-app1,       app,    ota_1,   0x1A0000,    0x190000,
-storage,    data,   spiffs,  0x330000,    0x0D0000 
\ No newline at end of file
+app0,       app,    ota_0,   0x010000,    0x1A0000,
+app1,       app,    ota_1,   0x1B0000,    0x1A0000,
+storage,    data,   spiffs,  0x350000,    0x0B0000
```
davepl, line 8: You'll note I stole another x20000 bytes from storage to make room for code, but it seems to still fit everything in!

include/deviceconfig.h
```diff
@@ -94,7 +94,9 @@
 
 #define DEVICE_CONFIG_FILE          "/device.cfg"
 #define NTP_SERVER_DEFAULT          "0.pool.ntp.org"
-#define BRIGHTNESS_MIN              uint8_t(10)
+#ifndef BRIGHTNESS_MIN
+    #define BRIGHTNESS_MIN          uint8_t(10)
+#endif
 #define BRIGHTNESS_MAX              uint8_t(255)
```
davepl, line 100: Allows non-matrix effects to set the brightness to 0 if the need to, like lamp for example

include/effectmanager.h
```diff
@@ -153,9 +153,19 @@ class  EffectManager : public IJSONSerializable
...
 
-    std::shared_ptr<GFXBase> GetBaseGraphics()
+    // SetTempEffect - Sets a temporary effect to be played until remote changes it.
+    //                 The effect must have already had its Init() function called.
+
+    void SetTempEffect(std::shared_ptr<LEDStripEffect> effect)
+    {
+        _tempEffect = effect;
+    }
+
```
davepl, line 163: I could swear this functionality existed for the atomic lamp already, but could not find a way to actually set the temp effect, so added this method.  Let me know if I missed something that would have obviated the need for it!

include/effectmanager.h
```diff
@@ -153,9 +153,19 @@ class  EffectManager : public IJSONSerializable
...
+    // GetBaseGraphics - Returns the vector of GFXBase objects that the effects use to draw
+    
+    std::vector<std::shared_ptr<GFXBase>> & GetBaseGraphics()
     {
-        return _gfx[0];
+        return _gfx;
     }
```
davepl, line 169: I now return the entire array of graphics rather than just the first so that the VU meter effect can draw onto both spokes of the spiral lamp, which has two channels

include/globals.h
```diff
@@ -1736,9 +1775,8 @@ inline bool SetSocketBlockingEnabled(int fd, bool blocking)
 
 inline String formatSize(size_t size, size_t threshold = 1000)
 {
-    // If the size is less than the threshold, we don't need to worry about precision because
-    // we'll be showing whole units
-    const int precision = size >= threshold ? 2 : 0;
+    // If the size is above the threshold, we want a precision of 2 to show more accurate value
+    const int precision = size < threshold ? 0 : 2;
```
davepl, line 1779: Check me here, I could be crazy, but I think this was backwards, so flipping it

src/effects.cpp
```diff
@@ -276,7 +276,7 @@ void LoadEffectFactories()
         ADD_EFFECT(EFFECT_MATRIX_SMPICASSO3IN1,     PatternSMPicasso3in1, "Lines", 38);
         ADD_EFFECT(EFFECT_MATRIX_SMPICASSO3IN1,     PatternSMPicasso3in1, "Circles", 73);
         ADD_EFFECT(EFFECT_MATRIX_SMAMBERRAIN,       PatternSMAmberRain);
-        ADD_EFFECT(EFFECT_MATRIX_SMSTROBE_DIFFUSION,PatternSMStrobeDiffusion);
+ //     ADD_EFFECT(EFFECT_MATRIX_SMSTROBE_DIFFUSION,PatternSMStrobeDiffusion);
         ADD_EFFECT(EFFECT_MATRIX_SMRAINBOW_TUNNEL,  PatternSMRainbowTunnel);
```
davepl, line 280: I'm sure I had a good reason for turning off the Diffusion effect, but I think I just didn't like it.  If you or anyone else is partial to it, let me know.

include/effects/matrix/spectrumeffects.h
```diff
@@ -120,14 +120,14 @@ class VUMeter
...
     {
         if (g_Analyzer.MicMode() == PeakData::PCREMOTE)
             pPalette = &vuPaletteBlue;
 
-        int xHalf = pGFXChannel->width()/2;
-        pGFXChannel->setPixel(xHalf-i-1, yVU, ColorFromPalette(pPalette ? *pPalette : vu_gpGreen,  i*(256/xHalf)).fadeToBlackBy(fadeBy));
-        pGFXChannel->setPixel(xHalf+i,   yVU, ColorFromPalette(pPalette ? *pPalette : vu_gpGreen, i*(256/xHalf)).fadeToBlackBy(fadeBy));
+        int xHalf = GFX[0]->width()/2;
+        GFX[0]->setPixel(xHalf-i-1, yVU, ColorFromPalette(pPalette ? *pPalette : vu_gpGreen,  i*(256/xHalf)).fadeToBlackBy(fadeBy));
+        GFX[0]->setPixel(xHalf+i,   yVU, ColorFromPalette(pPalette ? *pPalette : vu_gpGreen, i*(256/xHalf)).fadeToBlackBy(fadeBy));
```
davepl, line 130: The existing VU meter effect draws a mirrored VU on the matrix, so assumes a single channel, and doesn't draw what I want for the lamp...  so I made a VUMeterVertical variation of it that does what I want.  Since it can no longer assume it uses just the matrix channel, I had to change GetBaseGraphics as noted.

include/soundanalyzer.h
```diff
@@ -639,16 +652,18 @@ class SoundAnalyzer : public AudioVariables
...
             {
-                _peak1Decay[i] = _Peaks[i];
+                _peak1Decay[i] = std::min(_Peaks[i], _peak1Decay[i] + maxIncrease);
                 _lastPeak1Time[i] = millis();
             }
             if (_Peaks[i] > _peak2Decay[i])
             {
-                _peak2Decay[i] = _Peaks[i];
+                _peak2Decay[i] = std::min(_Peaks[i], _peak2Decay[i] + maxIncrease);
             }
```
davepl, line 667: In a later checkin I fix this to use the individual decay rates 1 and 2 rather than basing them both off 1, but it doesn't impact the display.

include/effectmanager.h
rbergen: As far as I can remember, the `_tempEffect` could only be set through the constructor the whole time. But there is a theoretical chance that the possibility did exist a long time ago, and I removed it because the function wasn't used anywhere - I have done some dead code clean-up over time.

include/globals.h
rbergen: I think I'm neutral on the matter.

I am bound to remember this exchange if you change your mind about this in a couple of years, though. ;)

include/ledstripeffect.h
```diff
@@ -446,26 +449,28 @@ class LEDStripEffect : public IJSONSerializable
...
+            if (bMirrored)
+            {
+                setPixelOnAllChannels(iStart + i, rgb);
+                setPixelOnAllChannels(iStart + numToFill - i - 1, rgb);
+            }
+            else
+            {
+                setPixelOnAllChannels(iStart + i, rgb);
+            }
             for (int q = 1; q < everyNth; q++)
```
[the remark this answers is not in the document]
rbergen: To the extent I can trace things back (and trust GitHub's blame history), this code was originally submitted by yourself three years ago, probably when this repo was created. 

With which I just want to say that you're probably most qualified to have an opinion on this/decide changes are necessary.

include/soundanalyzer.h
```diff
@@ -502,6 +502,19 @@ class SoundAnalyzer : public AudioVariables
...
 
+    // These functions allow access to the last-acquired sample buffer and its size so that 
+    // effects can draw the waveform or do other things with the raw audio data
+    
+    const int16_t * GetSampleBuffer() const
+    {
+        return ptrSampleBuffer.get();
+    }
+
+    const size_t GetSampleBufferSize() const
```
rbergen, line 513: Can't this be a constexpr function?

include/soundanalyzer.h
rbergen: Ok.

src/effects.cpp
rbergen: To be honest, if it's just a matter of you not liking it then I'd like to suggest leaving it in and allowing the user to disable it with the web UI. That's what that feature is for.

If there is another good reason for disabling it, then that's fine of course. In which case I do (indeed) have a mild curiosity what the problem with it is.

src/effects.cpp
```diff
@@ -447,20 +447,29 @@ void LoadEffectFactories()
...
+        ADD_EFFECT(EFFECT_STRIP_FIRE_FAN, FireFanEffect, HeatColors_p, NUM_LEDS, 4, 5.0, 200, 8, 8, Sequential, true, true, true, 90);
+        ADD_EFFECT(EFFECT_STRIP_RAINBOW_FILL, RainbowFillEffect, 16, 3, true);
+        ADD_EFFECT(EFFECT_STRIP_METEOR, MeteorEffect, 2, 1, 15, .75, .75);
+        ADD_EFFECT(EFFECT_STRIP_COLOR_FILL, ColorFillEffect, "Off", CRGB::Black, 1);
+
     #elif SPIRALLAMP
 
-        #ifndef EFFECT_SET_VERSION
-            #define EFFECT_SET_VERSION  2   // Bump version if default set changes in a meaningful way
-        #endif
```
rbergen, line 454: Considering the pretty extended changes to this project's effect set I really think these `EFFECT_SET_VERSION` lines need to stay in, and that `EFFECT_SET_VERSION` for `SPIRALLAMP` in fact needs to be bumped to 3, here.

include/gfxbase.h
```diff
@@ -65,6 +65,7 @@
 #pragma once
 
 #include <stdexcept>
+#include <Arduino.h>
```
rbergen, line 68: With the 3.9.x FastLED compilation issues having been worked around, is this include still necessary?

rbergen (commented):
Looks like a nice project/effects upgrade!

I did make a few comments you may want to at least consider.

include/ledstripeffect.h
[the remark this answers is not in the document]
davepl: It makes sense now - it fills in the. pixels that the first loop did not!

src/effects.cpp
davepl: Ok, keeping it

include/gfxbase.h
davepl: Removed

src/effects.cpp
```diff
@@ -447,20 +447,37 @@ void LoadEffectFactories()
         ADD_EFFECT(EFFECT_STRIP_METEOR, MeteorEffect, 1, 1, 5, .15, .25);
         ADD_EFFECT(EFFECT_STRIP_METEOR, MeteorEffect); // Rainbow palette
 
+    #elif PLATECOVER
+
+        #ifndef EFFECT_SET_VERSION
+            #define EFFECT_SET_VERSION  3
+        #endif
```
rbergen, line 454: I'm changing this to 1 for PLATECOVER; the effect set version is project specific.

```suggestion
        #ifndef EFFECT_SET_VERSION
            #define EFFECT_SET_VERSION  1
        #endif    
```

rbergen:
I'm noting for the record that I've re-enabled the CI build for "heltecv3demo". The problem in PlatformIO that caused the project's build to fail was fixed in [platform-espressif32](https://github.com/platformio/platform-espressif32) version 6.9.0.

rbergen (approved):
Thanks for your changes. LGTM!

config/partitions_custom.csv
rbergen: We can rename the file to include 4M, but I think the fact of the matter is that we (still) treat 4M as the default. The 8M table does include the 8M in the name. 
Anyway, I don't think I'd merge a PR that implements only this name change, but wouldn't block it if it was part of a bigger change to the way we set up our partitions.

Concerning not using 16MB partition tables, I think the honest reason for that is that there currently is no scenario where we really need more than 8MB. So, why complicate matters by introducing and maintaining another partition table? (That may sound like a bit of a lazy argument, but I think it's valid in its core. :) )

I've made a fiery argument for not destroying user's data partitions in a willy-nilly way in the past, and from a consumer product perspective I stand by that. At the same time and at this moment, the consumer product that Mesmerizer could potentially be is not sold that way; it's a kit that requires the buyer to somewhat know what they're doing to put it together. It's also not being sold by the thousands, to put it very mildly. 

Mesmerizer is also really the only project with effects that rely on configuration a user could care about, and far fewer effects allow themselves to be configured than I personally would have liked (hence #294, opened a year and a half ago yesterday). Off the top of my head, we're really only talking about the weather, stocks and _possibly_ subscribers effects.
I therefore think that _at the moment_ we can still make "breaking" changes to the partition layout _used by Mesmerizer_, but I reserve the right to change my mind further down the line if one or more of the parameters for my current perspective change. 
The partition layout for other projects can probably remain in development for (much) longer in any scenario.

And yes, I can see the benefit of a crash partition. If you're willing to pick this one up, then I'm happy to review it. In the end, the main criterion is that all projects still build and flash, and that all data we want to put in the data partitions fits in there. For most projects, that is "just" the web UI files and the JSON config and current effect files. The GIFs are only flashed for Mesmerizer, at the moment.

rbergen: Wait, hang on. _All_ flashed files are embedded into the firmware image, so end up on the application partition. The SPIFFS data partition really only contains the JSON config files (and the Improv log file, if that's enabled - which it normally isn't). 

I think that probably makes the challenge of making room for a crash partition a little more manageable...

PR #668, opened 2024-11-20
Added heltec wifi lora v3 board

BigDurl:
## Description
Added new dev_heltec_wifi_lora_v3 device and helteclorav3demo environment.
Changed build flag for dev_heltc_wifi_v3 from -DWIFI_Kit_32=1 to -DWIFI_Kit_32_V3.
Both flags do the same thing but passing the later may help compatibility in the future.
Fixed line in screen.h to disable LoRa by default.

## Contributing requirements

### DIFF
--- a/include/screen.h
+++ b/include/screen.h
@@ -314,7 +314,7 @@ public:
         U8G2_SSD1306_128X64_NONAME_F_HW_I2C oled;
 
     public:
-        #if ARDUINO_HELTEC_WIFI_LoRa_32_V3
+        #if ARDUINO_HELTEC_WIFI_LORA_32_V3
             OLEDScreen(int w, int h) : Screen(w, h), oled(U8G2_R2, /*reset*/ 21, /*clk*/ 18, /*data*/ 17)
         #else
             OLEDScreen(int w, int h) : Screen(w, h), oled(U8G2_R2, /*reset*/ 16, /*clk*/ 15, /*data*/ 4)
--- a/include/screen.h
+++ b/include/screen.h
@@ -364,8 +364,8 @@ public:
         SSD1306Screen(int w, int h) : Screen(w, h)
         {
             Heltec.begin(true /*DisplayEnable Enable*/, false /*LoRa Enable*/, false /*Serial Enable*/);
-            #ifdef SCREEN_ROTATION
-                Heltec.display->screenRotate(SCREEN_ROTATION);
+            #if ROTATE_SCREEN
+                Heltec.display->screenRotate(ANGLE_180_DEGREE);
             #endif
         }
 
--- a/include/screen.h
+++ b/include/screen.h
@@ -314,8 +314,11 @@ public:
         U8G2_SSD1306_128X64_NONAME_F_HW_I2C oled;
 
     public:
-
-        OLEDScreen(int w, int h) : Screen(w, h), oled(U8G2_R2, /*reset*/ 16, /*clk*/ 15, /*data*/ 4)
+        #if ARDUINO_HELTEC_WIFI_LoRa_32_V3
+            OLEDScreen(int w, int h) : Screen(w, h), oled(U8G2_R0, /*reset*/ 21, /*clk*/ 18, /*data*/ 17)
+        #else
+            OLEDScreen(int w, int h) : Screen(w, h), oled(U8G2_R2, /*reset*/ 16, /*clk*/ 15, /*data*/ 4)
+        #endif
         {
             oled.begin();
             oled.clear();
--- a/include/screen.h
+++ b/include/screen.h
@@ -360,7 +360,7 @@ public:
 
         SSD1306Screen(int w, int h) : Screen(w, h)
         {
-            Heltec.begin(true /*DisplayEnable Enable*/, true /*LoRa Disable*/, false /*Serial Enable*/);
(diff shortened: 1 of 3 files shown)


include/screen.h
```diff
@@ -360,8 +363,10 @@ class Screen : public GFXBase
 
         SSD1306Screen(int w, int h) : Screen(w, h)
         {
-            Heltec.begin(true /*DisplayEnable Enable*/, true /*LoRa Disable*/, false /*Serial Enable*/);
-            Heltec.display->screenRotate(ANGLE_180_DEGREE);
+            Heltec.begin(true /*DisplayEnable Enable*/, false /*LoRa Enable*/, false /*Serial Enable*/);
+            #ifdef SCREEN_ROTATION
+                Heltec.display->screenRotate(SCREEN_ROTATION);
+            #endif
```
rbergen, line 369: Why this change?

platformio.ini
```diff
@@ -394,6 +403,14 @@ build_flags     = -DDEMO=1
                   -DUSE_SSD1306=1
                   ${dev_heltec_wifi_v3.build_flags}
 
+; Heltec LoRa V3 board, which is an S3 with USB-C and LoRa
+[env:helteclorav3demo]
+extends         = dev_heltec_wifi_lora_v3
+build_flags     = -DDEMO=1
+                  -DENABLE_WIFI=1
+                  -DUSE_SSD1306=1
```
rbergen, line 411: I know you copied this from the `heltecv3demo` env, but would you agree this should really be a define in the `dev_heltec_wifi_*` sections? (As in, it's a device trait, not a project one.)

rbergen (commented):
Thanks for submitting this. I have a couple of questions and (for now) one change request. 

Can you confirm that with these changes applied, you _do_ have a working Heltec Wifi Lora V3 board, including display?

include/screen.h
BigDurl: On the LoRa version if you have it in the plastic case/housing thing from heltec the display is upside down and I figure might as well add a way to easily define the screens rotation.

platformio.ini
BigDurl: I don't understand the code enough yet to give my opinion on were that should be defined. I did it this way because it's the same as all the other heltc env.

BigDurl:
Yes, everything is working now.

include/screen.h
rbergen: You mean that the original code turned the screen upside down on the LoRa version? (Which is indeed what rotating by 180 degrees should do, I guess...)

But in that case I think we should retain the old behavior for the non-LoRa version - the rotation was added to that for a reason. That is to say that for the non-LoRa version, `SCREEN_ROTATION` should then be defined to be `ANGLE_180_DEGREE`, maybe best within the definition block for the DEMO project in globals.h.

platformio.ini
rbergen: Fair enough. I'll chew on this myself some more when I have a little more time.

rbergen (commented):
Thanks for your responses and the commit.

Based on one of your answers I think one more change is required to make sure we retain the current behavior on non-LoRa boards.

include/screen.h
BigDurl: Actually it should be defined in the 'platformio.ini' for any devices or environments that needs it. I do not know which devices need which rotations or why the code originally specified rotating the display 180 degrees. I suspect this was personal preference of whoever originally wrote that section of code. On the non Lora version it doesn't really matter which way the screen is oriented as far as I know.

rbergen: My point is that we should not make the existing projects/environments act differently while adding another. In the current heltecv3demo project the screen is rotated 180 degrees, so we should keep it that way even if we do not want to do the same for the LoRa version. 

Concerning the location of the define, I understand your statement, but this taps into a long-running conversation about what build config goes where. I'll make a decision about this myself and implement the define accordingly in this PR when I also consider the location of the `USE_SSD1306` one. (There's actually another aspect, which is that we prefer to not use `#ifdef` conditionals. I'll also include that in my considerations.)

rbergen:
@BigDurl I've made, committed and pushed the following changes:

- I removed the heltec.h board defines (`WIFI_*_32_V3`) from the dev blocks in platformio.ini. It turns out they are already included in the board configurations provided by PlatformIO, so including them in platformio.ini yielded a heap of macro redefinition warnings.
- I moved the setting of `USE_SSD1306` to the dev blocks, in line with what I said before. What type of screen is connected is a board/device trait, not a project feature.
- I converted the `SCREEN_ROTATION` define to a `ROTATE_SCREEN` feature flag, that makes the code in screen.h rotate the screen 180 degrees when set to 1. I've set that flag to 1 for the heltecv3demo env in platformio.ini as you suggested.

I think that with this, Heltec WiFi Kit/LoRa V3 defines are where they belong, and structured in line with the rest of NightDriverStrip.

Now that it seems that CI is happy with the PR as it stands, please have a look to see if you can also agree with the changes I made. If so, I'll merge.

BigDurl:
Everything looks good.
One thing we might have to address eventually is. Currently we can only easily set the screen rotation to 0 and 180 and
I'm sure there's going to be cases where someone will want there display at 90 and 270.

rbergen:
That's a fair comment, but if we do run into this then we will have to address another (and probably bigger) issue as well: the on-device displays are never square, so a simple non-180-degree rotation won't work anyway. Which is why I was comfortable with doing it this way, at least for now.

PR #670, opened 2024-11-29
Make Stock effect compare against previous day's close

### LINKED ISSUE #662: Tweak StockTicker effect
The stock ticker effect currently has all the data it needs to display a proper stock quote, but the change is not displayed correctly.

The intent of the effect is to show CHANGE SINCE LAST CLOSE.  So the code needs to likely do something like look for the last price on a prior calendar day to determine the last close price (or it may be in the data).

The little graphic it displays should also be an indication of the total change from last close to now (though it only shows the most recent 64 data points).

If you go to Google and search for "MSFT stock quote", it will show you the correct delta that we should be showing, give or take 20-30 minutes of activity (our quotes are a bit delayed).

rbergen:
## Description

This closes #662 by making the effect compare against the previous day's close instead of today's open. It does this both in the numeric change shown, as in the graph underneath it.

There is still one "peculiarity" that people may notice, which is that the data points in the graph include after-hour trading, and we don't show this in the coloring of the graph (Google and other sources tend to show after-hour trading in a different color).

This means that the last data point for a calendar day may be positive (green) while the (correct) day change is negative, or vice versa. If we want to change this then we need the stockserver to indicate in its response to the effect what data points are outside of regular trading hours.

Also, the default host for the stockserver changed from davepl.com to davepl.dyndns.org, to better handle IP address changes for Dave's Internet connection.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternStocks.h
+++ b/include/effects/matrix/PatternStocks.h
@@ -61,7 +61,7 @@ using namespace std::chrono_literals;
 #define STOCKS_UPDATE_INTERVAL_SECONDS 6s
 #define STOCKS_FETCH_INTERVAL_SECONDS  60s
 
-#define DEFAULT_STOCK_SERVER           "davepl.com:8888"
+#define DEFAULT_STOCK_SERVER           "davepl.dyndns.org:8888"
 #define DEFAULT_TICKER_SYMBOLS         "AAPL,AMZN,TSLA,MSFT"
 
 // AnimatedText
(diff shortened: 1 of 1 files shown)


davepl (approved):
Nice!  Simple fix, doesn't mean it was simple to figure out :-)

rbergen:
Thanks! The hard work was actually done in the back-end (stockserver), where it belongs, a few weeks ago. :)

PR #680, opened 2025-01-18
Add section on COM problems to README

rbergen:
## Description

Adds a section to the README concerning COM port problems on the Windows platform.

It includes another minor change to the GitHub CI workflow file so it no longer uses jq's --argfile option, which is no longer available in jq on Ubuntu 24.04.

## Contributing requirements

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,7 +19,7 @@ jobs:
       id: envs
       run: |
         if [[ $(< config/ci_exclude.json) != '[]' ]]; then echo "::warning title=Some environments were skipped::Environments not built: $(< config/ci_exclude.json)"; fi
-        echo "envs=$(tools/show_envs.py | jq -c --argfile ex config/ci_exclude.json '. - $ex')" >> $GITHUB_OUTPUT
+        echo "envs=$(tools/show_envs.py | jq -c --slurpfile ex config/ci_exclude.json '. - $ex[0]')" >> $GITHUB_OUTPUT
 
     outputs:
       envs: ${{ steps.envs.outputs.envs }}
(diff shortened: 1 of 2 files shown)


rbergen:
"Crazy" is a rather heavy qualification which I won't apply, but after going over your comment and the second paragraph a few times I am certainly misreading the supposed misread, so to speak. 

Maybe that is because the second paragraph is intended as a blanket statement to cover both tines of the fork, where you may think it's only intended to cover the second? Also, I'm deliberately not describing how to to use either tool to do the analysis, in part because the exact steps to follow depend on the tool that is used and the exact configuration (there's quite a range of lower-level Windows device that can be handed an "MS-DOS" COM port moniker.) Going through all that would go too far for a README like this; the point is to put people on the track of investigation with tools that have proven to be useful, following that track is up to them. 

All that said, if you have an alternative phrasing that you think addresses the issue you are seeing, I'm happy to consider it as always.

rbergen:
> I think "crazy" is fair game when used on oneself. :-)

True. Still doesn't give me a free license to do the same to you too. ;)

> How about:
> "If you experience such difficulties accessing serial ports, the tools in..."

Proposal accepted.

> Additionally, given the time we all just spent chasing seemingly known Malwarebytes defects, I'm tempted to throw them under the bus in the first paragraph and then back up the bus and do it again in the second. "...culprit, as may be disabling your antivirus." I feel less strongly about that change.

I tend to draw the line at recommending to the general public that they disable virus scanning. The advice could be "get something else that works", but that feels like overstepping my mandate in this context.

> +1, Approval.

Thank you sir!

I'll update the second paragraph in line with your first proposal and then merge.

PR #687, opened 2025-01-29
Misc minor enhancements in gfxbase

robertlipe:
Minor readability fixes in gfxbase.
    Minor readability fixes in gfxbase.

    setPalette: use STL to find entry in a table instead of nine separate
    string compares. The functions above and below really should be combined
    similarly.

    BresenhamLine():
      More faithful reproduction of published implementations.
      Computes pixel index only once.
      Since we're amplifying 'e2' term, no need to recheck y in each iteration.
      Keep sign of err in the most commonlhy used form instead of double negativ
e.

Tested:   Run all effects on Mesmerizer hardware, our most demanding build of   that code.**

## Description
I know this isn't your favorite tyep of PR, but technically it DOES improve
BlinkenPerBit by taking fewer bits to do the same thing. Use of STL maximixes
SWE brain cell consumption: find() finds things. Smarter data. Dumber code.ZZ

## Contributing requirements

### DIFF
--- a/include/gfxbase.h
+++ b/include/gfxbase.h
@@ -1128,38 +1128,38 @@ public:
 
     void BresenhamLine(int x0, int y0, int x1, int y1, CRGB color, bool bMerge = false) const
     {
-	int dx = abs(x1 - x0); // Delta in x direction
-	int dy = abs(y1 - y0); // Delta in y direction
-	int sx = (x0 < x1) ? 1 : -1; // Step in x direction
-	int sy = (y0 < y1) ? 1 : -1; // Step in y direction
+        int dx = abs(x1 - x0); // Delta in x direction
+        int dy = abs(y1 - y0); // Delta in y direction
+        int sx = (x0 < x1) ? 1 : -1; // Step in x direction
+        int sy = (y0 < y1) ? 1 : -1; // Step in y direction
 
-	int err = dx - dy; // Initial error term
+        int err = dx - dy; // Initial error term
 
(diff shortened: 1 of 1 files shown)


include/gfxbase.h
```diff
@@ -1132,41 +1128,40 @@ class GFXBase : public Adafruit_GFX
...
+            }
 
             if (x0 == x1 && y0 == y1)
-                break;
+                break; // Exit the loop once we've reached the destination
 
-            int e2 = 2 * err;
-            if (e2 >= dy)
+            int e2 = 2 * err; // Error term multiplied by 2 for efficiency. Saves second test for Y.
+            if (e2 > -dy) // Move in the x direction if needed
```
rbergen, line 1151: Are you sure about the switch from "greater or equal" to "greater than"? Same question for the "less than" change 7 lines down. Phrased differently and actually more to the point: why _should_ this be changed?

rbergen:
Is that your answer or a question? ;)

Anyway, fair enough. I asked because it was the one difference I noticed that wasn't mentioned in a "sub-bullet" in the `BresenhamLine()` section of your PR description.

Back to the PR: I'm indeed reluctant with merging PRs that go against the "if it ain't broken, don't try to fix it" premise from a functional perspective. At the same time, that "rule" is not set in stone either.

In this case, the use of STL is objectively better, and the Bresenham's line implementation is cleaner.

So, PR approved, merge impending.

PR #693, opened 2025-02-23
Fix OLED behavior on Heltec Wifi Kit 32 V2 and V3

SpeedPermitting:
## Description
Updated code to ensure consistent OLED display behavior among most supported Heltec boards.

## Overview
This code builds/works as expected for the following device models, with no local modifications except to create/populate a secrets.h file:

- Heltec Wifi Kit 32 (V2): pio run -e heltecv2demo -t upload
- Heltec Wifi Kit 32 (V3): pio run -e heltecv3demo -t upload
- Heltec Wifi LoRa 32 (V3): pio run -e helteclorav3demo -t upload

The "CurrentEffectSummary" screen now displays correctly (i.e., legible, but monochrome) on all three devices.

Prior to my changes, the code on the main branch resulted in different behavior for each device:

- Heltec Wifi Kit 32 (V2): Screen flashes solid blocks of pixels briefly on device boot or whenever any would-be-displayed data updates (this prompted my discussion #692 ).
- Heltec Wifi Kit 32 (V3): Screen remains on persistently, but displays only two solid blocks of pixels (similar to issue #667  for the LoRa 32).
- Heltec Wifi LoRa 32 (V3): Screen displays correctly and remains on persistently.

## Change Details

**platformio.ini**
I added the build_flag "-DUSE_SSD1306=1" and lib_dep "heltecautomation/Heltec ESP32 Dev-Boards @ ^1.1.1" to "[dev_heltec_wifi_v2]". Both had been added previously to the "wifi_v3" and "lora_v3" environments. This fixed the issue where the "CurrentEffectSummary" screen won't remain visible on the V2 board without constantly forcing a full redraw.

**globals.h**
I simplified the section under "#if ARDUINO_HELTEC_WIFI_KIT_32", now that 3 of 4 supported Heltec boards should all use a consistent config. Basically, everything with this flag should set "USE_OLED", and the only time we still want to reduce "NUM_INFO_PAGES" to '1' is when "USE_SSD1306" isn't set, i.e., for the original Heltec Wifi Kit 32.

I also moved the "default" setting of "NUM_INFO_PAGES" to a new location (line 1531) to avoid compiler warnings for a duplicate definition. My rationale for the new location is that this section seems to contain display-related defaults, vs. environment-related defaults.

**screen.h**
I changed a comment to reflect that the SSD1306Screen section already applies to multiple Heltec models.

**screen.cpp**
Here I changed the condition for setting the background of the "CurrentEffectSummary" screen to black. My thinking is that the reason for this conditional is a specific display type more than a specific board.

This alternative seems less elegant to me, but I don't know which is "better" in terms of proper scoping:
`#if ARDUINO_HELTEC_WIFI_KIT_32_V2 || ARDUINO_HELTEC_WIFI_KIT_32_V3 || ARDUINO_HELTEC_WIFI_LORA_32_V3`

## Questions

Is it worth trying to track down somebody with an original Heltec Wifi Kit 32 to see if we can make the display behavior consistent for all Heltec devices? The only reason I'm still including the conditional to set "NUM_INFO_PAGES" to '1' is that I don't have an original/"V1" Heltec to test with, and without testing, forcing the "BasicInfoSummary" screen seems like a "safer" option.

## Contribution Guidelines

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -1334,10 +1334,6 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 #define NUM_RINGS 1
 #endif
 
-#ifndef NUM_INFO_PAGES
-#define NUM_INFO_PAGES 2
-#endif
-
 #ifndef COLOR_ORDER
 #define COLOR_ORDER EOrder::GRB
 #endif
@@ -1480,6 +1476,10 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 
 #endif // end USE_SCREEN
 
+#if USE_OLED
+    #define NUM_INFO_PAGES 1        // Only allow "Basic Info Summary" page on monochrome OLED screens
+#endif
+
 #if USE_LCD
     // These pins are based on the Espressif WROVER-KIT, which uses an ILI9314 chipset for its display
     // connected as follows:
@@ -1532,6 +1532,10 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 
 #ifndef MATRIX_CENTER_Y
 #define MATRIX_CENTER_Y ((MATRIX_HEIGHT + 1) / 2)
+#endif
+
+#ifndef NUM_INFO_PAGES
+#define NUM_INFO_PAGES 2
 #endif
 
 // When you press a color button on the remote, the color is used to create a temporary fill effect, but
--- a/include/globals.h
+++ b/include/globals.h
@@ -1441,7 +1441,7 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
 
         #define USE_OLED 1                                    // Enable the Heltec's monochrome OLED
         #ifndef USE_SSD1306
-            #define NUM_INFO_PAGES 1        // Only allow "Basic Info Summary" page on monochrome OLED screens
+            #define NUM_INFO_PAGES 1        // Only display "BasicInfoSummary" if not SSD1306
         #endif
 
     #elif USE_M5                                        // screen definitions for m5stick-c-plus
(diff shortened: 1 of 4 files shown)


include/globals.h
```diff
@@ -1440,15 +1436,13 @@ extern RemoteDebug Debug;           // Let everyone in the project know about it
...
-
-        #define USE_OLED 1
-        #define USE_SSD1306 1
-
-    #elif ARDUINO_HELTEC_WIFI_KIT_32
+    #if ARDUINO_HELTEC_WIFI_KIT_32
                         // screen definations for heltec_wifi_kit_32 or heltec_wifi_kit_32_v2
 
         #define USE_OLED 1                                    // Enable the Heltec's monochrome OLED
+        #ifndef USE_SSD1306
```
rbergen, line 1443: `USE_SSD1306` is a flag that can also be defined to 0, in which case the line below should not be included. I therefore think the `#ifndef` should be replaced, as follows:

```suggestion
        #if !(USE_SSD1306)
```

SpeedPermitting: Makes sense to me. I tested this change locally. Everything still worked as expected, and I verified omitting the "-DUSE_SSD1306=1" flag from the environment in platformio.ini will trigger this conditional, as expected.

rbergen:
LGTM, and CI is happy, so I'll proceed with the merge. Thanks again for submitting this!

To answer the question you included in your PR description:

> Is it worth trying to track down somebody with an original Heltec Wifi Kit 32 to see if we can make the display behavior consistent for all Heltec devices? 

Conceptually yes, but in the practical sense I wouldn't know how to. I'm happy with choosing the "safer" option for now - which I agree is the one you chose -, and wait for an actual original Heltec Wifi Kit 32 owner to report themselves, if and when that ever happens.

PR #702, opened 2025-03-17
update arduinoFFT to ^2.0 and FastLED to ^3.8.0

sixfootdad:
## Description
Upgrading the FastLED & arduinoFFT dependencies to the latest version. There were no breaking changes in more recent versions of FastLED so the upgrade was merely adding the caret in platform.ini, but arduinoFFT required a few code changes.

## Contributing requirements
I believe they do

### DIFF
--- a/include/soundanalyzer.h
+++ b/include/soundanalyzer.h
@@ -33,7 +33,7 @@
 
 #pragma once
 
-#include <arduinoFFT.h>
+#include <ArduinoFFT.h>
 #include <driver/i2s.h>
 #include <driver/adc.h>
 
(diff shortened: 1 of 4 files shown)


rbergen:
@sixfootdad Hi, thanks for submitting this.

To what extent did you test these changes? As in, which devices, projects, etc. did you use?

I also think it's best to reverse the main() change in videoserver.py. There's nothing to import in that file, and it really is intended as a straightforward Python script. In that context, wrapping all the code in main() is an unnecessary complication in my view.

rbergen:
Yeah, it does. Once you open a PR it tracks the head branch (sixfootdad:upstream) until the changes are merged into the base branch (PlummersSoftwareLLC:main). That allows the PR to be updated after reviews, etc., which is exactly what we'll be doing here. :)

It's also why I tend to use feature branches (i.e. create a separate branch in my fork for each change I plan to make), so I can indeed isolate changes if I plan to work on one thing while I have a PR already open for another.

If you tested the dependency switch in the videoserver Python scripts then I'm happy to include that too, it's just the main() thing I'd like to be reverted.

rbergen:
@sixfootdad I had to change the include of `arduinoFFT.h` in soundanalyzer.h to once again use the lower-case "a", to make builds work for me locally and in CI. I confirmed that in both cases, a 2.x version of arduinoFFT gets pulled in (2.0.4, to be specific). 

Can you confirm that the version that is now in your "upstream" branch also builds locally for you?

rbergen:
@sixfootdad I confirmed using a number of sources that `arduinoFFT.h` is the correct header filename to use. I'll therefore merge this PR in its current state.

Thanks again for submitting this!

rbergen:
It's not a problem. If we didn't need CI to occasionally catch something, I would have implemented it for nothing. ;)

Anyway, while looking into it I found a number of typos in the arduinoFFT README, opened a PR towards upstream to fix them, and had it merged within minutes. So, thanks to this confusion I am now forever featured in arduinoFFT's commit history. :)

PR #705, opened 2025-03-25
Improve JSON serialization memory overflow handling

rbergen:
## Description

This improves the handling of ArduinoJSON serialization memory overflows. Instead of triggering a device reset in case a JSON serialization memory overflow condition occurs, an error is logged and the failed serialization attempt is abandoned.

## Contributing requirements

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -222,9 +222,7 @@ class DeviceConfig : public IJSONSerializable
         if (includeSensitive)
             jsonDoc[OpenWeatherApiKeyTag] = openWeatherApiKey;
 
-        assert(!jsonDoc.overflowed());
-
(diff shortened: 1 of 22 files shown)


rbergen:
I'll take that for now. :) There's no rush here as far as I'm concerned. I do think making this change would be an improvement, but the current state of affairs is in itself not burning down houses or eating pets. And yes, if the analysis of the problematic environment is proven to be helped by this PR, then that would be the eating that proves the pudding, in my book.

davepl:
I wouldn’t assume too much.  I actually heard they’re eating the pets somewhere!

rbergen:
I don't think there's a "safe" way to respond to that remark, so I won't. 🙂

PR #709, opened 2025-04-01
Fix VU meter jitteriness in silent environments

rbergen:
## Description

This fixes an issue where the VU meter on matrix projects (Mesmerizer specifically) would show excessive movement in silent environments. This was introduced with #660 and is discussed in #706.

## Contributing requirements

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -54,7 +54,7 @@ void IRAM_ATTR AudioSamplerTaskEntry(void *)
 
     for (;;)
     {
-        uint64_t lastFrame = millis();
+        auto lastFrame = millis();
 
         g_Analyzer.RunSamplerPass();
         g_Analyzer.UpdatePeakData();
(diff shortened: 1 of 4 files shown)


src/audio.cpp
```diff
@@ -54,33 +54,34 @@ void IRAM_ATTR AudioSamplerTaskEntry(void *)
...
         g_Analyzer.RunSamplerPass();
         g_Analyzer.UpdatePeakData();
         g_Analyzer.DecayPeaks();
 
         // VURatio with a fadeout
 
-        static float lastVU = 0.0;
-        constexpr auto VU_DECAY_PER_SECOND = 4.0;
+        static auto lastVU = 0.0f;
+        constexpr auto VU_DECAY_PER_SECOND = 1.25f;
```
rbergen, line 66: This is now aligned with the decay for VU meter effects, as defined in soundanalyzer.h.

rbergen, line 66: This is now aligned with the decay of the various VU effects, as defined in soundanalyzer.h.

src/audio.cpp
```diff
@@ -54,33 +54,34 @@ void IRAM_ATTR AudioSamplerTaskEntry(void *)
...
-            lastVU += (millis() - lastFrame) / 1000.0 * VU_DECAY_PER_SECOND * VU_REACTIVITY_RATIO;
+            lastVU += frameDurationSeconds * VU_DECAY_PER_SECOND * VU_REACTIVITY_RATIO;
         else
-            lastVU -= (millis() - lastFrame) / 1000.0 * VU_DECAY_PER_SECOND;
+            lastVU -= frameDurationSeconds * VU_DECAY_PER_SECOND;
 
-        lastVU = std::max(lastVU, 0.0f);
-        lastVU = std::min(lastVU, 2.0f);
-        g_Analyzer._VURatioFade = lastVU;
+        g_Analyzer._VURatioFade = std::clamp(lastVU, 0.0f, g_Analyzer._VURatio);
```
rbergen, line 79: This is the actual fix for the problem mentioned in the PR description. The VU ratio wasn't capped to the value of `g_Analyzer._VURatio`, which it has to be.

rbergen:
Maybe. As it stands, we're still in the process of fixing a bug that got introduced half a year ago. This particular change is in the process of being tested; I put the PR up so that the review process could be run in parallel. 

Which is to say that I don't yet know if this particular alignment is indeed going to work the way I think it will. If that turns out to be the case then we can look at commonalities between the two places where audio-related calculations now take place. 

In a very small sense this already happened: `VU_REACTIVITY_RATIO` has already been extracted from both places. Maybe the same can be done for `VU_DECAY_PER_SECOND`, for starters. I guess we'll know once @heidepiek has executed his test round, this evening. :)

rbergen:
I've actually taken a quick look just now, and I think we can quite easily move the calculation logic in audio.cpp to the existing `SoundAnalyzer` class, merge that class with the `AudioVariables` struct it now derives from and replace direct member variable access with accessor methods - which would then be getters only bar maybe one exception. 

Partly depending on the exact VU meter code we end up with in the context of this bugfix, further optimizations can be looked at as well. There does seem to be some overlap between the calculations in `SoundAnalyzer::UpdatePeakData()` and `SoundAnalyzer::DecayPeaks()` on the one hand, and the VU meter calculations in audio.cpp on the other. But there are also differences: the former combination first increases and then decreases, the latter logic does one _or_ the other.

Anyway, first things first, which is bringing this bugfix to a conclusion.

PR #720, opened 2025-05-09
Implement Cyclic Cellular Automata pattern

robertlipe:
Alternates between Greenberg-Hastings and Cyclic models.
https://en.wikipedia.org/wiki/Greenberg%E2%80%93Hastings_cellular_automaton
should make all the math intuitively obvious to even the most casual
observer...

Future work: Perhaps use audio ractivity to excite ColorFromCellState
by frequency and using beat detection to repopulate the Seed functions
would be interesting.

"Thanx, Gemini!"

## Description
Add new effect for Mesmerizer. Somewhat like the 'life' patterns, Cellular
Automata allows patterns that grow and move about the display.

The math came from a variety of sources and in a first fro me, the
coding was very much done by Gemini.

"Hendrix goes Electric."

There are the #if RUTGER blocks where Gemini tried to add code
for settings but neither one of us could derive a sensible thing
from the examples, so Rutger and I need to work that out. (Or just
pull it...) Tweaking those variables CAN result in visually interesting
patterns, even if it's via recompilation.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternCyclicCA.h
+++ b/include/effects/matrix/PatternCyclicCA.h
@@ -1,0 +1,581 @@
+//+--------------------------------------------------------------------------
+//
+// File:        PatternCyclicCA.h
+//
+// NightDriverStrip - (c) 2025 Plummer's Software LLC.  All Rights Reserved.
+//
+// This file is part of the NightDriver software project.
+//
+//    NightDriver is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    NightDriver is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTlichkeit AND FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU General Public License for more details.
+//
+//    You should have received a copy of the GNU General Public License
+//    along with Nightdriver.  It is normally found in copying.txt
+//    If not, see <https://www.gnu.org/licenses/>.
+//
+//
+// Description:
+//
+//   Implements a 2D Cyclic Cellular Automata pattern, including a
+//   Greenberg-Hastings variant, based on a JavaScript implementation.
+//   This file combines the header and implementation for simpler
+//   integration into projects following this pattern.
+//
+//
+//---------------------------------------------------------------------------
+
+#ifndef PatternCyclicCA_H
+#define PatternCyclicCA_H
+
+// Define the possible modes for the Cellular Automata
+enum class CAMode
+{
+    GreenbergHastings,
+    Cyclic
+};
+
+#define RUTGER 0
+
+class PatternCyclicCA : public LEDStripEffect
+{
+private:
+#if RUTGER
+    // Define the setting specifications for the UI
+    // This requires a matching INIT_EFFECT_SETTING_SPECS() in effects.cpp or linker errors will ensue
+    DECLARE_EFFECT_SETTING_SPECS(mySettingSpecs);
+#endif
+    // Two buffers for the cellular automaton state
+    std::vector<std::vector<uint8_t>> m_buffer1;
+    std::vector<std::vector<uint8_t>> m_buffer2;
+
+    // Pointers to the current and next generation buffers
+    // Swapping these pointers is faster than copying vector contents
+    std::vector<std::vector<uint8_t>>* m_pb1; // Pointer to the previous generation buffer
+    std::vector<std::vector<uint8_t>>* m_pb2; // Pointer to the current generation buffer
+
+    // Pattern parameters, exposed as UI settings
+    uint8_t m_numStates = 24;      // Number of possible states for each cell
+    int m_speed = 60;              // Delay in milliseconds between generations
+    unsigned long m_lifetime = 10000; // How long a pattern runs before re-seeding (0 = forever)
+    float m_excited = 0.03;        // Probability of a cell starting in the 'excited' state (for GH)
+    float m_refractory = 0.64;     // Probability of a cell starting in a 'refractory' state (for GH)
+    uint8_t m_threshold = 1;       // Minimum number of neighbors in the 'next' state to trigger a transition
+    CAMode m_mode = CAMode::GreenbergHastings; // Use enum class for mode
+
+    // Function pointer to the current generation calculation method
+    void (PatternCyclicCA::*m_calcNextGen)();
+
+    uint8_t m_nextVal = 1;         // The 'target' state for neighbors to trigger a transition
+    unsigned long m_frameTimer = 0;    // Accumulator for the generation timer
+    unsigned long m_patternTimer = 0;  // Accumulator for the pattern lifetime timer
+    unsigned long m_lastDrawTime = 0;  // Time of the last Draw call, for calculating delta
+
+    int m_sum = 0; // Tracks changes in the last generation; used to detect if pattern died
+
+    // Helper functions for toroidal wrapping (wrapping around the matrix edges)
+    int WrapX(int x) const { return (x % MATRIX_WIDTH + MATRIX_WIDTH) % MATRIX_WIDTH; }
+    int WrapY(int y) const { return (y % MATRIX_HEIGHT + MATRIX_HEIGHT) % MATRIX_HEIGHT; }
+
+    // Allocate and initialize the frame buffers
+    void AllocateFrameBuffers()
+    {
+        // Resize the vectors to the matrix dimensions
+        m_buffer1.resize(MATRIX_WIDTH, std::vector<uint8_t>(MATRIX_HEIGHT));
+        m_buffer2.resize(MATRIX_WIDTH, std::vector<uint8_t>(MATRIX_HEIGHT));
+
+        // Set initial buffer pointers
+        m_pb1 = &m_buffer1;
+        m_pb2 = &m_buffer2;
+    }
+
+    // Swap the roles of the previous and current buffers
+    void SwapBuffers()
+    {
+        std::swap(m_pb1, m_pb2);
+    }
+
+    // Count neighbors in the 'm_nextVal' state using a 4-neighbor (von Neumann) neighborhood
+    int SumNeighborhood4(int x, int y, const std::vector<std::vector<uint8_t>>& buffer) const
+    {
+        int xm = WrapX(x - 1); // Use helper function
+        int xp = WrapX(x + 1); // Use helper function
+        int ym = WrapY(y - 1); // Use helper function
+        int yp = WrapY(y + 1); // Use helper function
+
+        int count = 0;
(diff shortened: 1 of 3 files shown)


include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+
+#define RUTGER 0
+
+class PatternCyclicCA : public LEDStripEffect
+{
+private:
+#if RUTGER
+    // Define the setting specifications for the UI
+    // This requires a matching INIT_EFFECT_SETTING_SPECS() in effects.cpp or linker errors will ensue
+    DECLARE_EFFECT_SETTING_SPECS(mySettingSpecs);
```
rbergen, line 52: This looks right.

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+    }
+
+
+protected:
+#if RUTGER
+    // Define the specifications for the UI settings sliders.
+    EffectSettingSpecs* FillSettingSpecs() override
+    {
+        return &mySettingSpecs;
+    }
```
rbergen, line 320: This function does need to actually create the `SettingSpec` instances for the settings this effect publishes; the Web UI uses this information to correctly display and validate the settings' values. You can refer to the implementation of this function for `PatternStocks` or `PatternSubscribers` for examples.

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+            bool newModeBool = false;
+            if (value.equalsIgnoreCase("true") || value.equalsIgnoreCase("1")) {
+                newModeBool = true;
+            } else if (value.equalsIgnoreCase("false") || value.equalsIgnoreCase("0")) {
+                newModeBool = false;
+            } else {
+                // Handle potential string names for modes if needed, e.g., "Greenberg-Hastings", "Cyclic"
+                // For now, assume boolean or integer string representation
+                return false; // Value not recognized
+            }
```
rbergen, line 346: I think we need to make a decision here about the `mode` setting type. It's either a boolean (in which case it should follow the parsing rules as implemented in the helper function `BoolFromText()` which is thus the most obvious way to parse the setting value), or it's a string. The setting type also has to be communicated in the `SettingSpec` instance for this setting, and stuff is displayed and parsed in the Web UI accordingly.

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+        // Serialize our specific settings.
+        // Using NAME_OF macro for consistency with SetSetting and FillSettingSpecs
+        jsonDoc[NAME_OF(m_numStates)] = m_numStates;
+        jsonDoc[NAME_OF(m_speed)] = m_speed;
+        jsonDoc[NAME_OF(m_lifetime)] = m_lifetime;
+        jsonDoc[NAME_OF(m_excited)] = m_excited;
+        jsonDoc[NAME_OF(m_refractory)] = m_refractory;
+        jsonDoc[NAME_OF(m_threshold)] = m_threshold;
+        // Serialize the enum as a boolean for compatibility with the existing JSON structure
+        jsonDoc[NAME_OF(m_mode)] = (m_mode == CAMode::Cyclic);
```
rbergen, line 404: Same remark about setting names (camelCase without the "m_" prefix).

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+    // Constructor for deserialization from JSON
+    PatternCyclicCA(const JsonObjectConst& jsonObject) : LEDStripEffect(jsonObject)
+    {
+        // Deserialize settings from JSON, using default values if keys are not found
+        m_numStates = jsonObject["numStates"] | m_numStates;
+        m_speed = jsonObject["speed"] | m_speed;
+        m_lifetime = jsonObject["lifetime"] | m_lifetime;
+        m_excited = jsonObject["excited"] | m_excited;
+        m_refractory = jsonObject["refractory"] | m_refractory;
+        m_threshold = jsonObject["threshold"] | m_threshold;
```
rbergen, line 442: - By project convention, the JSON properties used to persist settings to storage (and hence construct effects from storage) are three-letter acronyms.
- You can use the "fetch from JSON object or use default" syntax if you like, but I'd recommend using the `if (….is<`_`type`_>())` pattern to check if the property in the `jsonObject` is of the correct type, and only use the property value if it is.

Again, for an example of these two points you can refer to the `JsonObjectConst` constructor of `PatternStocks`.

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+
+        // Serialize our specific settings (using short keys for efficiency in JSON)
+        jsonDoc["ns"] = m_numStates;
+        jsonDoc["sp"] = m_speed;
+        jsonDoc["lt"] = m_lifetime;
+        jsonDoc["ex"] = m_excited;
+        jsonDoc["rf"] = m_refractory;
+        jsonDoc["th"] = m_threshold;
+        // Serialize the enum as a boolean for compatibility with the existing JSON structure
+        jsonDoc["mo"] = (m_mode == CAMode::Cyclic);
```
rbergen, line 559: - As mentioned before, these should be _three-letter_ acronyms.
- In all cases they **must** match the property names referred to in the deserialization constructor, because otherwise deserialization will always fail.

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_speed), "Speed", "Milliseconds per frame", SettingSpec::SettingType::Int, 1, 200, 1),
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_lifetime), "Lifetime", "Pattern lifetime in ms (0=forever)", SettingSpec::SettingType::UInt, 0, 60000, 1000),
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_threshold), "Threshold", "Neighbors to advance state", SettingSpec::SettingType::UInt8, 1, 8, 1),
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_numStates), "States", "Number of cell states", SettingSpec::SettingType::UInt8, 2, 32, 1),
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_excited), "Excited %", "Initial excited cell percentage (GH)", SettingSpec::SettingType::Float, 0.0, 0.2, 0.01),
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_refractory), "Refractory %", "Initial refractory cell percentage (GH)", SettingSpec::SettingType::Float, 0.0, 0.8, 0.01),
+    // Keep as Bool type, assuming the UI handles mapping bool toggle to the two modes
+    EffectSettingSpecs::AddSettingSpec(NAME_OF(m_mode), "Mode", "0=Greenberg-Hastings, 1=Cyclic CA", SettingSpec::SettingType::Bool),
+};
```
rbergen, line 578: This won't work. In fact, I'm quite positive this code doesn't compile or link with this included, which is probably the reason for the `#if RUTGER` blocks?

In any case:
- The `SettingSpec` instances have to be added in `FillSettingSpecs()`, as indicated in an earlier comment.
- In all cases, a `SettingSpec::SettingType` enum value has to be chosen that actually exists.
- The `INIT_EFFECT_SETTING_SPECS` line has to be included in `effects.cpp`, in the same way that it's already included for `LEDStripEffect`, `PatternSubscribers` and `PatternStocks`; searching for `INIT_EFFECT_SETTING_SPECS` in `effects.cpp` will instantly dig up the lines in question.

rbergen (commented):
I've limited my review to the JSON (setting) serialization/deserialization stuff at first, because you specifically asked me to (well, sort of 🙂).

Let me know if any further clarifications are necessary.

include/effects/matrix/PatternCyclicCA.h
```diff
@@ -0,0 +1,581 @@
...
+    // Handle incoming setting changes from the UI.
+    bool SetSetting(const String& name, const String& value) override
+    {
+        // Use helper macros provided by the framework to handle common types.
+        RETURN_IF_SET(name, NAME_OF(m_speed), m_speed, value);
+        RETURN_IF_SET(name, NAME_OF(m_lifetime), m_lifetime, value);
+        RETURN_IF_SET(name, NAME_OF(m_threshold), m_threshold, value);
+        RETURN_IF_SET(name, NAME_OF(m_numStates), m_numStates, value);
+        RETURN_IF_SET(name, NAME_OF(m_excited), m_excited, value);
+        RETURN_IF_SET(name, NAME_OF(m_refractory), m_refractory, value);
```
rbergen, line 331: For some reason this comment dropped out of my previous review, so I'll repeat it (the "Same remark about setting names (camelCase without the "m_" prefix)" comment referred to this one - my apologies for any confusion caused by that):

By project convention, the JSON property names of settings use camelCase naming, without the "m_" prefix. That means you can't use the `NAME_OF()` macro if you use the "m_" prefix for member variables - unless you do a "+2" to whatever the macro generates.

include/effects/matrix/PatternCyclicCA.h
rbergen: I'm just noting that the comment this one refers to fell out of the first set of review comments. I've added it again, separately.

include/effects/matrix/PatternCyclicCA.h
rbergen: While looking at this again, it occurred to me there is a supported middle ground: you can drop the "m" but keep the underscore prefix for member variables, and then use the `ACTUAL_NAME_OF()` macro to extract the camelCase name for a member variable. The underscore prefix is commonly used within this project (hence the macro), the "m_" prefix is not.

rbergen:
> ...and I didn't see a reason to push forward with settings (which we widely underuse in effects, IMO)

I agree whole-heartedly, which is why I appreciated the effort you made towards adding them so much.

> I seem to have forgotten to actually SAY that in the review, so I may have wasted your time, Rutger, on technical stuff that might not matter anyway. Sorry!

No, please. I appreciate the effort you made, and I also didn't ask what feedback you were looking for first. It was my own decision to dive in with this focus and level of detail, straight away.

> So before we proceed, is something like
> https://robertlipe.com/tmp/PatternCyclicCA.html
> likely to be accepted?

Well, I like it. I think there are some things that can be looked at, particularly towards the end of the pattern's development, but you already mention those yourself.

At the same time, this being a very visual contribution directly involving blinking LEDs, I'll ask Dave for his opinion as well. He's also been busy with other things, so in terms if lead time on this PR, I don't think we actually lost a lot of traction we could otherwise have had.

He or I will get back to you as soon as there is an opinion to share!

rbergen:
@robertlipe Conclusion: the pattern looks interesting and sufficiently different from what we already have, but looks like it needs tuning - with which we're basically again coming back to what you already said yourself.

Anyway, I'd say there is sufficient reason to continue working on this.

PR #746, opened 2025-08-16
Make SoundAnalyzer a template, remove Mic type; Use proper TFT_eSPI includes, fines; 

davepl:
Gets rid of the runtime Mic information, makes it a templated argument.
Uses the proper setup headers instead of manual definitions in platformio.ini

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -532,7 +532,7 @@ void IRAM_ATTR ScreenUpdateLoopEntry(void *)
         
         constexpr float kMaxFPS = 60.0f;
         const auto targetDelay = PERIOD_FROM_FREQ(kMaxFPS) * MILLIS_PER_SECOND / MICROS_PER_SECOND;
-        delay(max(1.0, targetDelay - (millis() - lastFrame)));
+        delay((unsigned long)max(1.0, targetDelay - (millis() - lastFrame)));
 
         bRedraw = false;
     }


rbergen (commented):
Summary: it's not technically broken, but I _really, really_ don't like the `std::sum` thing.

rbergen:
> It is, technically, broken to use a reserved namespace. That's why they're reserved. But, we agree that it's icky, so let's please just not.

I used Dave's usual standard for "technically broken" in cases like these, which is "it doesn't compile". ;) But yeah, you and I agree this is very disagreeable, which I think is the main point. 

About the question I asked about the "why", I've now seen the answer from @davepl (I really need to remember to scroll down the PR page to look at email exchanges before I comment...), and IMHO that's not enough to hack into the `std` namespace. One of my own cosmetic objections to using `std` stuff is that it often makes the code that uses them more verbose - and to me less intuitive - than tailor-made, self-baked alternatives, but I accept that sticking to the standard when possible has benefits that far outweigh that objection - no matter how cleverly the AI-generated code has been put together.

I don't think the "this is shorter when calling" argument warrants deviating from the standard in this case, let alone _extending_  it. And I agree that there is an actual benefit of showing intent by explicitly setting the init value, even in non-`string` cases.

> This is why we're on 2b instead of 2a, even.

I'm sure the -std build flag refers to 2a?

> Will that specific aspect of ranges work? Maybe. 

I think you knew very well that it doesn't. :) I added another comment on my earlier one. Those approaches (one of which I label as a form of over-engineering myself) I have confirmed to compile. 

Again/still, the main thing is that I strongly believe the whole (`std::`)`sum()` thing shouldn't exist at all in the first place.

>> It's the syntax one would use if they want to say "don't do the following if USE_I2S_AUDIO is explicitly set to 0". I.e. default to it being 1 when undefined.

> But we've tried to avoid that, right? Since I just invented the USE_i2s_audio path a week ago (and you even called out that I, alone, had boards using it that i hadn't submitted yet) we should be able to NOT have a super strange case here. Let's try to make it less odd.

When you boil it down, this is basically a shorthand for:

```cpp
#ifndef USE_I2S_AUDIO
    #define USE_I2S_AUDIO 1
#endif

#if !USE_M5 && (ELECROW || USE_I2S_AUDIO)
```

I agree that code is easier to follow, and more in line with our usual approach.

include/globals.h
```diff
@@ -893,6 +894,20 @@ constexpr std::array<T, N> to_array(const T (&arr)[N]) {
...
 }
 
+// Provide a single-parameter std::accumulate overload for ranges/containers
+// This allows: auto total = std::accumulate(container);
+namespace std {
+    template <typename Range>
+    inline auto accumulate(const Range& r)
+        -> std::remove_cv_t<std::remove_reference_t<decltype(*std::begin(r))>>
+    {
+        using T = std::remove_cv_t<std::remove_reference_t<decltype(*std::begin(r))>>;
```
rbergen, line 904: `std::remove_cv_t<std::remove_reference_t<decltype(*std::begin(r))>>` can be replaced by the more concise `std::decay_t<decltype(*std::begin(r))>`.

You may have to `#include <type_traits>` to make it compile.

PR #754, opened 2025-08-21
Simplify effect ID implementation

rbergen:
## Description

This builds on #748 by implementing the effect ID (renamed from `kId`) and `effectId()` static members in a class template between `LEDStripEffect` and the effects called `EffectWithId`, based on an `EffectId` that is passed as a template parameter when defining the effect class.

This:

- Removes the need for an effect-specific definition of `ID` and `effectId()` in most effect classes; the requirement remains only for effect classes that derive from `EffectWithId` indirectly, i.e. via another effect.
- Makes the setting of an `EffectId` for an effect class derived from `EffectWithId` a "pre-compile time check" in most development environments.
- Fixes a number of accessor specifiers, which made private members public.
- Removes a lot of trailing spaces, and makes the line white spacing in effect classes far more consistent.
- Removes unnecessary `Add...Disabled()` functions from effectsupport.h.

## Contributing requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,8 +230,8 @@ Example in globals.h:
 
 To add new effects, you:
 
-1. Derive from `LEDStripEffect` (or an existing effect class) and the good stuff happens in the only important function, `Draw()`.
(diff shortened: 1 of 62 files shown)


davepl (approved):
LGTM!

PR #759, opened 2025-08-23
Guard against clock underflow in drawing loop

davepl:
I finally caught the bug that could cause the spectrum project to drop to 1FPS for a while, but only after days of running.  If the time of day clock gets adjusted backwards, we could pick up a large time delta that persisted.

### DIFF
--- a/src/drawing.cpp
+++ b/src/drawing.cpp
@@ -30,6 +30,8 @@
 //---------------------------------------------------------------------------
 
 #include <mutex>
+#include <algorithm>
+#include <cmath>
 #include <ArduinoOTA.h> // Over-the-air helper object so we can be flashed via WiFi
 #include "globals.h"
 #include "colordata.h"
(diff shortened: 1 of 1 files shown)


src/drawing.cpp
```diff
@@ -156,10 +158,13 @@ int CalcDelayUntilNextFrame(double frameStartTime, uint16_t localPixelsDrawn, ui
 
     if (localPixelsDrawn > 0)
     {
-        const double minimumFrameTime = 1.0 / g_ptrSystem->EffectManager().GetCurrentEffect().DesiredFramesPerSecond();
-        double elapsed = g_Values.AppTime.CurrentTime() - frameStartTime;
-        if (elapsed < minimumFrameTime)
-            g_Values.FreeDrawTime = std::clamp(minimumFrameTime - elapsed, 0.0, 1.0);
+        const double fpsRaw = static_cast<double>(g_ptrSystem->EffectManager().GetCurrentEffect().DesiredFramesPerSecond());
+        // If FPS is invalid (<= 0 or non-finite), treat as unlimited (0s minimum frame time).
+        const double minimumFrameTime = (!std::isfinite(fpsRaw) || fpsRaw <= 0.0) ? 0.0 : (1.0 / fpsRaw);
```
rbergen, line 163: I don't think there is any way on our platform where a `size_t` (i.e. unsigned int) value could overflow the numerical range of `double`?

rbergen (approved):
Nice catch! I remember I did consider this when implementing the JSON reader/writer and network request scheduler. 

One nit-picky comment, but LGTM.

PR #761, opened 2025-08-25
Add effect and effect set hashing for effect serialization and deserialization, as well as effect set versioning

rbergen:
## Note

The changes in this PR have not yet been tested - I'm opening this PR to confirm the NightDriverStrip main contributors are happy with where this is going.

## Description

This pull request, which very much builds on the work done by @davepl in #755, introduces several significant refactorings and enhancements to how effects are identified, managed, and (de)serialized in the codebase. The most important changes include the replacement of the `EffectId` enum with a more flexible `uint32_t` hash, updates to effect factory and manager classes to support effect IDs and hashing, and the standardization of effect class inheritance for improved type safety and maintainability.

Again, much of the work that is mentioned next was implemented first in #755, to the extent it concerns implementing effect hash-based IDs. I have reimplemented it to isolate it from coding style and white space changes, and fix some previously uncovered cases - like effects that should have their own effect ID hashes but would share it with other effects because they derived from them.

### Effect ID Refactoring and Standardization

* Replaced the hardcoded `EffectId` enum in `effects.h` with a `using EffectId = uint32_t;` typedef, removing the need to maintain a fixed list of effect numbers and allowing for more flexible, hash-based, effect identification.
* Updated all matrix effect classes to inherit from `EffectWithId<EffectClass>` either directly or indirectly - which adds the generated hash-based effect ID to the effect's attributes - instead of using specific enum values, improving type safety and reducing coupling to numeric IDs.

### Effect Factory Enhancements

* Refactored `EffectFactories` to use `EffectId` as the key type for effect factories, added support for optional factory IDs, and introduced a hash string for configuration integrity. Also added methods for retrieving factory IDs and managing the hash string.

### Effect Manager Serialization Changes

* Changed the effect set version tracking in `EffectManager` from an integer to a hash string, improving compatibility and robustness when effect definitions change. Updated serialization/deserialization logic to use the hash strings.
* Renamed the method for loading effects from `LoadJSONAndMissingEffects` to `LoadJSONEffects` for clarity and to reflect its current behavior - deserialization either works in which case effects are deserialized from JSON, or the default effect set is loaded instead. 

These changes collectively modernize the effect management system, making it more flexible, maintainable, and robust against future changes in effect definitions.

## Contributing requirements

### DIFF
--- a/include/effects/strip/stareffect.h
+++ b/include/effects/strip/stareffect.h
@@ -495,10 +495,10 @@ class StarEffectBase : public EffectWithId<StarEffectBase<StarType, TEffect>>
 };
 
 template <typename StarType>
-class StarryNightEffect : public StarEffectBase<StarType, StarryNightEffect<StarType>>
+class StarEffect : public StarEffectBase<StarType, StarEffect<StarType>>
 {
   public:
-    using StarEffectBase<StarType, StarryNightEffect<StarType>>::StarEffectBase;
+    using StarEffectBase<StarType, StarEffect<StarType>>::StarEffectBase;
 };
 
 template <typename StarType> class BlurStarEffect : public StarEffectBase<StarType, BlurStarEffect<StarType>>
@@ -517,7 +517,7 @@ template <typename StarType> class BlurStarEffect : public StarEffectBase<StarTy
 
     virtual void Clear()
     {
-        StarryNightEffect<StarType>::setAll(32,0,0);
+        StarEffect<StarType>::setAll(32,0,0);
     }
 };
 
--- a/src/effects.cpp
+++ b/src/effects.cpp
@@ -181,7 +181,7 @@ void LoadEffectFactories()
             Effect<FireEffect>("Medium Fire", NUM_LEDS, 1, 3, 100, 3, 4, true, true),
             Effect<BouncingBallEffect>(3, true, true, 1),
             Effect<MeteorEffect>(4, 4, 10, 2.0, 2.0),
-            Effect<StarryNightEffect<QuietStar>>("Rainbow Twinkle Stars", RainbowColors_p, kStarryNightProbability, 1, LINEARBLEND, 2.0, 0.0, kStarryNightMusicFactor),
+            Effect<StarEffect<QuietStar>>("Rainbow Twinkle Stars", RainbowColors_p, kStarEffectProbability, 1, LINEARBLEND, 2.0, 0.0, kStarEffectMusicFactor),
             Effect<PaletteEffect>(RainbowColors_p)
         );
     #endif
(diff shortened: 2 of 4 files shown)


include/effects/strip/paletteeffect.h
```diff
@@ -129,18 +130,18 @@ class PaletteEffect : public EffectWithId<idStripPalette>
 
         if (_gapSize == 0)
         {
-          for (int i = 0; i < _cLEDs; i+=_lightSize)
+          for (int i = 0; i < LEDStripEffect::_cLEDs; i+=_lightSize)
           {
             iColor = fmodf(iColor + _density, 256);
-            setPixelsOnAllChannels(i, _lightSize, ColorFromPalette(_palette, iColor, 255 * _brightness, _blend), false);
+            LEDStripEffect::setPixelsOnAllChannels(i, _lightSize, ColorFromPalette(_palette, iColor, 255 * _brightness, _blend), false);
```
davepl, line 136: Why is the class name required here?

src/effects.cpp
```diff
@@ -1,626 +1,607 @@
...
+            Effect<ColorFillEffect>("Solid White", CRGB::White, 1),
+            Effect<FireFanEffect>(HeatColors_p,      NUM_LEDS, 1, 2.5, 200, 2, 15, Sequential, true, false),
+            Effect<FireFanEffect>(GreenHeatColors_p, NUM_LEDS, 1, 2.5, 200, 2, 15, Sequential, true, false),
+            Effect<FireFanEffect>(BlueHeatColors_p,  NUM_LEDS, 1, 2.5, 200, 2, 15, Sequential, true, false),
+            Effect<FireFanEffect>(RainbowColors_p,   NUM_LEDS, 1, 2.5, 200, 2, 15, Sequential, true, false),
+            Effect<FireFanEffect>(HeatColors_p,      NUM_LEDS, 1, 2.5, 200, 2, 15, Sequential, true, false, true),
+            Effect<RainbowFillEffect>(120, 0),
+            Effect<PaletteEffect>(RainbowColors_p, 4, 0.1, 0.0, 1.0, 0.0),
+            Effect<BouncingBallEffect>(3, true, true, 8),
+            Effect<StarryNightEffect<MusicStar>>("Rainbow Twinkle Stars", RainbowColors_p, kStarryNightProbability, 1, LINEARBLEND, 0.0, 0.0, kStarryNightMusicFactor),
```
davepl, line 523: It might be a win to "Rename Symbol" of StarryNightEffect to just StarEffect<>....

davepl (approved):
Looks good!

include/effects/strip/paletteeffect.h
rbergen: Short answer: because it won't compile without it. :) 

More meaningful version: because it's a _class template_ inheriting from another class template, instead of an actual _class_ doing the same, the compiler requires that any references to functions of superclasses are made explicit. 

There are other ways to do it (`this->` instead of the class name qualifier, or pulling it in with `using`), but I figured it didn't hurt to be specific here.

include/effects/strip/stareffect.h
```diff
@@ -548,31 +480,38 @@ template <typename StarType> class StarryNightEffect : public EffectWithId<idStr
         }
         else
         {
```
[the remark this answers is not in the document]
rbergen: See previous comment.

src/effects.cpp
rbergen: That's entirely your call. It's been called StarryNightEffect forever, and I was thinking that maybe you were attached to the name.

But, happy to change it if that's what you want.

rbergen:
> Looks good!

Thanks! I will test the thing with Mesmerizer and Spectrum then, and wait for @robertlipe's review as well.

rbergen:
> The only real question I have is if compatibility with existing saved effects tunings works. Maybe it's handled in some way I just don't see. But it looks like this will break existing user configurations if they've tuned their effects settings.

You're correct, a new default effect set for a project will invalidate the existing persisted settings for that effect set. Which goes against a point I've made in the past to some extent (although that also concerned device settings, which are not impacted by this).

The reasons are:

- After years of effect configurability being a capability that's available, the number of configurable effect settings is still (very) small, compared to the effect properties that are potentially configurable. I.e.: the impact of losing these tunings is currently still relatively small.
- The most regular customer-like hardware setup for this project is Mesmerizer, which is only available in kit form and seems to logically attract a somewhat tech/twiddle-savvy audience, anyway.
- Over time, I've heard far more complaints about stale persisted effect settings than effect tunings disappearing when an effect set version got bumped.
- As mentioned, the device settings are not impacted by this, which is by far the bulk of settings that exist, and people care about.

I think how big of a problem/nuisance this will be, will partly depend on how often the default effect sets get changed in the future. If it does become a problem, I already have ideas on how to migrate settings from effects that haven't changed in the default effect set, to the new ones. As it stands, I prefer only adding that complexity if it's proven to be necessary.

PR #788, opened 2025-12-04
Dev QOL fix: don't overcompile TJpg_decoder, QRCode, and UrlEncode

robertlipe:
## Description

Don't fetch/compile 50 copies of libs when 2 are needed for SmartMatrix, for example, which us used by only mesmerizer and mesmerizer_lolin. Fetching, installing, and building 50 copies of a each library for things where they aren't used isn't a great use of developer resources.

This change moves the TJpg_Decoder, QRCode, and UrlEncode library dependency from global to
specifically the a new configuration of libraries for "graphics" builds, which correlates to USE_MATRIX or EFFECTS_MATRIX or other 2D things.

Save a (very tiny) tree.

## Contributing requirements
Negative bits per blinken.

### DIFF
--- a/src/effects.cpp
+++ b/src/effects.cpp
@@ -56,7 +56,7 @@
     #include "effects/strip/faneffects.h"       // Fan-based effects
 #endif
 
-#if EFFECTS_MATRIX
+#if EFFECTS_FULLMATRIX
 #include <TJpg_Decoder.h>
 #endif // EFFECTS_MATRIX
 


rbergen:
It's very simple: with the decoder initialization code being where your PR aims to put it, non-HUB75 matrix configurations that try to show JPEGs may build but they will not work, because the callback that does the actual drawing of pixels on the matrix is never set. 

That code really needs to be pulled out of the splash screen initialization function, and executed for all `EFFECTS_FULLMATRIX` configurations, without further conditions. 

I'd say: leave it where it is (in main.cpp) and wrap an `#if EFFECTS_FULLMATRIX` directive around it, or put it in a more effects-focused place (effects.cpp, or effectsmanager.cpp if you really prefer) that is executed for all `EFFECTS_FULLMATRIX` configurations as well.

robertlipe:
Oh, and the license on the font file is OK because we already have numerous dependencies on Adafruit libraries like gfxfont.h, so the licensing and compliance are the same.

src/effects.cpp
```diff
@@ -56,6 +56,10 @@
     #include "effects/strip/faneffects.h"       // Fan-based effects
 #endif
 
+#if EFFECTS_MATRIX
+#include <TJpg_Decoder.h>
+#endif // EFFECTS_MATRIX
```
rbergen: I think this should be #if EFFECTS_**FULL**MATRIX? EFFECTS_MATRIX is set nowhere in the codebase, as far as I can tell.

Which raises an interesting question, being how this compiles for configurations that do depend on TJpg_Decoder.h - the answer probably being that it's indirectly pulled in via another file (my guess is an effect) already.

rbergen (changes requested):
Thank you for the follow-up. There are two conceptually smaller things I came across, discussed in three comments.

PR #792, opened 2025-12-16
Refactor: Modernize MeteorEffect (SoA -> AoS)

### LINKED ISSUE #791: Color Meteor effect does not use first LED in strip
I set up the EFFECTS_ATOMLIGHT configuration and noticed the color meteor effect was not using the first LED in the strip (LED 0).

Making the following changes locally fixed it, but I do not have permissions to create a branch.

`meteoreffect.h`

`if (iPos[i] < meteorSize - 1)
 {
    bLeft[i] = false;
    iPos[i] = meteorSize - 1;
}`

`if ((x <= static_cast<int>(pGFX->GetLEDCount())) && (x >= 0))`

robertlipe:
Refactors MeteorEffect to use AoS (struct Meteor) instead of SoA. Modernizes Draw loop, uses std::clamp, and fixes Pixel 0 drawing.

Fixes #791.

### DIFF
--- a/include/effects/strip/meteoreffect.h
+++ b/include/effects/strip/meteoreffect.h
@@ -38,78 +38,78 @@
 class MeteorChannel
 {
   private:
+    struct Meteor {
+        float hue;
+        float pos;
+        float speed;
+        bool  movingLeft;
+    };
 
-    std::vector<float> hue;
-    std::vector<float> iPos;
-    std::vector<bool>  bLeft;
-    std::vector<float> speed;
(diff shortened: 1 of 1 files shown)


Cyborg-Squirrel:
I built and ran the color meteor effect from this branch. Issue https://github.com/PlummersSoftwareLLC/NightDriverStrip/issues/791 is resolved. All LEDs on the strip were used. You may want to update the block comment at the top of the file.

PR #803, opened 2025-12-31
Fix: Smoke/Noise matrix scaling asymmetry

robertlipe:
Fixes math in FillGetNoise causing discontinuities on non-square matrices. This was the cause of ugliness in 'smoke' on the horizontal and vertical planes hwen run on a tiny mesmerizer screen, but not on a 64x64 that I used to fix this issue momths ago.  This includes a targeted fix for the "Smoke" matrix effect and improvements to the underlying noise engine:

### 1. Smoke Effect Artifacts
- Addressed vertical discontinuities ("bars") in the Smoke effect by tuning the `MoveFractionalNoise` smearing parameters (reduced from 3 to 1).
- This restores the intended smooth, drifting aesthetic, removes contentious 'wipers'

### 2. GFXBase Precision
- Upgraded fractional noise calculations in `gfxbase.cpp` to use strict 32-bit integer arithmetic (`int32_t`/`uint32_t`).
- This prevents potential overflow issues and rendering glitches on larger matrices, such as if we'd run this on an LCD.
- Added explicit template instantiations for `FillGetNoise` to ensure linker satisfaction across all build configurations. This allows removal of one slightly large, confusing block that was completely redundant.

- [x] 

     - [x] I read the contribution guidelines in [CONTRIBUTING.md](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
     - [x] I understand the BlinkenPerBit metric and maximized it in this PR.
     - [x] I selected main as the target branch.
     - [x] All code herein is subject to the license terms in [COPYING.txt](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

### DIFF
--- a/include/gfxbase.h
+++ b/include/gfxbase.h
@@ -1382,7 +1382,7 @@ public:
         // function templates) are included in gfxbase.cpp, because of the way C++ demands things to be
         // structured.
         //
-        // The default approach for the tenplated functions is determined by the value of _defaultNoiseApproach,
+        // The default approach for the templated functions is determined by the value of _defaultNoiseApproach,
         // which is defined earlier in this class.
         template<NoiseApproach = _defaultNoiseApproach>
         void MoveFractionalNoiseX(uint8_t amt, uint8_t shift = 0);


rbergen (commented):
I'm seeing three changes:

- A switch from `MATRIX_WIDTH`/`HEIGHT` to `_width`/`_height`. Which may be logical from a conceptual standpoint, but:
  - does make the noise function implementations for `NoiseApproach::Two` inconsistent (i.a., other noise functions for this approach do still use `MATRIX_WIDTH`/`HEIGHT`). 
  - doesn't change anything in the practical sense, because for non-hexagon configurations, `_width` is initialized to `MATRIX_WIDTH` and `_height` to `MATRIX_HEIGHT`.
- A switch from 8 bit to 16 bit loop counters, which should only matter if width and/or height are larger than 255.
- A replacement of `CENTER_X_MINOR` and `CENTER_Y_MINOR` with slightly different calculations based on width and height respectively. 

Is the third change then the actual fix?

rbergen:
I think we can officially call this a collaborative PR by now. :) Maybe the first to this extent?

In any case, thank you for picking this one up in the first place. Based on the exchange on the related issue, you weren't the only one bothered by this.

PR #806, opened 2026-01-01
CLI: Add fuzzy effect selection, tab completion, and brightness nudges

robertlipe:
- Enhanced effect command: Now supports next, prev, numeric indices, and partial case-insensitive name matches.
- Added + and - commands: Bumps brightness up/down by 10 steps.
- Context-Aware Tab Completion(!): Typing effect Sm<TAB> completes to effect Smoke. If multiple matches exist (like RadialFire and RadialWave), rad<TAB> completes the common prefix radial.
- Refined Matching: Replaced strncasecmp with project-consistent StartsWithInsensitive and ContainsInsensitive helpers.

    - [x] I read the contribution guidelines in [CONTRIBUTING.md](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
    - [x] I understand the BlinkenPerBit metric and maximized it in this PR.
    - [x] I selected main as the target branch.
    - [x] All code herein is subject to the license terms in [COPYING.txt](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

### DIFF
--- a/src/debug_cli.cpp
+++ b/src/debug_cli.cpp
@@ -631,10 +631,6 @@ void ProcessCLIByte(uint8_t byte)
             cmd.pop_back();
             cli_printf("\b \b");
         }
-        else
-        {
-            // Optional: Ring bell or ignore if buffer empty
-        }
         break;
 
     case '\r':
@@ -645,7 +641,6 @@ void ProcessCLIByte(uint8_t byte)
         {
             // If buffer was empty (just Enter), RunCommand("") handles the prompt
             RunCommand("");
-            cmd.clear();
         }
         else
         {


src/debug_cli.cpp
```diff
@@ -120,6 +120,24 @@ bool StringCompareInsensitive(std::string_view sv, const char *in_s)
     return *s == '\0';
 }
 
+bool StringStartsWithInsensitive(std::string_view haystack, std::string_view needle)
```
rbergen: Ha!

rbergen (commented):
Nice additions. Couple of comments, basically related to managing the expanding scope of the capability('s code).

src/main.cpp
rbergen: I may well be speaking before my turn as you haven't asked for a new review yet. Still, just for clarification: my suggestion was, and is, to turn the whole lambda (i.e. the entire byte handler) into a function that moves to one of the debug_cli.* files. 

I think the fact that bytes come in one by one is basically an understandable trait of serial traffic - which is why Improv also handles them that way -, so creating a function `process_cli_byte()` (or something) with the exact signature of the lambda is still generic enough to be defendable.

robertlipe:
PTAL

Oh, THAT lambda. Also agreed. I split up the other one, but I see now what wasn't what you meant. I think that was in good taste, too, so I'm keeping it.


This should have gone in the original description, but this is another thing we now have that is ridiculously awesome, even if I'm the only one that uses it.

1. Numbers: "effect 3" -> switches to exact index.
2. (Quoted) Words: "effect Smoke" -> switches to "Smoke".
3. (Fuzzy) words: "effect smo" -> switches to "Smoke". (if only one match).
4. (Ambiguous Fuzzy) Words: "effect s" ->  lists candidates (e.g., "Smoke", "Sparks", "Snow"), waits for completion.                                                               
5. (Nonsense Words) Not Found: effect xyz -> prints "No effect matching 'xyz' found."
6. (Nonsense Numbers) Range: effect 999 -> Should print "Error: Effect index 999 out of range."

Every block of code should try this hard to understand what the user wants. (Every code should have me not feeling well and pawing at things that make sense and make me mad when it JUST DOESN'T UNDERSTAND ME! :-) )

robertlipe:
Wait, where _is_ ProcessCLIByte()?  Maybe THATs why you sound grumpy. I seem to have lost a commit. I see it in my command history. 

Maybe I put it on the wrong branch or something... Stand by. (Well, looking at the clock, don't stand by.)

rbergen:
It's funny that you mention #811, because that PR's hiccup made me realize that this is exactly what may have happened. From my perspective this was the best-case explanation, and thank you for setting it straight. I'll go over the changes one more time, but I now expect an LGTM and merge very soon.

src/debug_cli.cpp
```diff
@@ -437,7 +601,71 @@ void cli_printf(const char *fmt, ...)
...
+    case 0x7f:
+        if (!cmd.empty())
+        {
+            cmd.pop_back();
+            cli_printf("\b \b");
+        }
+        else
+        {
+            // Optional: Ring bell or ignore if buffer empty
+        }
```
rbergen: I think we've made a choice here, so we can remove the else block?

PR #808, opened 2026-01-01
Standardize format specifiers for GCC 14/Arduino3 compatibility

robertlipe:
# PR Description: GCC 14 Format Specifier Standardization

_This is a hybrid of human and machine (Gemini, CLion, Clang-tidy, GCC) work. I've had some variation of this in my local trees now for about a year and a half and this PR started from that base. This version works on both G++14 and GCC8. I wrote a special wrapper that makes a compile_commands.json, grabs the flags out of it, uses G++14 instead of G++8, and then adds "-fsyntax-only" and, to be sure, "-Wformat=2," so I could at least trivially parse the source on the two different compilers at the same time. After that worked, I had Gemini build a "Rules of the Road" based on those numerous changes to record how passing should be done and then apply those rules to everything *outside* the ifdef brackets we can exercise. I let Gemini write this description because I was so disgusted with the final result._  

This PR establishes and applies a standardized set of "Rules of the Road" for `printf`-style format specifiers across the core codebase. This is a critical preparation step for the upcoming migration to GCC 14 and the enforcement of the `-Wformat=2` warning flag.

## Technical Rationale: Why so many casts?

In earlier GCC versions, the compiler was often lenient if you passed a type that was bit-compatible with the specifier (e.g., passing a 32-bit `unsigned int` to a `%lu` which expects a 32-bit `unsigned long`). GCC 14 treats these as distinct type identities.

### 1. Type Identity Consistency
On the ESP32 toolchains, `uint32_t` is typically an alias for `unsigned long`. However, another toolchain might define it as `unsigned int`. By explicitly casting to `(unsigned long)` and using `%lu`, we:
- **Guarantee a match** across all ESP32 variants (S2, S3, C3, Classic).
- **Prevent environment-specific warnings** that occur when toolchains disagree on type aliases.

### 2. Standardizing `size_t`
The most frequent mismatch was `size_t` being logged with `%d`. This often works solely due to luck (stack layout) and fails on 64-bit or stricter 32-bit audits. We have standardized on `%zu` with an explicit `(size_t)` cast.

### 3. "Rules of the Road" Applied
- **`uint32_t`** -> `%lu` with `(unsigned long)` cast.
- **`int32_t`** -> `%ld` with `(long)` cast.
- **`size_t`** -> `%zu` with `(size_t)` cast.
- **`CRGB` (Hex Colors)** -> `%08lX` with `(unsigned long)` cast.
- **Small Integers** (`uint8_t`, `int16_t`, etc.) -> `%d` with `(int)` cast.

## Impact
While this adds verbosity ("casts") to the logging calls, it provides the "Type Safety contract" necessary to keep the project build warning-clean. A warning-clean build is essential for catching real logic bugs in a project with this many build environments.

## Files Included (20)
- Core: `main.cpp`, `systemcontainer.h`, `types.h`, `taskmgr.h`, `ledbuffer.h`
- Drivers/Graphics: `screen.cpp`, `hub75gfx.cpp`, `LV_Helper.cpp`, `audio.cpp`
- Comms: `network.cpp`, `socketserver.cpp`, `webserver.cpp`, `jsonserializer.cpp`, `improvserial.h`
- Effects/Utils: `remotecontrol.cpp`, `ntptimeclient.cpp`, `debug_cli.cpp`, `effectmanager.cpp`
- Effect Headers: `PatternAnimatedGIF.h`, `particles.h`

---

## Future Path: Why not `std::print`?

While the amount of explicit casting is admittedly "noisy," it is currently the most flash-efficient way to achieve strict type safety. 

We acknowledge that C++23's `std::print` and `std::format` provide a much cleaner syntax that handles these type identities automatically. However, even in GCC 14, the standard library implementation of these features often brings in significant overhead (locales, complex transitives) that can "blow out" flash size on constrained ESP32 targets.

Until the toolchain's support for `std::print` matures to be as lean as `printf`, these explicit "Type Safety contracts" are the necessary evil to keep the build both warning-clean and small enough to fit.

[X] I read the contribution guidelines in [CONTRIBUTING.md](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
[X] I understand the BlinkenPerBit metric, and maximized it in this PR.
[X] I selected main as the target branch.
[X] All code herein is subjected to the license terms in [COPYING.txt](https://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

### DIFF
--- a/include/effects/matrix/PatternAnimatedGIF.h
+++ b/include/effects/matrix/PatternAnimatedGIF.h
@@ -292,7 +292,7 @@ public:
         int offsetY = (MATRIX_HEIGHT - dstHeight) / 2;
 
         debugI("GIF scaling: %dx%d -> %dx%d (scale %.2f,%.2f) offset (%d,%d)",
-               gifWidth, gifHeight, dstWidth, dstHeight, scaleX, scaleY, offsetX, offsetY);
+               (int)gifWidth, (int)gifHeight, (int)dstWidth, (int)dstHeight, scaleX, scaleY, (int)offsetX, (int)offsetY);
 
         g_gifDecoderState._offsetX   = offsetX;
         g_gifDecoderState._offsetY   = offsetY;
--- a/include/effects/strip/particles.h
+++ b/include/effects/strip/particles.h
@@ -310,7 +310,7 @@ template <typename Type = DrawableParticle> class ParticleSystem
 
     virtual void Render(const std::vector<std::shared_ptr<GFXBase>>& _gfx)
     {
-        debugV("ParticleSystemEffect::Draw for %d particles", _allParticles.size());
+        debugV("ParticleSystemEffect::Draw for %zu particles", (size_t)_allParticles.size());
 
         while (_allParticles.size() > 0 && _allParticles.front().Age() >= _allParticles.front().TotalLifetime())
             _allParticles.pop_front();
--- a/include/improvserial.h
+++ b/include/improvserial.h
@@ -339,7 +339,7 @@ protected:
                         log_write(".Sending details for SSID %s", WiFi.SSID(i).c_str());
                         // Send each ssid separately to avoid overflowing the buffer
                         std::vector<uint8_t> data = improv::build_rpc_response(
-                            improv::GET_WIFI_NETWORKS, {WiFi.SSID(i), str_sprintf("%d", WiFi.RSSI(i)), WiFi.encryptionType(i) != WIFI_AUTH_OPEN ? "YES" : "NO"}, false);
+                            improv::GET_WIFI_NETWORKS, {WiFi.SSID(i), str_sprintf("%ld", (long)WiFi.RSSI(i)), WiFi.encryptionType(i) != WIFI_AUTH_OPEN ? "YES" : "NO"}, false);
                         this->send_response_(data);
                     }
                 }
--- a/include/ledbuffer.h
+++ b/include/ledbuffer.h
@@ -122,7 +122,7 @@ class LEDBuffer
 
         if (payloadLength < length32 * sizeof(CRGB) + cbHeader)
         {
-            debugW("command16: %d   length32: %d,  payloadLength: %d\n", command16, length32, payloadLength);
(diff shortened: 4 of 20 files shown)


robertlipe:
I asked copilot for a code review. We had a detailed chat where it suggested things like using the PRI_ macros (which I hate) and suggested using CI to test the g++14 code path (intrigued, but once landed into my Arduino3 branch, a bit redundant) ... and then crashed. It then told me it had no recovery. Then it tells me that it can't post an LGTM, but will give me text saying that it gave me an LGTM (no trust issues here...).  My experience with Copilot was not a great one. Still, it agrees there is nothing sneaky here.


LGTM.

Summary & context:
- This PR standardizes printf-style format specifiers across the codebase as a preparation for GCC 14 and enforcement of -Wformat=2.
- The PR author used a combined human+tooling process (compile_commands, GCC/G++ variations, and automated rule application) to locate and fix mismatches; changes are limited to format specifier casts and matching format strings.
- The approach standardizes casts (e.g., uint32_t -> (unsigned long) with %lu, int32_t -> (long) with %ld, size_t -> (size_t) with %zu, etc.) to ensure type-identity correctness across toolchains and to keep builds warning-clean without incurring heavy std::format/std::print overhead on constrained targets.

What I reviewed and confirm:
- I reviewed the PR description and the changes in this branch.
- I confirm the changes are consistent with the description: the PR does not modify files or behavior outside the stated scope of standardizing format specifiers and adding explicit casts.
- I carefully considered the design choices and potential side effects described in the conversation and did not find any unintended modifications or regressions introduced by these commits.

Why this is acceptable:
- The additional casts are verbose but intentionally provide a "type safety contract" that prevents environment-specific warnings and helps keep the project warning-clean, which is important across multiple build targets and toolchains.
- Given the constraints of ESP32 flash size and current standard library implementations, this approach is reasonable until a lean std::print/std::format path is viable.

Conclusion:
- Approving this as-is. Thanks for the clear description and for addressing the points raised.

PR #809, opened 2026-01-01
Python Client: Layout mapping and spectrum display fix

robertlipe:
## Description

This PR improves the Python client's live view by adding support for different physical LED layouts.

### Changes:
- **Noisy Log Fix**: Moved the 'Peak brightness' print behind the --verbose flag.
- **Layout Mapping**: Added a --mapping flag to support row-major, serpentine (column-major boustrophedon), and spectrum (tiled 16x16) layouts.
- **Auto-Detection**: The client now automatically detects 48x16 grids and applies the spectrum mapping, fixing the scrambled display on spectrum devices.
- **Legacy Compatibility**: Defaults to row-major for all other detected sizes unless overridden.

### DIFF
--- a/tools/nightdriver_client.py
+++ b/tools/nightdriver_client.py
@@ -491,7 +491,7 @@ def save_raw_frames(frames, output_filename, verbose=False):
 
     if verbose: print(f"save_raw_frames: Saved raw frames to {output_filename}")
 
-def live_view(host, layout="flat", verbose=False, gain=1.0, scale=None):
+def live_view(host, layout="flat", verbose=False, gain=1.0, scale=None, mapping="row-major"):
     """
     Displays a live, real-time view of the device output.
     """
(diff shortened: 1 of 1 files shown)


robertlipe:
Thanx.. that's not actually a _bad_ trait in a swe. 😉

Once I (ok,.we - Gemini helped) had the math worked out, I was happy that
it generalized well to more common cases. I started thinking about only
spectrum, but it was wieth bringing it out to an option. As you could tell,
I'm increasingly happy with the zany combinations of things we're able to
"wire" together without wires.

For example, if I ever get it together and clean up and submit my hexagon
gfx layer (now that is also gnarly math) and you wanted to see it run
before merging it (far from unreasonable) but don't have the obscure
hardware it takes, you can now use live view. We could build more
configurations into the preview pretty easily.

Want to debug beat detection or audio in a silent room? Turn it on fake
music at runtime. Want to change the number of LEDs or (soon, I hope, the
pin numbers)? Ditto.

Like a beaver, I'm all about improving my own (our own) environment to make
development easier.

Thanks for spending your New Years day on this. I know it looked like a DOS
attack on the review queue, but I've been pulling out some of these things
I've been sitting on for a while and making them generally a available.
This one, specifically, was a newly found need to fix.

robertlipe:
Indeed. I was thinking just yesterday that I should probably put together a
"what's new in ND"  post explaining some of this stuff so it's as useful to
others as it is to me. I've not really "pitched" them in the submits and
they really deserve some.

I'd like to also submit and AGENTS.md that explains this to machines. I
have one that enforces those printf specifiers, for example .

Letting an agent "see" the display via telnet is incredibly powerful for
example. You've seen that I have a WiFi AP mode in the wings; I developed
that very much by coaching Gemini add DebugA prints and watching real world
state transitions and add a few new disposable commands to change them on
demand. It was pretty swanky for embedded!

I'll see if that R guy can cook something up.


RJL

PR #811, opened 2026-01-05
Replace UrlEncode library with local implementation.

robertlipe:
## Description
Removed `UrlEncode` third-party dependency.
* Add `include/UrlEncode.h` proto and new (better) body in `src/network.cpp`.
Uses  RFC 1738 style percent-encoding.
* Guarded the `urlEncode` function and its usage in `src/deviceconfig.cpp` with `#if ENABLE_WIFI`. It may be possible to do better later, but we agree for now that without WIFI, there is no URL.
* Removed unused includes of `UrlEncode.h` from `PatternStocks.h` and `PatternSubscribers.h`.

## Contributing requirements

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -117,11 +117,11 @@
     #endif
 #endif
 
-#define XSTR(x) STR(x)              // The defs will generate the stringized version of it
+#define XSTR(x) ND_STR(x)           // The defs will generate the stringized version of it
 #if FLASH_VERSION > 99
-    #define STR(x) "v"#x
+    #define ND_STR(x) "v"#x
 #else
-    #define STR(x) "v0"#x
+    #define ND_STR(x) "v0"#x
 #endif
 #define FLASH_VERSION_NAME_X(x) "v"#x
 #define FLASH_VERSION_NAME XSTR(FLASH_VERSION)


include/UrlEncode.h
```diff
@@ -0,0 +1,34 @@
...
+//   Helper to URL-encode strings for API calls.
+//   Implemenation provides RFC 1738 style percent-encoding.
+//
+//---------------------------------------------------------------------------
+
+#pragma once
+
+#include <Arduino.h>
+
+String urlEncode(const String& str);
```
rbergen: I suggest we add a comment to indicate where the definition lives, that not being UrlEncode.cpp. 
Which is something I was actually going to challenge, but I can see the logic after thinking it through from a few different angles.

robertlipe:
Yes. Distracted with other projects... Some of which are ND, but in
different trees.

The rationale was to minimize and localize the change. That's why I
kept/made the header.  We have a ton of audio/networking code that isn't
"correctly" conditionalized so I didn't go on a recursive chase for that. I
minimized the blast radius.

But I understood that's what it takes, so I'll come back to this.


As for the "yeah, right", thing, I probably lead you to that. My distaste
of the platformio approach is not exactly a secret. I have another, more
extreme version of this that brings in ALL (?) the libs (and has a python
thingy that checks them out/updates them).

That lets us actually FIX some of the abandoned and misguided code we're
depending on and reduces our dependencies on Platformio's buggy, slow build
system, which is two things that I'm not a fan of in any software.

It makes builds less slow because we're not waiting for 200+network
fetches, making 50 copies of everything, and filling our disks with
redundant copies. Trying to get patches accepted upstream into abandoned
projects just isn't working and I really don't want to become a general
purpose fork of those packages; I just want them integrated sensibly.

This one was just easier to cold replace. Im testing the waters. 😎

But, like so many things in my ND world right now, it's piled up behind
other things.


The Arduino3/P4 thing may depend on it...

I actually took a break from that very recently and developed two new fun
effects from scratch. I have a few more in the idea incubator, too.

I also have a note at home that there was something else I recently
half-proposed that you'd pre-approved in principle, so it's on my near-term
list, too. Oh, and some doc.

I've not forgotten about the project. Far from it.

PR #824, opened 2026-03-08
Refactor: Replace RemoteDebug with custom Logger and Telnet server

robertlipe:
### Description

This PR replaces the `RemoteDebug` library with a custom-built, lightweight logging and console management system. While `RemoteDebug` served us well, it brought along unnecessary complexity and dependencies that didn't perfectly align with the project's long-term architectural goals.

### Key Improvements

1.  **Lightweight Infrastructure**: Replaced a heavy third-party library with a tailored implementation (`ConsoleManager`, `ConsoleSession`, `Logger`) that fits the project's specific needs.
2.  **Global Logging**: The new `Logger` hooks directly into the ESP-IDF `esp_log` system. This means that logs from the system, third-party libraries, and our own `debugX` macros are all consistently captured and broadcast.
3.  **RFC 854 Telnet Server**: Implemented a solid, BSD-socket based Telnet server. It properly handles Telnet negotiation (WILL ECHO for server-side echo, WILL SGA for character-at-a-time mode) and RFC 854 line ending conventions.
4.  **Enhanced Debug CLI**: The CLI has been rewritten to be more memory-efficient. The new tokenizer uses `std::string_view` for zero-allocation parsing, and tab completion has been improved for both commands and effect names.
5.  **Clean Multiplexing**: Output is now intelligently routed. Commands executed over Telnet stay on Telnet, while global broadcasts (like system logs) reach all active sessions.

### Technical Details

-   **Memory Efficiency**: The new system avoids unnecessary string copies and large stack buffers. The `vprintf` hook uses a two-pass approach to measure and then allocate exactly the required heap memory for log lines.
-   **Concurrency**: Uses `std::recursive_mutex` in `ConsoleManager` to ensure thread-safe access to sinks during broadcasts and session management.
-   **Standard Compliance**: The Telnet server includes a state machine to handle IAC sequences and RFC 854 `\r\0` / `\r\n` sequences correctly, even when split across packet boundaries.

### Validation

Verified on hardware:
-   Serial console remains fully functional with Improv support.
-   Telnet server accepts connections, handles multiple commands, and supports tab completion.
-   System logs are correctly interleaved with CLI output.
-   ANSI color coding works correctly in modern terminal emulators.


include/nd_network.h
```diff
@@ -29,49 +29,50 @@
...
+#include <WString.h>
+
+String GetMacAddress();
+String GetMacAddressPretty();
+bool GetWiFiHostByName(const char* hostname, IPAddress& ip);
+String GetWiFiLocalIP();
+int GetWiFiRSSI();
+int GetWiFiStatus();
+void SetWiFiModeSTA();
+const char* WLtoString(int status);
```
rbergen: Same comment about #if ENABLE_WIFI wrappers, with the added question if (some of) these should maybe be combined with the other declarations. Phrased differently: why some higher up in the file, and some here?

src/logger.cpp
rbergen: the vprintf hook is global, which can introduce global concurrency issues (the deadlock kind of fun). I know you've tested these changes, but have you done so under "logging load"?

src/audio.cpp
rbergen: Again, audio changes in a debug PR. Please explain why these are here.

rbergen (commented):
I've made a few comments, some of which I care far more about than others; I'm sure you can figure out which are which.

As for the audit scrips: those you want to run, you can add to the extra_scripts directive in platformio.ini, which is at line 338 in the updated version. To not bloat platformio.ini too much (yeah, I know...) you could write and only invoke a post-build audit utility Python script that invokes them.

rbergen:
@robertlipe In part because I see a stack of new PRs from you coming in (some of which are drafts, it seems... brrr), can we please move forward with this one?

include/nd_network.h
```diff
@@ -29,49 +29,50 @@
...
+#include "esp_mac.h"
+#include "types.h"
 
+// Prototypes that should exist regardless of ENABLE_WIFI to keep callers clean
+void GetMacAddressRaw(uint8_t *mac);
+void InitNetworkCLI();
+bool IsWiFiConnected();
+void NetworkHandlingLoopEntry(void *);
+bool SetSocketBlockingEnabled(int fd, bool blocking);
+
```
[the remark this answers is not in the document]
robertlipe: Fair point. But if we're going to clean up this garbage, we should get the other garbage elsewhere in this same file, right? IMO, good taste (and Google3 style) is that everything in our headers should be namespaced—certainly all methods in the same header.

So I'll make you an offer of an IOU. (If you want it, make a ticket and assign it to me.) 

I changed this to bracket the above into a namespace.  I also fixed all the callers, of course.

I did _not_ push everything else in tis file (UpdateNTPTime, ReadWiFiConfig, ClearWiFiConfig, etc.) into that same namespace because it just caused an explosion of creep that kept getting further and further from RemoteDebug. I don't MIND doing that slugwork if you want it, but I'm not going to jam this up yet further.  After this lands, I'll come back for that "jira ticket" if you think it's worthy. (But why stop in nd_network? Why not all our headers? And if we're doing that, is it "namesapce improv" but "namespace _effect_id_detail" or "namespace ArduinoJson"? If we're going to get all namespacey and start decreeign rules, maybe we actually have to decide what the spelling rules are. 

Heck, I couldn't even be consistent with myself - I used 'nd_network' here after the name of the file, but namespace CamelCase elsewhere after their respective classes...but here, we weren't IN a class.

THIS is why all my PRs seem to collapse into the gravity of the sun...  😄

robertlipe:
PTAL

This is my new happy place with this PR.

I know I'm headed for a merge collision with ... myself. I just moved around a bunch of stuff in nd_network.h (stuff that used to make sense, but that was before the giant IWYU and Networking patches around the holidays.) That busy work about the MAC fetcher that gets a tweak here, gets torched in the P4 world. We don't HAVE networking on P4, so we can't print a MAC address at boot. We don't even know if we have a chip yet that can load code that can _have_ a MAC.  So #836 gets a whole load of logic that allows async discovery of the MAC address as we bring up the APB, enumerate networking devices, and discover MACs post-boot. It's also increasingly clear that the existing scheme of "decl vs. body" based on if have_wifi is just not working. 

...but I can't keep recursively patching/developing these. I'm making myself crazy and I can tell that my efforts to get some of this work cleaned up and stable are making you a bit crazy (that's OK!) and I don't want to do that. I think this is the largest single piece of atomic  outstanding work and while it's veered a bit beyond "replace RemoteDebug"  I think it's actually quite good now.

Please feel free to submit a ticket and assign it to me. I'll tend to it after #836 and a few others land. I'll even make it easy for you to ask:

Title: [DEBT] Refactor nd_network  for Arduino3
Followup from #824.
1) Refactor nd_network namespace. Transition from nd_network:: getters to a structured WiFiStatus object or std::optional or something to handle ENABLE_WIFI states gracefullly.
2) Clean up identity vs. telemetry groupings.
3) Reconsider split of decls based on ENEABLE_WIFI so maybe we get rid of a ton of ifdefs and weird logic.
4) Think harder about cases where we just plain don't HAVE WiFi. (Whate if you ENABLE_WIFI on something like the WaveShare boards that don't HAVE Wifi? Made sense for ESP32-nothing, but something's gotta give way to new world orders.
5) Generally better integrate the new boot-time MAC APIs to prevent crashes when printing identity before the network coprocessor is ready. Maybe #4 being "never" is a special case of this one.

rbergen:
> ...but I can't keep recursively patching/developing these. I'm making myself crazy and I can tell that my efforts to get some of this work cleaned up and stable are making you a bit crazy (that's OK!) and I don't want to do that. 

I'm actually okay on the crazy part. But yes, I'm not a fan of six somewhat interconnected PRs from one author open at the same time - with two in draft status - exactly because I _know_ it tends to lead to merge conflicts which then delay the whole process of reviewing, fixing and merging across at least some of them.

> Please feel free to submit a ticket and assign it to me. I'll tend to it after #836 and a few others land. I'll even make it easy for you to ask:

Haha, I've opened issues for myself before, and I think what you wrote after that line was a prime candidate. I was insufficiently unhappy with the previous state of our WIFI code to change it and I am happy with merging what's in this PR as it now is, so I don't have a strong opinion about future follow-ups from a standpoint of necessity. But that said, what you wrote does look like a very good idea. :wink:

rbergen (approved):
LGTM!

robertlipe:
Thanx!

You know that most everything I've opened in the last couple of quarters Is
'sort of connected' and why. It's been pretty weird trying to keep things
separate enough to be submittable.

I've kept the 'nullptr deref' as a draft because I KNOW it will collide
hard with this one; everything else is now marked ready to go. (That wasn't
true even a few hours ago.) The SIOF PR is necessary for GCC14 (undefined
behaviour changed)  but I plan to come back and eliminate most of that
code, too. That's another task for "later," or else we can't keep pushing
the whole project forward.

Now that we've both seen that crack in the wall of network.*, though, with
or without a ticket one of us is going to have to return to it, though. :-)

PR #826, opened 2026-03-09
Move -g3 from build_flags to build_src_flags

zackees:
## Summary

- Removes `-g3` from `[base] build_flags` so the ESP32 core and all libraries no longer compile with maximum debug info
- Adds `-g3` to `build_src_flags` so sketch sources still get full macro debug info for debugging
- This significantly speeds up compilation and reduces object file bloat for release builds

Fixes #825

## Test plan

- [ ] Verify `pio run -e demo` compiles successfully
- [ ] Confirm sketch .o files still contain DWARF macro info (`readelf --debug-dump=macro`)
- [ ] Confirm library/core .o files no longer contain macro debug info

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -36,7 +36,6 @@ build_type      = release
 upload_speed    = 2000000
 monitor_speed   = 115200
 build_flags     = -std=gnu++2a
-                  -g3
                   -Ofast
                   -ffunction-sections
                   -fdata-sections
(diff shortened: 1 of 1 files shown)


rbergen:
-g3 change is rolled up in #891, addition of dev_esp32_c6 env has not been adopted as per @robertlipe's comment.

zackees:
Try out [fbuild](https://github.com/fastled/fbuild) if you want super fast compile times. It's my re-write of the platformio but in rust and your repo was used as a build benchmark. It's a drop in replacement, persists the build artifacts and doesn't poison itself. It's got a compiler cache that operates in microseconds per file.

PR #836, opened 2026-04-07
Fix:  Very Serious maintainability improvements in nd_network.h, network.cpp

robertlipe:
## Description

_This PR changed directions 2026-04-08. You may need to refresh your mental model. _

[Old] On ESP32-P4 boards using an ESP32-C6 companion chip for WiFi, querying `WiFi.macAddress()` before the driver is initialized causes a fatal `nullptr` dereference (`hosted_memcpy` crash). 
[New] That is true. But I found another API that doesn't have this problem. It turns out that _all_ ESP32s have a MAC address that's readable via the eFuse API (never mind that it's not actually set by blowing fuses...) that never changes and is suitable enough for our purposes of displaying a unique number in the debug console. But by the time we discovered that, I was deep into a cleanup that Rutger and I'd agreed upon earlier int he day.

We'd agreed upon direction (it was pretty clear) and I relied heavily on automation for the edits, but I babysat the process through probably a hundred edits and builds. Cue robot voice:

✦ This PR has evolved from a targeted bugfix for the ESP32-P4 into a foundational refactor of
  the networking subsystem. By moving to an eFuse-direct identity model and implementing a
  "Clean Facade" architectural pattern, we have eliminated boot-time indeterminacy ("crash" -- RJL) on the P4
  while significantly reducing the technical debt across the entire project.

  Summary of Changes

  1. Unified Hardware Identity (The "Real" P4 Fix)
  The original crash on the ESP32-P4 was caused by polling the MAC address via radio drivers
  before the C6 companion chip was fully initialized. This PR resolves that by:
   * Switching to esp_efuse_mac_get_default() as the single source of truth for the device's
     MAC address.
   * Bypassing the radio driver layer entirely for identity, making the MAC available at the
     very first instruction of the bootloader.
   * Consolidating 6+ redundant and inconsistent MAC functions into a single, standardized
     nd_network::GetMacAddress(const char* separator) API.

  2. Clean Facade Networking Architecture
  Eiminates the "ifdef soup" leaking into headers and caller code:
   * Implements a unified nd_network namespace that contains all networking prototypes.
   * Removed #if ENABLE_WIFI guards from nd_network.h, providing a stable API surface for the
     rest of the app regardless of compile-time flags.
   * Centralized real vs. stub implementations in network.cpp using a single #if ENABLE_WIFI / #else structure.
   * Provided global compatibility aliases (via using declarations) to maintain parity with existing logic while enforcing a cleaner namespace structure for new code. (We can remove those when we next touch them.)

  3. Maintenance & Modernization
   * Tool Compliance: Verified against project audit scripts; alphabetized system includes (IWYU), removed trailing whitespace, and ensured build parity between mesmerizer
     (ESP32-Nothing w/Radio) and tab5_demo (P4/C6) environments.

  Technical Impact
   * Stability: No observed boot-time instability or radio-dependency crashes on the P4 across ~200 reboots.
   * Readability: The networking header is now a clean API definition rather than a preprocessor (il)logic puzzle.
   * Enforcement: Legacy snake_case and "Pretty" variants have been removed, forcing the codebase to use the new, deterministic hardware identity API.

## Contributing requirements

### DIFF
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -265,7 +265,8 @@ namespace nd_network
 
         #if INCOMING_WIFI_ENABLED
             g_ptrSystem->GetSocketServer().release();
-            g_ptrSystem->GetSocketServer().begin();
+            if (false == g_ptrSystem->GetSocketServer().begin())
+                throw std::runtime_error("Could not start socket server!");
         #endif
         #if ENABLE_OTA
             SetupOTA(String(WiFi.getHostname()));
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -89,17 +89,57 @@ struct NetworkReader::ReaderEntry
 
 static DRAM_ATTR WiFiUDP l_Udp;              // UDP object used for NNTP, etc
 
+#include <atomic>
+
+static std::atomic<bool> l_bWifiDriverReady{false};
+
+static bool l_cachedMacAddress_valid = false;
+static String l_cachedMacAddress;
+
+static bool l_cachedMacAddressPretty_valid = false;
+static String l_cachedMacAddressPretty;
+
 String get_mac_address()
 {
+  if (l_cachedMacAddress_valid)
+    return l_cachedMacAddress;
+
+#if ENABLE_WIFI
+  if (l_bWifiDriverReady)
+  {
+    uint8_t mac[6];
+    WiFi.macAddress(mac);
+    l_cachedMacAddress = str_sprintf("%02x%02x%02x%02x%02x%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+    l_cachedMacAddress_valid = true;
+    return l_cachedMacAddress;
+  }
+#endif
+
+  // Fallback if driver not ready (e.g. early boot on P4)
   uint8_t mac[6];
-  WiFi.macAddress(mac);
+  get_mac_address_raw(mac);
   return str_sprintf("%02x%02x%02x%02x%02x%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
 }
 
 String get_mac_address_pretty()
 {
+  if (l_cachedMacAddressPretty_valid)
+    return l_cachedMacAddressPretty;
+
+#if ENABLE_WIFI
+  if (l_bWifiDriverReady)
+  {
+    uint8_t mac[6];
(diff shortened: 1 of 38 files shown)


robertlipe:
This IS going to collide with #824. No ways around it. Let's land that one and I'll come back to repair this one.

robertlipe:
PTAL

I'd already started running with the ideas (and generally recoiling from the disgust at looking at the current state of nd_network.h - and even accepted that I helped make that mess - when I had the Aha moment of using a differnet API. https://docs.espressif.com/projects/esp-idf/en/v6.0/esp32p4/api-reference/system/misc_system_api.html#_CPPv425esp_efuse_mac_get_defaultP7uint8_t This API doesn't "flicker" while we're booting.

By that time, I'd already implemented a whole bunch of the cleanups and fixes we spoke of earlier today.

I told myself I'd limit this change to ONLY **/*network* and I didn't quite hit that goal (if I saw bad whitespace or an out-of-order include, I fixed it, much like I'd pick up trash when walking by) but I didn't sprinkle nd_network:: around the whole tree.

Overall, I think this is a good balance of 
A) Making networking actually work on ESP32-P4.
B) Cleaning up the icky mess that we talked about earlier.

Thanks for the review. (You actually sort of already started on #824. I just struck while the iron was still hot and it was fresh in my head instead of making you copy-paste my own words to me.)

With this change applied, tab5_demo (demo, not spectrum2 - that still needs SIOF fixed - boots and runs *with microphone and screen*. Lights at the end of the tunnel!

src/network.cpp
```diff
@@ -56,1150 +59,678 @@
...
+        if (bPreviousConnection)
             return WiFiConnectResult::Connected;
-        }
 
+        bPreviousConnection = true;
+        bReportedDisconnected = false;
+
+        #if INCOMING_WIFI_ENABLED
+            g_ptrSystem->GetSocketServer().release();
+            g_ptrSystem->GetSocketServer().begin();
```
rbergen: We no longer seem to check for socket server startup failures, where we explicitly threw a runtime_error before if a failure occurred. What's the motivation for this change?

robertlipe: Incompetence? :-)

I thought that if begin() encountered a fatal error IT would commit hari-kari, but you're right. 

Obvious change brought over.

rbergen (approved):
LGTM!

PR #843, opened 2026-04-21
Reference

davepl:
## Description
This makes g() a reference instead of a vtable pointer lookup.  It also fixes the blanking during effect cross-fading.  I can't prove it, but I suspect this is much more efficient.

The bulk of the change is the one change to g() and then everywhere that calls it, which is pretty much a mechanical replacement.

### DIFF
--- a/include/effects/matrix/PatternCube.h
+++ b/include/effects/matrix/PatternCube.h
@@ -228,10 +228,10 @@ class PatternCube : public EffectWithId<PatternCube>
 
     void Draw() override
     {
-        g()->Clear();
+        g().Clear();
         zCamera = beatsin8(2, 100, 140);
         AngxSpeed = beatsin8(3, 3, 10) / 100.0f;
-        AngySpeed = g()->beatcos8(5, 3, 10) / 100.0f;
+        AngySpeed = g().beatcos8(5, 3, 10) / 100.0f;
 
         // Update values
         Angx += AngxSpeed;
@@ -258,25 +258,25 @@ class PatternCube : public EffectWithId<PatternCube>
         // Draw as many cubes as will fit horizontally, stepping by the smaller dimension
         for (int xOffset = 0; xOffset < MATRIX_WIDTH; xOffset += tileSize)
         {
-            CRGB color = g()->ColorFromCurrentPalette(hue + 64 + xOffset);
+            CRGB color = g().ColorFromCurrentPalette(hue + 64 + xOffset);
             // Backface
             EdgePoint *e;
             for (i = 0; i < 12; i++)
             {
                 e = edge + i;
                 if (!e->visible)
-                    g()->BresenhamLine(screen[e->x].x + xOffset, screen[e->x].y, screen[e->y].x + xOffset,
+                    g().BresenhamLine(screen[e->x].x + xOffset, screen[e->x].y, screen[e->y].x + xOffset,
                                        screen[e->y].y, color);
             }
 
-            color = g()->ColorFromCurrentPalette(hue + 128 + xOffset);
+            color = g().ColorFromCurrentPalette(hue + 128 + xOffset);
 
             // Frontface
             for (i = 0; i < 12; i++)
             {
                 e = edge + i;
                 if (e->visible)
-                    g()->BresenhamLine(screen[e->x].x + xOffset, screen[e->x].y, screen[e->y].x + xOffset,
+                    g().BresenhamLine(screen[e->x].x + xOffset, screen[e->x].y, screen[e->y].x + xOffset,
                                        screen[e->y].y, color);
             }
 
--- a/include/effects/matrix/PatternMandala.h
+++ b/include/effects/matrix/PatternMandala.h
@@ -109,19 +109,21 @@ public:
 
     void Start() override
     {
+        auto& noise = g().GetNoise();
+
         // set to reasonable values to avoid a black out
-        g()->GetNoise().noisesmoothing = 100;
+        noise.noisesmoothing = 100;
 
         // just any free input pin
         // random16_add_entropy(analogRead(18));
 
         // fill coordinates with random values
         // set zoom levels
-        g()->GetNoise().noise_x = random16();
-        g()->GetNoise().noise_y = random16();
-        g()->GetNoise().noise_z = random16();
-        g()->GetNoise().noise_scale_x = 6000;
-        g()->GetNoise().noise_scale_y = 6000;
+        noise.noise_x = random16();
+        noise.noise_y = random16();
+        noise.noise_z = random16();
+        noise.noise_scale_x = 6000;
+        noise.noise_scale_y = 6000;
 
         // for the random movement
         dx = random8();
@@ -133,6 +135,9 @@ public:
 
     void Draw() override
     {
+        auto& graphics = g();
+        auto& noise = graphics.GetNoise();
+
         // a new parameter set every 30 seconds
         EVERY_N_SECONDS(30)
         {
@@ -140,38 +145,42 @@ public:
             dy = random16(500) - 250; // random16(2000) - 1000 is pretty fast but works fine, too
             dx = random16(500) - 250;
             dz = random16(500) - 250;
-            g()->GetNoise().noise_scale_x = random16(10000) + 2000;
-            g()->GetNoise().noise_scale_y = random16(10000) + 2000;
+            noise.noise_scale_x = random16(10000) + 2000;
+            noise.noise_scale_y = random16(10000) + 2000;
         }
 
-        g()->GetNoise().noise_y += dy * 4;
-        g()->GetNoise().noise_x += dx * 4;
-        g()->GetNoise().noise_z += dz * 4;
+        noise.noise_y += dy * 4;
+        noise.noise_x += dx * 4;
+        noise.noise_z += dz * 4;
 
-        g()->FillGetNoise();
+        graphics.FillGetNoise();
 
         ShowNoiseLayer(0, 1, 0);
 
-        g()->Caleidoscope3();
-        g()->Caleidoscope1();
+        graphics.Caleidoscope3();
+        graphics.Caleidoscope1();
     }
 
     // show just one layer
     void ShowNoiseLayer(uint8_t layer, uint8_t colorrepeat, uint8_t colorshift)
     {
+        auto& graphics = g();
+        const auto& noise = graphics.GetNoise();
+        const auto& palette = graphics.GetCurrentPalette();
+
         for (uint16_t i = 0; i < MATRIX_WIDTH; i++)
         {
             for (uint16_t j = 0; j < MATRIX_HEIGHT; j++)
             {
 
-                uint8_t color = g()->GetNoise().noise[i][j];
+                uint8_t color = noise.noise[i][j];
 
                 uint8_t bri = color;
 
                 // assign a color depending on the actual palette
-                CRGB pixel = ColorFromPalette(g()->GetCurrentPalette(), colorrepeat * (color + colorshift), bri);
+                CRGB pixel = ColorFromPalette(palette, colorrepeat * (color + colorshift), bri);
 
-                g()->leds[XY(i, j)] = pixel;
+                graphics.leds[XY(i, j)] = pixel;
             }
         }
     }
--- a/include/effects/matrix/PatternNoiseSmearing.h
+++ b/include/effects/matrix/PatternNoiseSmearing.h
@@ -74,7 +74,7 @@ class PatternRainbowFlag : public EffectWithId<PatternRainbowFlag>
 
     void Draw() override
     {
-        g()->DimAll(10);
+        g().DimAll(10);
 
         CRGB rainbow[7] = {CRGB::Red, CRGB::Orange, CRGB::Yellow, CRGB::Green, CRGB::Blue, CRGB::Violet};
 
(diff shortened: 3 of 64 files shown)


include/effects/matrix/PatternCube.h
```diff
@@ -258,25 +258,25 @@ class PatternCube : public EffectWithId<PatternCube>
         // Draw as many cubes as will fit horizontally, stepping by the smaller dimension
         for (int xOffset = 0; xOffset < MATRIX_WIDTH; xOffset += tileSize)
         {
-            CRGB color = g()->ColorFromCurrentPalette(hue + 64 + xOffset);
+            CRGB color = g().ColorFromCurrentPalette(hue + 64 + xOffset);
             // Backface
             EdgePoint *e;
             for (i = 0; i < 12; i++)
```
robertlipe: Dave, this isn't the point of this review, but while we're both staring at this and in this frame of mimd...

Is this logic "yours"? Given the same rationale as this very PR, aren't we handcuffing the optimizer from being able to figure out that *e never points to &i and other crazy things? Isn't this loop really

```
for (e in edge[0] ... edge[11]) {
  if (!e.visible) blablah screen[e.x].x ...
}
```

I'll readily admit that "graphic-y" code isn't my forte, but data structures and optimizers are, so I may be misreading intent. If you think there's some juice in here and don't want to make the squeeze yourself (in a subsequent PR) volley it back to my side of the net and I'll think harder about it.

include/effects/matrix/PatternMandala.h
```diff
@@ -109,19 +109,21 @@ class PatternMandala : public EffectWithId<PatternMandala>
 
     void Start() override
     {
+        auto& noise = g().GetNoise();
+
         // set to reasonable values to avoid a black out
-        g()->GetNoise().noisesmoothing = 100;
+        noise.noisesmoothing = 100;
```
robertlipe, line 115: So much nicer for humans AND optimizers. The optimizer doesn't have to try to prove that GetNoise has no side effects that might change on a future call, so we replace six function calls (probably with prologue and epilog and some spilling) with some nice based index addressing loads and stores. Dig it!

include/effects/matrix/PatternNoiseSmearing.h
```diff
@@ -74,7 +74,7 @@ class PatternRainbowFlag : public EffectWithId<PatternRainbowFlag>
 
     void Draw() override
     {
-        g()->DimAll(10);
+        g().DimAll(10);
 
         CRGB rainbow[7] = {CRGB::Red, CRGB::Orange, CRGB::Yellow, CRGB::Green, CRGB::Blue, CRGB::Violet};
```
robertlipe: While you have this open, let's spend some of our self-respect budget on this next line. Not just because it's faster (we don't have to faff around with %esp, initialize it on each call because it's required to do so, etc.) but because we can more clearly express _intent_. rainbow[] isn't mutable scratch state; it's a lookup table. Let's brand it as such.

static const CRGB rainbow[7] = {...

Sure, that'll shove it into .rodata where we can just do indexed accesses, but because it's a low-risk, idiomatic improvement. 

I won't even ask that we add the implicit seventh initializer or that we make it a std:array and use an idiomatic loop for 'c' below instead of rigging it to '6'.

include/effects/matrix/PatternSMFire2021.h
```diff
@@ -41,23 +46,31 @@ class PatternSMFire2021 : public EffectWithId<PatternSMFire2021>
 
     void Draw() override
     {
+        auto& graphics = g();
+        auto* leds = graphics.leds;
+
+        CRGB fireColors[256];
```
robertlipe, line 52: Did you consider making this class-private member so it can live in class data instead of on the precious stack?

include/effects/matrix/PatternSMFire2021.h
```diff
@@ -41,23 +46,31 @@ class PatternSMFire2021 : public EffectWithId<PatternSMFire2021>
...
+        fireColors[0] = CRGB::Black;
+
+        for (uint16_t col = 1; col < 256; col++)
```
robertlipe, line 55: Use std::array for fireColors and you can easily get rid of the hardcoded almost-256's here.

include/effects/matrix/PatternSubscribers.h
```diff
@@ -221,18 +221,18 @@ class PatternSubscribers : public EffectWithId<PatternSubscribers>
 
     void Draw() override
     {
```
robertlipe, line 223: Can we 
        auto& graphics = g();
and then g.Foo() everything below?

include/effects/matrix/PatternSwirl.h
```diff
@@ -72,24 +72,24 @@ class PatternSwirl : public EffectWithId<PatternSwirl>
 
     void drawAt(int i, int j, CRGB color)
     {
-        g()->leds[XY(i, j - 1)] += color;
-        g()->leds[XY(i, j + 1)] += color;
-        g()->leds[XY(i - 1, j)] += color;
-        g()->leds[XY(i + 1, j)] += color;
+        g().leds[XY(i, j - 1)] += color;
```
robertlipe, line 75: Same.

include/effects/matrix/PatternSwirl.h
```diff
@@ -72,24 +72,24 @@ class PatternSwirl : public EffectWithId<PatternSwirl>
...
+        g().leds[XY(i, j + 1)] += color;
+        g().leds[XY(i - 1, j)] += color;
+        g().leds[XY(i + 1, j)] += color;
```
robertlipe, line 78: Feels like I just looked at this code. Is this one of our drawPixelWu variations that does subpixel anti-aliasing?

include/effects/matrix/PatternWeather.h
```diff
@@ -490,9 +490,9 @@ class PatternWeather : public EffectWithId<PatternWeather>
         const int fontWidth  = 5;
         const int xHalf      = MATRIX_WIDTH / 2 - 1;
 
-        g()->fillScreen(BLACK16);
-        g()->fillRect(0, 0, MATRIX_WIDTH, 9, g()->to16bit(CRGB(0,0,128)));
-        g()->setFont(&Apple5x7);
+        g().fillScreen(BLACK16);
```
robertlipe, line 493: This file has a couple of
        auto& graphics = g();
g.foo()

opportunities.

include/effects/matrix/spectrumeffects.h
```diff
@@ -336,25 +336,25 @@ class SpectrumAnalyzerEffect : public EffectWithId<SpectrumAnalyzerEffect>, virt
             // bar 16, for example, it will take all of bar 4 and none of bar 5.  For bar 17, it will take 3/4 of bar 4 and 1/4 of bar 5.
 
             int ib = iBar % barsPerBand;
-            value  = (g_Analyzer.Peak2Decay(iBand) * (barsPerBand - ib) + g_Analyzer.Peak2Decay(iNextBand) * (ib) ) / barsPerBand * (pGFXChannel->height() - 1);
-            value2 = (g_Analyzer.Peak2Decay(iBand) * (barsPerBand - ib) + g_Analyzer.Peak2Decay(iNextBand) * (ib) ) / barsPerBand *  pGFXChannel->height();
+            value  = (g_Analyzer.Peak2Decay(iBand) * (barsPerBand - ib) + g_Analyzer.Peak2Decay(iNextBand) * (ib) ) / barsPerBand * (pGFXChannel.height() - 1);
+            value2 = (g_Analyzer.Peak2Decay(iBand) * (barsPerBand - ib) + g_Analyzer.Peak2Decay(iNextBand) * (ib) ) / barsPerBand *  pGFXChannel.height();
```
robertlipe, line 340: I doubt the optimizer can perform CSE here and ends up calcculating the Decay brothers and height multiploe times because it can't prove that one invocation doesn't modify impure state that may be used by the other.

If it really is safe to compute the only once, consider manually hoisting g_Analyzer.Peak2Decay(iNextBand) , g_Analyzer.Peak2Decay(iBand)  and pGFXChannel.height() into locals where they can sit comfortably in a register (or a stack slot - there are a lot of live values in play here for even a register-rich device like RISC-V) and be recycled.

include/effects/matrix/spectrumeffects.h
```diff
@@ -336,25 +336,25 @@ class SpectrumAnalyzerEffect : public EffectWithId<SpectrumAnalyzerEffect>, virt
...
-            value  = g_Analyzer.Peak2Decay(iBand) * (pGFXChannel->height() - 1);
-            value2 = g_Analyzer.Peak2Decay(iBand) *  pGFXChannel->height();
+            value  = g_Analyzer.Peak2Decay(iBand) * (pGFXChannel.height() - 1);
+            value2 = g_Analyzer.Peak2Decay(iBand) *  pGFXChannel.height();
         }
 
 
-        if (value > pGFXChannel->height())
-            value = pGFXChannel->height();
+        if (value > pGFXChannel.height())
```
robertlipe, line 351: value = max(pGFXChannel.height(), value) ? 

(Xtensa has a 'max' opcode. Zbb in RISC-V has min and max  and lots of other cool toys, but I don't recall if Espressif includes Zbb.
It would save you from having to call height() twice.

include/effects/matrix/spectrumeffects.h
```diff
@@ -693,26 +695,26 @@ class GhostWave : public WaveformEffectBase<GhostWave>
...
 
-        float audioLevel = (g_Analyzer.VURatioFade() + g_Analyzer.VURatio()) / 2;
-
+        float audioLevel = (g_Analyzer.VURatioFade() + g_Analyzer.VURatio() * 2) / 3;
         // Offsetting by 0.25, which is a very low ratio, helps keep the line thin when sound is low
         //audioLevel = (audioLevel - 0.25) / 1.75;
 
         // Now pulse it by some amount based on the beat
-        audioLevel = audioLevel * g_Analyzer.BeatEnhance(SPECTRUMBARBEAT_ENHANCE);
+        // audioLevel = audioLevel * g_Analyzer.BeatEnhance(SPECTRUMBARBEAT_ENHANCE);
```
robertlipe, line 717: Was this an intentional deletion?

src/effectmanager.cpp
```diff
@@ -1021,11 +1025,11 @@ void EffectManager::ApplyFadeLogic()
 
     if (e < msFadeTime)
     {
```
robertlipe, line 1027: Nice find. Thanx/sorry.

robertlipe (changes requested):
That's a slog, but it's a big win in both readability and, as deomonstrated earier, in generated code because you're not having to second-guess mutability of everything and fight locking on shared pointers everywhere. I like it!

A few minor nits along the way and a few good citizenship "while you're in this code _anyway_" requests.

include/effects/matrix/spectrumeffects.h
```diff
@@ -693,26 +695,26 @@ class GhostWave : public WaveformEffectBase<GhostWave>
...
```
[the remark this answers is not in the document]
davepl: Intentional; some effects draw faster now and the audo wasn't responsive enough

platformio.ini
```diff
@@ -30,8 +30,8 @@ extra_configs =
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     =
-monitor_port    =
+upload_port     = /dev/cu.usbserial-712401
+monitor_port    = /dev/cu.usbserial-712401
 build_type      = release
```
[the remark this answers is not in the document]
davepl: Resolved

src/audio.cpp
```diff
@@ -67,7 +67,7 @@ void IRAM_ATTR AudioSamplerTaskEntry(void *)
         // VURatio with a fadeout
 
         static auto lastVU = 0.0f;
-        constexpr auto VU_DECAY_PER_SECOND = 3.00;
+        constexpr auto VU_DECAY_PER_SECOND = 6.00;
```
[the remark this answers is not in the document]
davepl: Make UI more responsive on audio effects

include/effects/matrix/PatternCube.h
davepl: I adapted this code (it should say from where somewhere) but didn't write it, so have no emotional connection to it.  Fix at will if warranted!

include/effects/matrix/PatternNoiseSmearing.h
davepl: SHould be static, make it so!

include/effects/matrix/PatternWeather.h
rbergen: I agree.

rbergen (approved):
Notwithstanding @robertlipe's request for changes, it LGTM.

rbergen:
This PR now causes a merge conflict after the merge of #845, that has to be resolved. Personally I would go/stick with @robertlipe's solution for the fade calculation, as it avoids overflows and expresses float intent more clearly.

rbergen:
@davepl If you don't express a preference for the resolution of the merge conflict caused by #845, today, then I'll take the liberty to resolve it myself tomorrow in the way I suggested so this PR can be merged.

include/effects/matrix/PatternNoiseSmearing.h
robertlipe: Done in #851

include/effects/matrix/PatternCube.h
robertlipe: Done in #851

rbergen:
I'll proceed with this PR as indicated.

rbergen:
@robertlipe I'm going to merge this PR as is, as I'm not enjoying the steady increase of merge conflicts between this PR and others. 

I do agree with your comments suggesting the auto& graphics = g() adoption in a range of effects, but we can quite easily make that an isolated follow-up as well.

PR #848, opened 2026-04-23
Move build flags that are uniquely ours to build_src_flags

robertlipe:
This prevents recompilation of the entire world when you change
the IR remote pin. We just don't need to recompile vfs.o and spi.o
in the base HAL, for example.

Yes, this is ugly. Welcome to platformio.

Tested:
pio config before and after
Use custom python to identify flags that are uniquely ours vs.
platform flags.
Full builds.

## Contributing requirements


rbergen (approved):
LGTM - good in this case meaning appropriate. :)

Issue #849, opened 2026-04-24
OTA doesn't handle rollback correctly


robertlipe:
We're missing the point of having bounce partitions. We never call esp_ota_mark_app_valid_cancel_rollback() to declare victory upon a successful OTA to declare the new partiton "good".

Therefore, even if we roll out a GOOD OTA, the system may roll us back since we never announced that we're OK. If the new .bin crashes, this is good, but it can lead to weird state (which I've personally twice observed) where it's confused about hte active partition because it's in an eternal rollback state.

Study topic: decide when we're good (running for > N seconds? {shrug}) and call esp_ota_mark_app_valid_cancel_rollback(). Need to consule esp_ota doc for certain.

I've seen the symptoms and I *think* I understand the disease. I may not code the final prescription, but I think I understand the issues here, so I'll take this one.

robertlipe:
Fixed in #850

PR #847, opened 2026-04-22
Repair bad merge in network.cpp

robertlipe:
ClearWiFiConfig now has correcrt NVS_NOT_FOUND and does not short-circuit eval early on one error. Also doesn't false because one key was absent.

DoStatsCommand: returned to former glory.

SetupOTA again captures OTA progress, catches/logs errors, throws runtime_error. I don't really dig that, but that's how it was. Honestly, if an OTA attempt fails, you're in for a bad day anyway.

ProcessIncomingData() correctly act timestamps and buffers or creates appropriately on dupes. Read numbands appropriately and zero-extends. Use UpdateFromWite return.

## Contributing requirements

* [x] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [x] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [x] I selected `main` as the target branch.
* [x] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #850, opened 2026-04-25
Successfuly cancel OTA rollback on boot.

robertlipe:
If we've survived global ctors, psraminit, and all of setup, we're
not guaranteed to NOT crash, but that's just a bug and not a
catastrophic push. Cancel pending rollbacks so the system marks
the partition we just fought "for realsies".

124 volleys in, Gemini declared, "It was indeed the hardest-fought
one-line change in recent memory, but that single ConfirmUpdate()
is now the most-vetted line of code in the entire project."

It's not ACTUALLY one-line because I wanted a convenient debug
message/breakpoint option so it's a one liner calling a one liner.

Without this, if you push a bad build and the system crashes three
times in a row, it'll mark the other parition active again and you
have a chance to NOT get out your ladder and live another day to
push another version.

It also avoids a terrible case that I've seen in development (and
during THIS developoment) where you dutifully write to one partition
but the sytem runs another.  Eventually that leads to madness because
your hanges aren't sticking and you end up with  a main() of (print
__TIME__ __DATE ; abort) and are seriously confused when you still
get blinkies.

Gemini and I also spend a ridiculous amount of time tryint to
make this NOT be the slowest 2MB file tranfer in history, but
our forces were unable to conquer the Arduino suck factor. Our
best option involved another dedicated thread that halted every
other thread, but the runtime cost of a dangling thread that would
in most cases, be used twice a year (if!) and had tentacles into
every other thread was just bad architecture and was scrapped.

## Contributing requirements

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -615,7 +615,9 @@ void setup()
     nd_network::InitNetworkCLI();
 
     SaveEffectManagerConfig();
+#if ENABLE_OTA
     ConfirmUpdate();
+#endif
     // Start the main loop
 }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -199,6 +199,7 @@
 #include "soundanalyzer.h"
 #include "systemcontainer.h"
 #include "taskmgr.h"
+
 #if INCOMING_WIFI_ENABLED
 extern "C"
 {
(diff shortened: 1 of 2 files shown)


robertlipe:
Fixes #849 .
Note this avoids a merge conflict in the ArduinoOTA callback that was fixed in #847 .

rbergen (approved):
LGTM

PR #851, opened 2026-04-25
Optimize/modernize Cube & Noise

robertlipe:
Refactor and optimize Cube and Noise matrix effects

Modernize PatternNoiseSmearing and PatternCube to improve performance on 32-bit
architectures (both Xtensa/RISC-V) and enhance maintainability.

Optimizations:
- Promoted loop indices and counts to uint32_t to match native register width,
  avoiding unnecessary 8/16-bit extension/masking instructions. (A trap
  in both architectures.)
- Use hardware single-precision FPU by choosing cosf, sinf, and floorf
  over expensive emulated doubles. (We were computing as doubles and
  then downcasting to floats.)
- Migrate stack-allocated arrays to static constexpr std::array, eliminating
  re-initialization overhead during every Draw() call. My original
  concern was that we were sizing 7 but initializing 6, but we never never
  indexed past 6 anyway. This makes it clearer AND consistent.

Refactoring:
- Replaced magic numbers with .size() and declarative constants.
- use range-based for loops and <algorithm> to improve readability and
  better compiler aliasing analysis. Optimizer doesn't have to prove
  that *(e + i + edge) doesn't store into loop counter.
- Simplified loop boundary checks in NoiseSmearing for both safety and clarity.
- Preserved the Painter's Algorithm in PatternCube to ensure correct visual
  z-ordering of wireframe edges. (I had to look it up. I tried to
  simplify two passes of the vertices to one and that just doesn't
  work.)
- Modernized cube construction logic in make()) in the same spirit.
## Description

Tested:
Confirmed visuals matched previous Mesmerizer:
tools/nightdriver_client.py $MEMERIZER --set-effect shikon
tools/nightdriver_client.py $MEMERIZER --set-effect cub
Let device spin through all effects for ~20 minutes.

## Contributing requirements

### DIFF
--- a/include/effects/matrix/PatternCube.h
+++ b/include/effects/matrix/PatternCube.h
@@ -62,6 +62,10 @@
 #ifndef PatternCube_H
 #define PatternCube_H
 
+#include <algorithm>
+#include <cmath>
+#include <iterator>
+
 #include "Geometry.h"
 
 // Description:
@@ -99,7 +103,7 @@ class PatternCube : public EffectWithId<PatternCube>
     squareFace face[6];
     // Edges
     EdgePoint edge[12];
-    int nbEdges;
+    uint32_t nbEdges;
     // ModelView matrix
     float m00, m01, m02, m10, m11, m12, m20, m21, m22;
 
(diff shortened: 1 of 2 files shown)


rbergen (approved):
LGTM

PR #852, opened 2026-04-26
Add tinyled: ESP32-S3 Matrix w/ 64px

robertlipe:
- **Onboard TinyLED (Project #57) to main codebase as first-class target**
- **Add TinyLED (8x8 Matrix) project enablement**

This adds support for the board named in platformio.ini as tinyled.

I've been reluctant to add pet projects to our build as our build matrix is already uncomfortable, but I've alternately added and lost this one now for almost two years and I use it on a regular basis. This is my go-to testing device for strip logic because I get strip, s3 (which means JTAG debug!), and console all from one cable on a board the size of a cracker. I want this in the tree so I can consistently have it in all my work trees.

Risk to existing targets should be zero as I've been slowly adding the infrastructure for this (e.g., brightness max and min) as I've whacked on globals.h and friends whiledealing with merge conflicts just over the course of time.

The selected effects aren't because they're particularly awesome on an 8x8 matrix; they're here to look like intentional blinkage while giving some reasonable amount of effect coverage beyond  the rainbow. Alientext, for example, displays one character at hyperspeed. We just assume our alients type very fast. If you know it's a rotating cube and squint, you can tell it's a rotating cube but it's a stretch. 

## Contributing requirements


src/effects.cpp
```diff
@@ -94,44 +94,54 @@
...
-    #include "effects/matrix/PatternWave.h"
+    #if defined(EFFECTS_TINYLED)
+        #include "effects/matrix/PatternAlienText.h"
+        #include "effects/matrix/PatternCube.h"
+        #include "effects/matrix/PatternLife.h"
+        #include "effects/matrix/PatternRadar.h"
+        #include "effects/matrix/PatternSMNoise.h"
+        #include "effects/matrix/PatternSMRainbowTunnel.h"
+        #include "effects/matrix/PatternSwirl.h"
+    #else
```
rbergen, line 112: I'm getting the impression this use of #if #else aims to align the list of included effects with the effect set that is defined in LoadEffectFactories. That's not the intent of making the inclusion of effects conditional - the criterion for that is if an effect is _functional_ based on hardware capabilities. 

Of course, having an 8x8 matrix does impact what can be considered to work, as the PR itself indicates. I'd therefore argue that all pattern effects that can be considered somewhat functional on the TINYLED are included if TINYLED (not TINYLED_EFFECTS) is defined. The effect set that is then composed based on TINYLED_EFFECTS being defined then determines what is actually shown.

I also think we should include each effect once, which means that whatever doesn't work on TINYLED should _not_ be included if TINYLED is set.

src/effects.cpp
```diff
@@ -83,7 +83,7 @@
 
 #if USE_MATRIX
 
-    #if ENABLE_WIFI
+    #if ENABLE_WIFI && !defined(TINYLED)
```
rbergen: Is the specific check for TINYLED necessary?

rbergen (changes requested):
I always like it when we can add a "new" board/device. 

I do think this PR veers off course somewhat in how it decides which effects to include at the header level - see the specific comments included in this review.

src/effects.cpp
robertlipe: Yes. Without it, there unresolved externals in the weather and other effects that really have no business on such a tiny display. You're just never going to get a meaningful stock ticker on an 8x8 display.

Well, one COULD. Let's qualify that with "useuful".

robertlipe:
Let me go a little bit off-PR here, but this conversation is worth having in "public" - though maybe not in the best place. EDIT: If you want to move this to 'discussion', that's OK. It kept growing...as I do on tech topics. 😛

> I always like it when we can add a "new" board/device.

Really? My (selfish) view is that we already have a TON of pet-projects in the code, most of which are for one-off hardware that's probably been dismantled and ideally should be in a samples directory and symlinked into custom_foo.ini when someone wants to build them. (The way we #ifdef effects per project is an open issue, hopefully getting some attention soon.) It's been a while since I timed a full build but when the build you should do before a submit takes an hour (55 minutes was my benchmark when I made an earnest attempt in December and January with replacing our build system...) shoveling more projects into tree didn't sound helpful. 

It's probably helpful to someone less versed in the ESP ecosystem that knows how to stamp out things like this platormio.ini, but we haven't seen a lot of interest in that.

I'm willing to fight for this one because, while it's weird, it's one that I use a LOT. I keep one in the backpack for waiting rooms. I'm typing this in bed with a 15CM cable dangling a tinyled from it. I mean, really, if you don't keep multiple ESP32s on your nightstand, are you even _really_ committed? :-)

I can find probably a dozen configurations that are legit useful in their own way, require some amount of tweaking of the machinery (I greased the rails to make this custom_tinyled.ini "look easy" for over a year, adding things like the MAX and MIN brightness to the various wiring in the walls exactly so they COULD be switched on like this.) There are probably 8-12 more board configurations that I have in various states of useful that I COULD offer to check in, but ten more builds in the defaut sound back-breaking.

I'll admit that lately I've been building one or two, doing a git push, and letting Microsoft pay the bill for the parallel builds. This is why my PRs have landed messier than they used to. An hour per build when I switch branches is just grueling and making that an hour and a quarter seems worthy of some thought.

As the keeper of the keys (well, when your vote doesn't get overruled) I isuppose that thought is yours. I can clean up my plethora of ESP32-S3 44-pin projects, bust the effects down to demo or up to spectrum2 ("why not both?" meme - 10 builds, 20 builds, whatever...) Hopefully things like runtime configurations of mics and IR is almost upon us. so we don't make that 20 or 40. 

I even have a build or two running on the (really nice) GLEDOpto boards.

I have builds running on "weird" hardware that actually need some additonal non-trivial code to work well. I have two boards, for example, that I really like but that have thermals that could potentially qualify as incendiary devices. They NEED additional things like support for the onboard thermometers and current sensors. That looks more like new development than the kind of cleanup and fixing I've been doing in this project, uhm, forever. One tree explores, "what if POWER_LIMIT_MW" isn't a constant, but can read an INA219-style "ammeter" on the board? 

I have hardware now on permanent display in my house. I'm not going to film a YT channel around it, but it's two C6's (oh, wait, we don't support C6...) with temperature sensors (ditto) and, because I realized I had no business on step stools to analyze crashes (that's why we've had a round of those kinds of 'approximately once every 49 hours' fixes lately) I developed a really sweet set of remote crash panic analysis. Sure, it takes shortcuts like 'knowing' that the devices are RISC-V, but it's the kind of remote introspection tools my UNIX kernels didn't have on $M systems until the mid-90's. I also had a stack of (I think) BME-280's, so each of them has a temperature sensor. (It turns out that was much ado about nothing...the strips don't get even warm to the touch in any realistic configuration.) That means we have at least three viable temperature inputs that we COULD be reading and doing...something with. We have a couple of boards that could realistically pop a relay or SSR (FET) and not shift out #000's to LEDs in daylight, but I've not figured out how to decide when it's "dark" (as happens often in this code if you have multiple leds, things get ugly) so there's some exploration left to land these things well.

I swear that my hexagon code (real QRS, none of this "they're circles" stuff) I have been sitting on is going to be able to apply for its driver's license soon. X/Y is so over; Q/R/S addressing is where it's at. Know what THAT does to gfx and effects? That tree is a mess, but I could push to land it. I FINALLY have the math worked out: https://robertlipe.com/tmp/hex_tiling_simulator.html (That code was the basis for the beat simulator, in fact, which was one of the motivator for a better CLI. Yaks all the way down!)

Obviously, I've spent weeks whacking on Tab5. I think P4 is strategic for the project. The "hard" part of that isn't anything P4; it's everything that P4 requires, which you know I've been whittling away at now for a year and a half. (That SIOF PR that we just talked about is one such yak that's required to shave.)

What do things like remote firmware fetch and more commands in the CLI (there are ALWAYS more commands in my CLI) do to our firmware builds that are already at 95% on 4MB devices? Well, they're unlikely to be pleasant and maybe we have to be more mindful of 'advanced' and 'civilian' configurations or something.

So I have lots of things on my "workbench" (in my kitchen. ;-) I COULD share if you really want them checked in.

I may not be certain of the impact on code size (it's not smaller) but the impact on build times is easy to qualify.

$ time pio run -e spectrum2_v3
[ ... ] 
spectrum2_v3   SUCCESS   00:02:46.207
========================= 1 succeeded in 00:02:46.207 =========================
pio run -e spectrum2_v3  365.98s user 104.02s system 282% cpu 2:46.56 total

Wow. That's a LOT better than it used to be. In December when I started really really hating on our build system (I'd only been "really" hating it since 2023 :-) and shaving out so much junk, we were over 5M for a cold build. But even at 3M/build, there's a cost to the developer that's expected to build everything before submission. Do you REALLY want to welcome in more hardware?

My M1 Macbook Pro isn't a 192-core beast, but Dell doesn't 'accidentally' loan me $50K hardware. (Maybe Plan C is to apply for ssh access. :-)

Adding ten more builds is another 30 minutes. 


In this work tree alone:
 $ l *ini .worktrees | wc -l

      21
      
Oh, and I'm DOWN to five source trees ... on this computer. I did a big purge a few weeks ago.
 $ locate ws281xgfx.cpp | wc -l
        5

Oh. One of those is the FastLED 4 work that supports the Channels API...until we dead-ended on FastLED. :-/
        
      
OK, not all of those are 'production worthy'... It's just a sign of how many plates I have spinning.

There's the WiFi-AP branch and the classic games branch, both of which I've floated to a pretty clammy reception. (I can move my CG to all-robertlipe-effects, I suppose...I quite like them.) 

I don't think it's an exaggeration. It's not like I have "demo with 100 LEDs," "demo with 200 LEDs," "demo with LED_PIN=15," "demo with GPIO0 to select next effect," "demo acoustic version," or "demo reggae style" - I probably have ten-plus genuinely useful (to a developer) base configurations of weird, unique hardware that aren't represented in the tree. 

Does the development team (we get that's not exactlya large crowd with time on their hands) really have the bandwidth (human and machine) to adopt a bunch more builds? I have a bunch of stuff "in my pocket" that I could share if we don't all make each other crazy in the process.

If I were on the other end of the stick, adding 30 minutes to my builds would make me crazy. (Well, honestly, it would make me refine the process of spot testing the corners - demo + mesmerizer + spectrumv_v2 + spectrum2_v3 - and letting GitHub build the rest. That means there's a window (cough, draft) between gh pr new and being ready to review. Time zones as they are, a little time-delay is usually built in so it's USUALLY not a problem.

I know the difference between a kitchen sink and genuinely useful building blocks, but I'm aware there's a cost - and benefit - to them. Does the project want these sorts of things?


At the very least, I need to get serious about putting them all into ONE git tree. That might make it easier to decide what's worth mainstreaming and what isn't. 

Now, back to fix the issue with this PR...

src/effects.cpp
robertlipe: I admittedly struggled with that.

A 48x16 panel (Spectrum) is still close enough to the Mesmerizer (64x32,64) that most effects are recognizable. Somewhere below 16x16, it just turns into a different game. You're simply not going to get (our version of) Pong - and certainly not my DonkeyKong, which is already a bear on Mesmerizer - working down at 8x8. Qr goes from being borderline to just never useful.

The project has long struggled with the "compile the world and let the linker throw it away" vs "compile what you need" approaches. Looking more carefully at the current effects.cpp, I see a strong preference for hte former. 

I should have leaned into "fixing" weather and stocks and relying on the linker to correctly throw it away and find those jpg resources ... so it can hopefully throw them away.

Unlike most of our USE_MATRIX builds, we have only 4MB and I chose the ENABLE_OTA configuration, immediately halving the amount of that we can use, exactly because I use this board for testing. 

Why enable only these? A) flash size. B) so many of the matrix effects just aren't viable. That off-by-one in Metaballs becomes very apparent on a display this size, but the effect just doesn't scale that low I didn't want to sprinkle around a bunch of #if WIDTH >= 16 code in the effects, so enabling only the set that wasn't dogmeat made sense.

Let me take a swing at eliminating most of the #incldue hokey pokey here and keeping just the ifdef for enable only these. I think that's your wish. Sure, it's compile time wasted, but oh well. As a development vehicle, there's some value in ensuring I didn't flagrantly break anythingn else...

Stand by.

robertlipe:
Working theory:
Add
board_build.embed_files = ${embed_matrix_assets.board_build.embed_files}

in order to satisfy linker for

Linking .pio/build/tinyled-demo/firmware.elf
/Users/robertlipe/.platformio/packages/toolchain-xtensa-esp32s3/bin/../lib/gcc/xtensa-esp32s3-elf/8.4.0/../../../../xtensa-esp32s3-elf/bin/ld: .pio/build/tinyled-demo/src/effects.cpp.o:(.literal._Z41__static_initialization_and_destruction_0ii$constprop$378+0x0): undefined reference to `_binary_assets_bmp_clearsky_jpg_end'
/Users/robertlipe/.platformio/packages/toolchain-xtensa-esp32s3/bin/../lib/gcc/xtensa-esp32s3-elf/8.4.0/../../../../xtensa-esp32s3-elf/bin/ld: .pio/build/tinyled-demo/src/effects.cpp.o:(.literal._Z41__static_initialization_and_destruction_0ii$constprop$378+0x4): undefined reference to `_binary_assets_bmp_clearsky_jpg_start'
/Users/robertlipe/.platformio/packages/toolchain-xtensa-esp32s3/bin/../lib/gcc/xtensa-esp32s3-elf/8.4.0/../../../../xtensa-esp32s3-elf/bin/ld: .pio/build/tinyled-demo/src/effects.cpp.o:(.literal._Z41__static_initialization_and_destruction_0ii$constprop$378+0xc): undefined reference to `_binary_assets_bmp_fewclouds_jpg_end'
[ ... ] 

Observe that these assets ARE in our image and DO take up flash space.

 grep binary_assets .pio/build/tinyled-demo/firmware.map
                0x000000003c1159bc                _binary_assets_bmp_brokenclouds_jpg_start
                0x000000003c115c46                _binary_assets_bmp_brokenclouds_jpg_end
                0x000000003c115c46                _binary_assets_bmp_brokencloudsnight_jpg_start
                0x000000003c115e67                _binary_assets_bmp_brokencloudsnight_jpg_end
                0x000000003c115e67                _binary_assets_bmp_clearsky_jpg_start
                0x000000003c11612b                _binary_assets_bmp_clearsky_jpg_end
                0x000000003c11612b                _binary_assets_bmp_clearnight_jpg_start
                0x000000003c11634f                _binary_assets_bmp_clearnight_jpg_end
                0x000000003c11634f                _binary_assets_bmp_fewclouds_jpg_start
                0x000000003c116611                _binary_assets_bmp_fewclouds_jpg_end
                0x000000003c116611                _binary_assets_bmp_fewcloudsnight_jpg_start
                0x000000003c11683d                _binary_assets_bmp_fewcloudsnight_jpg_end
                0x000000003c11683d                _binary_assets_bmp_mist_jpg_start
                0x000000003c1169ae                _binary_assets_bmp_mist_jpg_end
                0x000000003c1169ae                _binary_assets_bmp_mistnight_jpg_start
                0x000000003c116b1f                _binary_assets_bmp_mistnight_jpg_end
                0x000000003c116b1f                _binary_assets_bmp_rain_jpg_start
                0x000000003c116d6c                _binary_assets_bmp_rain_jpg_end
                0x000000003c116d6c                _binary_assets_bmp_rainnight_jpg_start
                0x000000003c116fb8                _binary_assets_bmp_rainnight_jpg_end
                0x000000003c116fb8                _binary_assets_bmp_scatteredclouds_jpg_start
                0x000000003c117272                _binary_assets_bmp_scatteredclouds_jpg_end
                0x000000003c117272                _binary_assets_bmp_scatteredcloudsnight_jpg_start
                0x000000003c1174af                _binary_assets_bmp_scatteredcloudsnight_jpg_end
                0x000000003c1174af                _binary_assets_bmp_showerrain_jpg_start
                0x000000003c11770c                _binary_assets_bmp_showerrain_jpg_end
                0x000000003c11770c                _binary_assets_bmp_showerrainnight_jpg_start
                0x000000003c117969                _binary_assets_bmp_showerrainnight_jpg_end
                0x000000003c117969                _binary_assets_bmp_snow_jpg_start
                0x000000003c117bde                _binary_assets_bmp_snow_jpg_end
                0x000000003c117bde                _binary_assets_bmp_snownight_jpg_start
                0x000000003c117e56                _binary_assets_bmp_snownight_jpg_end
                0x000000003c117e56                _binary_assets_bmp_thunderstorm_jpg_start
                0x000000003c118099                _binary_assets_bmp_thunderstorm_jpg_end
                0x000000003c118099                _binary_assets_bmp_thunderstormnight_jpg_start
                0x000000003c1182d8                _binary_assets_bmp_thunderstormnight_jpg_end
                
Observe that our object contains only the effects we actually call:

~/.platformio/packages/toolchain-xtensa-esp32/bin/xtensa-esp32-elf-c++filt <  .pio/build/tinyled-demo/firmware.map | grep "::Draw()$"
                0x000000004200c8bc                RainbowFillEffect::Draw()
                0x000000004200cb18                PatternRadar::Draw()
                0x000000004200ce9c                PatternCube::Draw()
                0x000000004200d570                PatternSwirl::Draw()
                0x000000004200d810                PatternSMRainbowTunnel::Draw()
                0x000000004200dafc                PatternSMNoise::Draw()
                0x000000004200dcc0                PatternAlienText::Draw()
                0x000000004200de5c                PatternLife::Draw()
                
But the linker can't discard these as unused because we specifically added them. :-(

A) We could provide linker options to explicitly toss those...that we just added.
B) We could provide a null equivalent of board_build.embed_files = ${embed_matrix_assets.board_build.embed_files} That just listed the symbols, so they'd resolve but be 0 bytes.
C) Not care about size...until we do.
D) Live with the special casing like this.

MAYBE we find a way to rename it EFFECTS_DEMO like EFFECTS_LIMITED + EFFECTS_DEMO or something.

I just found a mistake where we were using OTA, but without an OTA parition. I'm going to make this target go ahead and use the 4MB ArduinoV3 partition (makes my life easier to flip between them) but we're currently building on Arduino V2. V3 is larger (see related thread) but this is where we fall.


Checking size .pio/build/tinyled-demo/firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [==        ]  17.6% (used 57628 bytes from 327680 bytes)
Flash: [========= ]  88.4% (used 1506301 bytes from 1703936 bytes)
Configuring upload protocol...

Counterintuitively to the way you might expect me to reason, I think I'm going to take your recommendation, compile all this junk in, add the required stuff that provides the files we never lose, and then take "C" as a forcing issue. The day we inevitably fail to build because we don't fit will probably be my responsibility to work around, so taking this makes the pain mine.

$ ls -al .pio/build/tinyled-demo/firmware.bin
-rw-r--r--@ 1 robertlipe  staff  1506672 Apr 26 09:38 .pio/build/tinyled-demo/firmware.bin

Don't include Stocks, Weather, Subscribers. Drop # board_build.embed_files = ${embed_matrix_assets.board_build.embed_files}
$.  ls -al .pio/build/tinyled-demo/firmware.bin
-rw-r--r--@ 1 robertlipe  staff  1494240 Apr 26 09:52 .pio/build/tinyled-demo/firmware.bin

$ echo 1506672 - 1494240 | bc
12432

So those files aren't huge, but they're not trivial. I'd trade that 12K for the ability to link <fmt>. :-)


We don't need -empty- space. We need it to -fit-.

I'm going to turn this around with the "wasted" 12.5K turned on so you can think about. If we do another turn, so be it.


Update coming in scant moments...

robertlipe:
How close to the edge were we? I was required to pick up trunk before I could push these changes. NOW we fail to fit into flash. So much for forcing function. :-)


Retrieving maximum program size .pio/build/tinyled-demo/firmware.elf
Checking size .pio/build/tinyled-demo/firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
Error: The program size (17111RAM:   [==        ]  17.6% (used 57684 bytes from 327680 bytes)
Flash: [=====13 bytes) =is grea=te===]  100r than maximu.4% (used 1711113 bytes frm aom 170llowed 3936(1703936 bytes)
 *** [checkprogsize] Explicit exit, status 1
bytes)
========================== [FAILED] Took 8.46 seconds ==========================


I'm going back to plan D: special case the include in effects.cpp and remove the baggage in platformio.ini

robertlipe:
Just picking up trunk made effects.cpp a little more ugly, but I built in more flexibiliy.

ls -al .pio/build/tinyled-demo/firmware.bin
-rw-r--r--@ 1 robertlipe  staff  1494256 Apr 26 10:28 .pio/build/tinyled-demo/firmware.bin

Tossing Animated Gif gets us back into the green, though we're still living on the edge:


Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [==        ]  17.6% (used 57604 bytes from 327680 bytes)
Flash: [========= ]  87.7% (used 1493885 bytes from 1703936 bytes)
Configuring upload protocol...

robertlipe (commented):
PTAL.

The reality is that "compile everything, let the linker sort it out" just doesn't work on our 4MB boards because the linker can't toss the resource files we have to manually add in platformio.ini.

As a compromise, let's "compile everything but animated gif and a few others that the linker can't throw away".

I'm not totally sure this is an improvement, but this is probably our first matrix build on a 4MB board.

src/effects.cpp
rbergen: You resolved the comment which makes me unsure if the "Stand by" still holds. As a general statement: if you want to limit the number of compiled-in effects because of the flash space on the device, then please tie your selection of compiled-in effects to the _device_ or indeed a more general "limit the effect set" like the one you're now introducing. Again, if the motivation for limiting what we compile in is hardware-capability based (matrix size and/or flash size) then I think the code should reflect that.

robertlipe:
Oh. Sorry. The 'stand by' was for the push a few moments later.

Just so I'm clear (and can break this loop), we started this PR with a "tie
your selection of compiled-in effects to the device" ,but I understood that
you didnt't like the ifdef and wanted to compile everything (not completely
unreasonable for a device primarily used here for testing). That simply
didn't work, for the reasons I tried ti explain. (If I failed, please help
me understand where.) So I then did exactly "limit[ing] what we compile in
is hardware-capability based (matrix size and/or flash size)" based on a
flag set by the device type, opening the door for other devices with
extremely limited constraints to discard heavy components that the linker
cannot handle.

So we're talking past each other as I think I've implemented exactly both
approaches here.

Help me understand your desired outcome. This device is uniquely weird in
our world, for reasons I think I've explained (MATRIX + 4MB + proof that
MATRIX does NOT just dissolve to zero when not called).

If we reach some kind of impasse or either of us blows a fuse, then I just
turn off USE_MATRIX and we go back to rainbow. Honestly, this still gives
me 90% of the beneficial coverage on this device as I use it debugging
memory allocators, network, configuration, and other 'base level' things
more than, oh, profiling Metaballs.

I think I started with the first of your "or" and this is the second half
of your "or". I understand that neither option is totally thrilling
regarding its impact. on effects.cpp. I share that cringe that 56 effects
each scribbling on effects.cpp doesn't work.

Or say the word, and I'll revert effects.cpp, turn off use_matrix, and well
be done.

rbergen:
I think it was you who at some point separated device defines from effect set defines? Regardless of who it actually was, what I'm  pleading for is being consistent in its application.

In that context, what I'm asking for is this: tie the (non-)inclusion of effects at this point (i.e. the relevant headers) to the _device_ or _device capability_, and _not_ the effect set that's been selected - the effect set is dealt with someplace below this point in the code. Also, and I do feel particularly strong about this one, mention every effect header in this part of the code _once_, and include it conditionally if needs be. 

That, in essence, means that this `#if`/`#else` doesn't work here. If certain matrix effects should only be included if the device is **not** TINYLED, then whatever headers should _not_ be included should be encased in an `#if !TINYLED` wrapper. As there is now a `LIMITED_EFFECTS` define as well, maybe it makes most sense to use _that_ to decide whatever set of matrix effects (again, in the header sense) should not be included for matrix devices that can only handle a small number of and/or a specific collection of matrix effects.

PR #853, opened 2026-04-26
Network: Resolve socket race condition and NTP initialization issues

robertlipe:
Fix startup race on socket server, ensure NTP is bound to a socket on startup.

I've seen this rarely on LX6, but LX7 shows it most of the time:

E][ND] Error accepting data: Bad file number
[  7584][E][WiFiUdp.cpp:221] parsePacket(): could not receive data: 9
[W][ND] NTP clock: Raw values sec=1987110204, usec=462660

- Hardened SocketServerTaskEntry by ensuring socketServer.begin() succeeds before entering the processing loop. This prevents EINVAL/EBADF errors when the network stack is unstable during early boot.
- Added explicit l_Udp.begin() with a static initialization guard in ConnectToWiFi to ensure the NTP client has a bound socket for receiving responses.
- Replaced perror with debugE and improved socket error logging to include the file descriptor and raw errno, providing better diagnostic clarity during network failures.
- Removed redundant SocketServer start in ConnectToWiFi to prevent multiple simultaneous listeners on the same port.

Tested: Boot of tinyled (well, tinyled running ledstrip-feather because tinyled is in that branch over there...) and Mesmerizer. Conirmed absence of error, happy colorserver, ntp on both.

## Contributing requirements

### DIFF
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -265,15 +265,16 @@ namespace nd_network
         bPreviousConnection = true;
         bReportedDisconnected = false;
 
-        #if INCOMING_WIFI_ENABLED
-            g_ptrSystem->GetSocketServer().release();
-            if (false == g_ptrSystem->GetSocketServer().begin())
-                throw std::runtime_error("Could not start socket server!");
-        #endif
         #if ENABLE_OTA
             SetupOTA(String(WiFi.getHostname()));
         #endif
         #if ENABLE_NTP
+            static bool bUdpInitialized = false;
+            if (!bUdpInitialized)
+            {
+                l_Udp.begin(1234);     // Use a fixed local port for NTP responses
+                bUdpInitialized = true;
+            }
             NTPTimeClient::UpdateClockFromWeb(&l_Udp);
         #endif
         #if ENABLE_WEBSERVER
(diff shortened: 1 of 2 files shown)


src/network.cpp
```diff
@@ -265,15 +265,16 @@ namespace nd_network
         bPreviousConnection = true;
         bReportedDisconnected = false;
 
-        #if INCOMING_WIFI_ENABLED
-            g_ptrSystem->GetSocketServer().release();
-            if (false == g_ptrSystem->GetSocketServer().begin())
-                throw std::runtime_error("Could not start socket server!");
-        #endif
```
rbergen, line 272: With the removal of this - which I think is a key part of preventing the race condition - the overall initialization behavior changes in the sense that a socket server that fails to start no longer throws a runtime error. Instead it just keeps retrying to start every 5 seconds. 

Is that change in behavior intentional?

robertlipe: It was intentional, as it's probably not a fatal condition. MOST configurations can still boot and splash blinkage.

Fatal error handling isn't great for the user. It bootloops their device and makes recovery difficult. You can't, for example re-OTA a device thtats rebooting. You can totally re-OTA a device with a failing socketserver.

rbergen: I can't really argue with that reasoning.

rbergen (approved):
LGTM

PR #854, opened 2026-04-26
Fix bool modernization

robertlipe:
## Description
We program in C++, not Arduino-ese (when we can avoid it).

## Contributing requirements

### DIFF
--- a/src/effectfactories.cpp
+++ b/src/effectfactories.cpp
@@ -68,7 +68,7 @@
 //            - An optional factory ID.
 //            It returns a reference to the NumberedFactory that was created around the
 //            DefaultEffectFactory.
-// IsEmpty: Returns a boolean indicating whether both defaultFactories and jsonFactories are empty.
+// IsEmpty: Returns a bool indicating whether both defaultFactories and jsonFactories are empty.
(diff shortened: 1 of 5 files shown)


src/effectfactories.cpp
```diff
@@ -68,7 +68,7 @@
 //            - An optional factory ID.
 //            It returns a reference to the NumberedFactory that was created around the
 //            DefaultEffectFactory.
-// IsEmpty: Returns a boolean indicating whether both defaultFactories and jsonFactories are empty.
+// IsEmpty: Returns a bool indicating whether both defaultFactories and jsonFactories are empty.
```
rbergen, line 71: I'll note just for the record that in this particular bit of documentation "boolean" was used to refer to the "data type that can either be true or false", which is indeed called `bool` in C++. It would be the same as writing I'm returning an integer, while the actual C++ data type is `int`, `int8_t`, `int16_t`, `int32_t` or even one of its unsigned variations.

The same applies for the use of the word boolean in the documentation in ledstripeffect.h.

At the same time, I don't care anywhere enough about this to make a change request out of it.

rbergen (approved):
LGTM

PR #857, opened 2026-04-26
Dynamic LED Settings, WS2812B output driver

davepl:
## Description

Random pre-note: the biggest user-facing change in this whole build is that I increased VU decay rate and it makes the VU meter and related effects a lot more reactive.  I quite like the change, curious on your thoughts.

This check-in moves the project toward a runtime-configurable settings model for WS2812B builds, while keeping the compile-time limits explicit and safe.  There’s nothing I can do about HUB75 for now, so if you build a mesmerizer, you can’t drive WS2812B with it unless you recompile.

The main architectural change is that we now distinguish between:

compiled capabilities: driver type, maximum LED capacity, maximum channel count, board-specific defaults
active runtime configuration: matrix width/height, serpentine layout, WS2812B channel count, output pins, and selected audio input pin

Why I did this:

The old settings path was great for simple scalar values like brightness, but it was the wrong shape for topology and output configuration. Those settings are interdependent: width/height affect LED count, output pins depend on channel count, and some changes are live while others require reboot or even rebuild. Extending the old flat form-post model further would have made the code harder to reason about and easier to misuse. The new model makes those boundaries explicit.

Major changes

All of the old settings code lives on as Is, really, with additions:

Unified settings model and schema
I added a structured /api/v1/settings plus /api/v1/settings/schema alongside the legacy /settings endpoints.

The unified API groups settings logically into:
device
topology
outputs
effects

The schema reports what the firmware can actually do:

compiled limits
current runtime values
whether a setting is live-applied, reboot-applied, or rejected as recompile needed

I kept the legacy endpoints for compatibility, but the new unified API is the intended long-term contract if we agree.  The current UI uses the old api, but I’m not sure about other consumers.  Either way, it’s backward compatible.

Standardized validation for nontrivial settings

I added DeviceConfig validators for the constrained settings and made both the legacy and unified APIs use them consistently before applying changes.

This matters because topology/output/audio settings should be validated as a coherent request rather than scattered setter calls. The code now rejects invalid combinations up front instead of partially applying the config and discovering a failure later.

In practical terms:

brightness
power limit
audio input pin
topology
output driver
WS2812B channel/pin config
full runtime output/topology config
all now follow the same validation pattern.

Runtime topology for WS2812B within compiled capacity

For WS2812B builds, matrix topology is now runtime-configurable within the compiled LED budget.

Important detail:

We no longer treat compile-time MATRIX_WIDTH and MATRIX_HEIGHT as hard runtime bounds for strip builds. Instead, the important ceiling is compiled LED capacity (NUM_LEDS).

So if the firmware was built with LED headroom, runtime width/height may change freely as long as:

width * height <= compiled max LED count

That gives us the flexibility we actually want for strip matrices without pretending HUB75 is equally dynamic.

Runtime WS2812B output manager

This is the other major change.

Previously, WS2812B hardware registration was tied directly to the templated FastLED transport setup. That made the LED count easy to vary, but channel count and pin selection were effectively baked into the build.

The new WS2812B output path keeps FastLED where it is strong:

pixel math
palettes
effects
CRGB handling
...but moves hardware transport decisions behind a runtime output manager.

What this gives us for WS2812B builds:

active channel count can be changed at runtime within compiled max channels
channel GPIO pins can be changed at runtime
logical matrix size and serpentine mapping can change at runtime
active LED count follows runtime topology
buffers, stats, and socket sizing now track the active runtime size

This is the key reason for the refactor: users can reconfigure strip topology/output without rebuilding firmware, as long as they stay within the compiled envelope.  My goal was you could program an ESP32 and then configure it separate from programming, making changes through the web config, etc.

Runtime config logging

We added a runtime config banner so it is obvious at boot and at change time what the device thinks its active configuration is:
driver
matrix width/height
active LED count
serpentine
channel count
audio input pin
WS2812B pins

This makes bring-up and field debugging much easier.

Audio input pin setting

Audio input pin is now part of persisted/reportable config and is validated like the other nontrivial settings.

Current limitation:
it is boot-applied, not live-recreated yet.

The reason is not that it is impossible, but that audio is still owned by a long-running analyzer/task lifecycle that has not yet been refactored into a clean stop/recreate/start model.  Basically there’s a global singleton, and it would need to be reallocated dynamically, but I had to draw the line somewhere.

HUB75 remains intentionally strict

I specifically did not pretend SmartMatrix is runtime-geometry-safe.  It is not.  Whether its needed for perf or not I can’t say, but a lot of the matrix geometry and characteristics get baked in at compile time.

For HUB75:

topology and output settings are still visible through the new settings model
unsupported geometry/output changes are rejected with "recompile needed”

That is deliberate. SmartMatrix still has compile-time geometry assumptions deep in its refresh/DMA path, so exposing “runtime geometry” there would be misleading and unsafe.

Net result

For WS2812B builds, this checkin gives us real live settings support for topology and outputs inside compiled limits, backed by a cleaner API, explicit validation, and much better observability.
For HUB75 builds, it gives us a cleaner settings architecture without claiming runtime capabilities the driver does not truly have.

This is primarily an architecture check-in: it makes future settings work safer and more coherent, and it removes a lot of hidden assumptions from the strip path.

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -32,6 +32,7 @@
 
 #include "globals.h"
 
+#include <array>
 #include <tuple>
 #include <vector>
 
@@ -124,6 +125,39 @@
 
 class DeviceConfig : public IJSONSerializable
 {
+  public:
+    enum class OutputDriver : uint8_t
+    {
+        WS281x,
+        HUB75
+    };
+
+    struct RuntimeTopology
+    {
+        uint16_t width = MATRIX_WIDTH;
+        uint16_t height = MATRIX_HEIGHT;
+        bool serpentine = true;
+    };
+
+    struct RuntimeOutputs
+    {
+        OutputDriver driver =
+        #if USE_HUB75
+            OutputDriver::HUB75;
+        #else
+            OutputDriver::WS281x;
+        #endif
+        size_t channelCount = NUM_CHANNELS;
+        std::array<int8_t, NUM_CHANNELS> outputPins{};
+    };
+
+    struct RuntimeConfig
+    {
+        RuntimeTopology topology;
+        RuntimeOutputs outputs;
+    };
+
+  private:
     // Add variables for additional settings to this list
     String  hostname = cszHostname;
     String  location = cszLocation;
@@ -141,12 +175,20 @@ class DeviceConfig : public IJSONSerializable
     CRGB    globalColor = CRGB::Red;
     bool    applyGlobalColors = false;
     CRGB    secondColor = CRGB::Red;
+    int8_t  audioInputPin = INPUT_PIN;
+    RuntimeTopology runtimeTopology = {};
+    RuntimeOutputs runtimeOutputs = {};
 
     std::vector<SettingSpec, psram_allocator<SettingSpec>> settingSpecs;
     std::vector<std::reference_wrapper<SettingSpec>> settingSpecReferences;
     size_t writerIndex;
 
     void SaveToJSON() const;
+    bool SetTimeZoneInternal(const String& newTimeZone, bool skipWrite);
+    static std::array<int8_t, NUM_CHANNELS> GetCompiledWS281xPins();
+    static const char* DriverName(OutputDriver driver);
+    static bool IsHub75Build();
+    void LogRuntimeConfig(const char* reason) const;
 
     template <typename T>
     void SetAndSave(T& target, const T& source)
@@ -191,6 +233,13 @@ class DeviceConfig : public IJSONSerializable
     static constexpr const char * GlobalColorTag = NAME_OF(globalColor);
     static constexpr const char * ApplyGlobalColorsTag = NAME_OF(applyGlobalColors);
     static constexpr const char * SecondColorTag = NAME_OF(secondColor);
+    static constexpr const char * MatrixWidthTag = "matrixWidth";
+    static constexpr const char * MatrixHeightTag = "matrixHeight";
+    static constexpr const char * MatrixSerpentineTag = "matrixSerpentine";
+    static constexpr const char * OutputDriverTag = "outputDriver";
+    static constexpr const char * WS281xChannelCountTag = "ws281xChannelCount";
+    static constexpr const char * WS281xPinsTag = "ws281xPins";
+    static constexpr const char * AudioInputPinTag = "audioInputPin";
 
     DeviceConfig();
 
@@ -236,6 +285,7 @@ class DeviceConfig : public IJSONSerializable
     void SetRememberCurrentEffect(bool newRememberCurrentEffect);
 
     uint8_t GetBrightness() const { return brightness; }
+    static ValidateResponse ValidateBrightness(int newBrightness);
     static ValidateResponse ValidateBrightness(const String& newBrightness);
     void SetBrightness(int newBrightness);
 
@@ -243,6 +293,7 @@ class DeviceConfig : public IJSONSerializable
     void SetShowVUMeter(bool newShowVUMeter);
 
     int GetPowerLimit() const { return powerLimit; }
+    static ValidateResponse ValidatePowerLimit(int newPowerLimit);
     static ValidateResponse ValidatePowerLimit(const String& newPowerLimit);
     void SetPowerLimit(int newPowerLimit);
 
@@ -257,4 +308,73 @@ class DeviceConfig : public IJSONSerializable
 
     void SetColorSettings(const CRGB& globalColor, const CRGB& secondColor);
     void ApplyColorSettings(std::optional<CRGB> globalColor, std::optional<CRGB> secondColor, bool clearGlobalColor, bool applyGlobalColor);
+
+    static constexpr uint16_t GetCompiledMatrixWidth() { return MATRIX_WIDTH; }
+    static constexpr uint16_t GetCompiledMatrixHeight() { return MATRIX_HEIGHT; }
+    static constexpr size_t GetCompiledLEDCount() { return NUM_LEDS; }
+    static constexpr size_t GetCompiledChannelCount() { return NUM_CHANNELS; }
+    static constexpr int GetCompiledAudioInputPin() { return INPUT_PIN; }
+    static std::array<int8_t, NUM_CHANNELS> GetCompiledPins() { return GetCompiledWS281xPins(); }
+    static OutputDriver GetCompiledOutputDriver()
+    {
+        #if USE_HUB75
+            return OutputDriver::HUB75;
+        #else
+            return OutputDriver::WS281x;
+        #endif
+    }
+
+    RuntimeConfig GetRuntimeConfig() const { return RuntimeConfig{runtimeTopology, runtimeOutputs}; }
+    const RuntimeTopology& GetTopology() const { return runtimeTopology; }
+    const RuntimeOutputs& GetOutputs() const { return runtimeOutputs; }
+    uint16_t GetMatrixWidth() const { return runtimeTopology.width; }
+    uint16_t GetMatrixHeight() const { return runtimeTopology.height; }
+    bool IsMatrixSerpentine() const { return runtimeTopology.serpentine; }
+    size_t GetActiveLEDCount() const { return static_cast<size_t>(runtimeTopology.width) * runtimeTopology.height; }
+    int GetAudioInputPin() const { return audioInputPin; }
+    OutputDriver GetOutputDriver() const { return runtimeOutputs.driver; }
+    size_t GetChannelCount() const { return runtimeOutputs.channelCount; }
+    const std::array<int8_t, NUM_CHANNELS>& GetWS281xPins() const { return runtimeOutputs.outputPins; }
+    // When runtime output settings still match the compiled WS281x transport, we can keep using the
+    // long-proven FastLED controller path and reserve the new runtime manager for true pin/count changes.
+    bool UsesCompiledWS281xTransport() const
+    {
+        return runtimeOutputs.driver == GetCompiledOutputDriver()
+            && runtimeOutputs.channelCount == GetCompiledChannelCount()
+            && runtimeOutputs.outputPins == GetCompiledPins();
+    }
+    bool SupportsLiveTopology() const { return !IsHub75Build() && runtimeOutputs.driver == OutputDriver::WS281x; }
+    bool SupportsLiveOutputReconfigure() const { return !IsHub75Build() && runtimeOutputs.driver == OutputDriver::WS281x; }
+    bool SupportsConfigurableAudioInputPin() const
+    {
+        #if ENABLE_AUDIO && !USE_M5 && (USE_I2S_AUDIO || ELECROW)
+            return true;
+        #else
+            return false;
+        #endif
+    }
+    bool SupportsLiveAudioInputReconfigure() const { return false; }
+    String GetAudioInputModeName() const
+    {
+        #if !ENABLE_AUDIO
+            return "disabled";
+        #elif USE_M5
+            return "m5_internal";
+        #elif USE_I2S_AUDIO || ELECROW
+            return "i2s";
+        #else
+            return "adc_fixed";
+        #endif
+    }
+    bool RequiresRecompileForCurrentRuntimeConfig() const { return runtimeOutputs.driver != GetCompiledOutputDriver(); }
+    String GetCompiledDriverName() const { return DriverName(GetCompiledOutputDriver()); }
+    String GetRuntimeDriverName() const { return DriverName(runtimeOutputs.driver); }
+
+    ValidateResponse ValidateAudioInputPin(int pin) const;
+    ValidateResponse ValidateTopology(uint16_t width, uint16_t height, bool serpentine) const;
+    ValidateResponse ValidateOutputDriver(OutputDriver driver) const;
+    ValidateResponse ValidateWS281xSettings(size_t channelCount, const std::array<int8_t, NUM_CHANNELS>& pins) const;
+    ValidateResponse ValidateRuntimeConfig(const RuntimeConfig& config) const;
+    bool SetRuntimeConfig(const RuntimeConfig& config, bool skipWrite = false, String* errorMessage = nullptr);
+    void SetAudioInputPin(int newAudioInputPin);
 };
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -175,7 +175,7 @@ class DeviceConfig : public IJSONSerializable
     CRGB    globalColor = CRGB::Red;
     bool    applyGlobalColors = false;
     CRGB    secondColor = CRGB::Red;
-    int8_t  audioInputPin = INPUT_PIN;
+    int8_t  audioInputPin = AUDIO_INPUT_PIN;
     RuntimeTopology runtimeTopology = {};
     RuntimeOutputs runtimeOutputs = {};
 
@@ -313,7 +313,7 @@ class DeviceConfig : public IJSONSerializable
     static constexpr uint16_t GetCompiledMatrixHeight() { return MATRIX_HEIGHT; }
     static constexpr size_t GetCompiledLEDCount() { return NUM_LEDS; }
     static constexpr size_t GetCompiledChannelCount() { return NUM_CHANNELS; }
-    static constexpr int GetCompiledAudioInputPin() { return INPUT_PIN; }
+    static constexpr int GetCompiledAudioInputPin() { return AUDIO_INPUT_PIN; }
     static std::array<int8_t, NUM_CHANNELS> GetCompiledPins() { return GetCompiledWS281xPins(); }
     static OutputDriver GetCompiledOutputDriver()
     {
--- a/include/gfxbase.h
+++ b/include/gfxbase.h
@@ -116,6 +116,8 @@ constexpr static inline uint8_t WU_WEIGHT(uint8_t a, uint8_t b)
 
 class Boid;
 
+uint16_t XY(uint16_t x, uint16_t y);
+
 class GFXBase : public Adafruit_GFX
 {
 #if USE_NOISE
@@ -129,6 +131,7 @@ protected:
     size_t _width;
     size_t _height;
     size_t _ledcount;
+    bool _serpentine = true;
 
     // 32 Entries in the 5-bit gamma table
     static const uint8_t gamma5[32];
@@ -185,13 +188,17 @@ public:
     virtual ~GFXBase() override;
 
     #if USE_NOISE
+    void EnsureNoise();
+
     Noise &GetNoise()
     {
+        EnsureNoise();
         return *_ptrNoise;
     }
 
     const Noise &GetNoise() const
     {
+        const_cast<GFXBase*>(this)->EnsureNoise();
         return *_ptrNoise;
     }
     #endif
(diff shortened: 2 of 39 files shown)


davepl:
@copilot resolve the merge conflicts in this pull request

include/deviceconfig.h
```diff
@@ -124,6 +125,39 @@
...
+    {
+        WS281x,
+        HUB75
+    };
+
+    struct RuntimeTopology
+    {
+        uint16_t width = MATRIX_WIDTH;
+        uint16_t height = MATRIX_HEIGHT;
+        bool serpentine = true;
```
robertlipe, line 139: Before we freeze V1, is this the place to think about orientation?

rotate 180 is handy with a mesmerizer when you want the cables on the same side as the Macbook's. (I know. First worls problem. Welcome to my world.)

it's also a common request in strips to flip them end for end because the power cable dictated the cable orientation.

(If this is the place for it, please go ahead and add it. If you don't want to sling the bits, I will...after this lands.)

include/deviceconfig.h
```diff
@@ -124,6 +125,39 @@
...
+  public:
+    enum class OutputDriver : uint8_t
+    {
+        WS281x,
+        HUB75
+    };
+
+    struct RuntimeTopology
+    {
+        uint16_t width = MATRIX_WIDTH;
```
robertlipe, line 137: It sounds nutty, but make these both uint32s everywhere.

Both RISC-V and Extensa will do lots of goofy loading and making to ensure that uint16s wrap (which we almost never intentionally do) in the 'right' way, even though everything is a register.

Unlike x86, register are 32-bits and working with anything smaller than that where performance counts is just throwing away clock cycles.

(It's not premature. It matters and I've shown disassembly in this project where it matters.)

include/deviceconfig.h
```diff
@@ -141,12 +175,20 @@ class DeviceConfig : public IJSONSerializable
     CRGB    globalColor = CRGB::Red;
     bool    applyGlobalColors = false;
     CRGB    secondColor = CRGB::Red;
+    int8_t  audioInputPin = INPUT_PIN;
```
robertlipe: Please called this analogAudioInputPin.

Let's make space for 
I2S_BCLK_PIN
I2S_WS_PIN
I2S_DATA_PIN

in the same way. I extracted those from your hard-coded constants a year or so ago that were buried in soundanalyzer and we have platforms setting these in platformio.ini these days. (With more on the way.)

include/deviceconfig.h
```diff
@@ -191,6 +233,13 @@ class DeviceConfig : public IJSONSerializable
     static constexpr const char * GlobalColorTag = NAME_OF(globalColor);
     static constexpr const char * ApplyGlobalColorsTag = NAME_OF(applyGlobalColors);
     static constexpr const char * SecondColorTag = NAME_OF(secondColor);
+    static constexpr const char * MatrixWidthTag = "matrixWidth";
```
robertlipe, line 236: For all these that are constants, prefer
char foo[] = over char *.

1) Less memmory. the symbol IS at the address of the data, not a pointer TO the data.
2) one else indiretion at runtime.
3) sizeof() works like God Herself intended.

Isn't the rule that Classes are ClassCased and data are snake_case or camel cased? It's weird to mix them in the same structure like this.

include/deviceconfig.h
```diff
@@ -236,13 +285,15 @@ class DeviceConfig : public IJSONSerializable
...
 
     uint8_t GetBrightness() const { return brightness; }
+    static ValidateResponse ValidateBrightness(int newBrightness);
     static ValidateResponse ValidateBrightness(const String& newBrightness);
     void SetBrightness(int newBrightness);
 
     bool ShowVUMeter() const { return showVUMeter; }
     void SetShowVUMeter(bool newShowVUMeter);
 
     int GetPowerLimit() const { return powerLimit; }
```
robertlipe, line 295: Do negative values ever make sense?

(I'll preload any answer but 'no' with std::optional)

include/deviceconfig.h
```diff
@@ -257,4 +308,73 @@ class DeviceConfig : public IJSONSerializable
 
     void SetColorSettings(const CRGB& globalColor, const CRGB& secondColor);
     void ApplyColorSettings(std::optional<CRGB> globalColor, std::optional<CRGB> secondColor, bool clearGlobalColor, bool applyGlobalColor);
+
+    static constexpr uint16_t GetCompiledMatrixWidth() { return MATRIX_WIDTH; }
```
robertlipe, line 312: uint32_t's or uint_fast32_t - which might be a 64-bit for our upcoming 64-bit version. (That's not crazy talk...)

include/deviceconfig.h
```diff
@@ -257,4 +308,73 @@ class DeviceConfig : public IJSONSerializable
...
+    bool RequiresRecompileForCurrentRuntimeConfig() const { return runtimeOutputs.driver != GetCompiledOutputDriver(); }
+    String GetCompiledDriverName() const { return DriverName(GetCompiledOutputDriver()); }
+    String GetRuntimeDriverName() const { return DriverName(runtimeOutputs.driver); }
+
+    ValidateResponse ValidateAudioInputPin(int pin) const;
+    ValidateResponse ValidateTopology(uint16_t width, uint16_t height, bool serpentine) const;
+    ValidateResponse ValidateOutputDriver(OutputDriver driver) const;
+    ValidateResponse ValidateWS281xSettings(size_t channelCount, const std::array<int8_t, NUM_CHANNELS>& pins) const;
+    ValidateResponse ValidateRuntimeConfig(const RuntimeConfig& config) const;
+    bool SetRuntimeConfig(const RuntimeConfig& config, bool skipWrite = false, String* errorMessage = nullptr);
```
robertlipe, line 378: Does SetRuntimeConfig write to errorMessage? Then that String* should be a const String&

include/gfxbase.h
```diff
@@ -116,6 +116,8 @@ constexpr static inline uint8_t WU_WEIGHT(uint8_t a, uint8_t b)
 
 class Boid;
 
+uint16_t XY(uint16_t x, uint16_t y);
```
robertlipe, line 119: As above.

include/gfxbase.h
```diff
@@ -206,6 +213,23 @@ class GFXBase : public Adafruit_GFX
         return _ledcount;
     }
 
+    virtual size_t GetMatrixWidth() const
```
robertlipe, line 216: ...and now we've size-extended. Let's just keep them unsigned all the way up and down the stack.

(I know we have a mess of inconsistencies in this area. This s about setting precedent, right?)

src/deviceconfig.cpp
```diff
@@ -42,6 +45,38 @@
...
+        #if NUM_CHANNELS >= 5
+        LED_PIN4,
+        #endif
+        #if NUM_CHANNELS >= 6
+        LED_PIN5,
+        #endif
+        #if NUM_CHANNELS >= 7
+        LED_PIN6,
+        #endif
+        #if NUM_CHANNELS >= 8
```
robertlipe, line 74: If you don't extend this to 16, I'll have to. (I already have in one of my forks.) Please just do it while this is open.

src/deviceconfig.cpp
```diff
@@ -148,19 +245,61 @@ bool DeviceConfig::DeserializeFromJSON(const JsonObjectConst& jsonObject, bool s
     SetIfPresentIn(jsonObject, rememberCurrentEffect, RememberCurrentEffectTag);
     SetIfPresentIn(jsonObject, powerLimit, PowerLimitTag);
     SetIfPresentIn(jsonObject, brightness, BrightnessTag);
+    // Persisted config predates the newer brightness guardrails in some installs, so treat an invalid
+    // saved brightness as "unset" and fall back to the normal 100% default instead of booting dark.
+    if (brightness < BRIGHTNESS_MIN || brightness > BRIGHTNESS_MAX)
```
robertlipe, line 250: brightness = min(brightness, BRIGHTNESS_MIN, BRIGHTNESS_MAX)

we have min opcode on xtensa if we make our intent clear.

src/deviceconfig.cpp
```diff
@@ -148,19 +245,61 @@ bool DeviceConfig::DeserializeFromJSON(const JsonObjectConst& jsonObject, bool s
...
+    SetIfPresentIn(jsonObject, updated.topology.height, MatrixHeightTag);
+    SetIfPresentIn(jsonObject, updated.topology.serpentine, MatrixSerpentineTag);
+
+    if (jsonObject[OutputDriverTag].is<String>())
+    {
+        const auto driverName = jsonObject[OutputDriverTag].as<String>();
+        if (driverName == DriverName(OutputDriver::HUB75))
+            updated.outputs.driver = OutputDriver::HUB75;
+        else if (driverName == DriverName(OutputDriver::WS281x))
+            updated.outputs.driver = OutputDriver::WS281x;
```
robertlipe, line 278: else debugE(invalid driverName)

src/network.cpp
```diff
@@ -529,7 +529,7 @@ namespace nd_network
                 [](const DebugCLI::cli_argv &) { NTPTimeClient::UpdateClockFromWeb(&l_Udp); }
             },
             #endif
-            { "stats", "Display system statistics", "Displaying statistics",
+            {"stats", "Display stats", "Displaying",
```
robertlipe: Grrr. We have different KINDS of stats.

src/remotecontrol.cpp
```diff
@@ -218,15 +218,27 @@ static const RemoteColorCode RemoteColorCodes[] =
...
-#define NEC_DECODE_MARGIN 200  // Tolerance in microseconds
+#define NEC_DECODE_MARGIN 300  // Tolerance in microseconds
 #define RMT_RESOLUTION_HZ 1000000
 
+namespace
+{
+    #if USE_WS281X
+    // Reserve legacy RMT channel 7 for IR receive on WS281x builds. LED output uses channels from 0 upward,
+    // so the only impossible case is asking for eight LED channels and a remote at the same time.
+    static_assert(NUM_CHANNELS < 8, "WS281x plus remote exceeds available legacy ESP32 RMT channels");
```
robertlipe, line 229: The number of RMT channels depends on the hardware. S3, for example, has 4 instead 8. Some have as few as 2. I'm pretty sure there's an ESP-IDF API for this.

src/remotecontrol.cpp
```diff
@@ -297,11 +316,13 @@ class RemoteControlImpl
         if (symbol_count < 67) return false;
 
         uint32_t data = 0;
-        int bit_idx = 0;
-
         for (int i = 2; i < 66; i += 2) {
             if (!match(get_time(i), 560)) return false;
 
+            // Keep the historical MSB-first assembly here because the existing command tables were
```
robertlipe, line 322: That's exactly the stated reason when I submitted it. :-)

src/soundanalyzer.cpp
```diff
@@ -432,7 +448,8 @@ void SoundAnalyzerBase::InitI2S_Modern()
 void SoundAnalyzerBase::InitI2S_Legacy()
 {
 #if (USE_I2S_AUDIO || ELECROW) && !IS_IDF5
-    debugI("Audio: Initializing I2S Digital Mic (Legacy) on BCLK:%d WS:%d DIN:%d", I2S_BCLK_PIN, I2S_WS_PIN, INPUT_PIN);
+    const auto inputPin = GetConfiguredAudioInputPin();
```
robertlipe, line 700: Please change the ESP-IDF5 "Modern()" functions in this case file. You can use spectrum2_v3 for a test case.

src/systemcontainer.cpp
```diff
@@ -273,6 +287,67 @@ void SystemContainer::SetupConfig()
         _ptrDeviceConfig = make_unique_psram<DeviceConfig>();
 }
 
+int SystemContainer::GetConfiguredAudioInputPin() const
```
robertlipe, line 290: Should this be unified to return an I2S structure?

src/webserver.cpp
```diff
@@ -550,19 +892,155 @@ void CWebServer::SetSettingsIfPresent(AsyncWebServerRequest * pRequest)
...
+    }
 
     // We return the current config in response
     GetSettings(pRequest);
 }
 
+void CWebServer::SetUnifiedSettings(AsyncWebServerRequest * pRequest, JsonVariantConst json)
+{
+    if (!json.is<JsonObjectConst>())
+    {
```
robertlipe, line 933: Please add a debugE() hre...and generally anywhere error-ish.

src/ws281xgfx.cpp
```diff
@@ -30,68 +30,102 @@
 
 #include "globals.h"
 
+#include <algorithm>
```
robertlipe, line 33: Does this pass tools/audit_includes?

src/ws281xgfx.cpp
```diff
@@ -30,68 +30,102 @@
...
+        #endif
+        #if NUM_CHANNELS >= 6 && LED_PIN5 >= 0
+            ADD_CHANNEL(5);
+        #endif
+        #if NUM_CHANNELS >= 7 && LED_PIN6 >= 0
+            ADD_CHANNEL(6);
+        #endif
+        #if NUM_CHANNELS >= 8 && LED_PIN7 >= 0
+            ADD_CHANNEL(7);
+        #endif
```
robertlipe, line 121: As above for LED_PIN...15, please

src/ws281xgfx.cpp
```diff
@@ -108,20 +142,66 @@ WS281xGFX::~WS281xGFX()
...
+    RebindCompiledLEDs(devices);
+    LogWS281xConfiguration(deviceConfig, devices, true, reason);
+}
+
+void WS281xGFX::ConfigureTopology(size_t width, size_t height, bool serpentine)
+{
+    const auto newLEDCount = width * height;
+    if (newLEDCount != GetLEDCount())
+    {
+        auto* newPixels = static_cast<CRGB*>(calloc(newLEDCount, sizeof(CRGB)));
```
robertlipe, line 156: Can we just use new?

src/ws281xoutputmanager.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+WS281xOutputManager::~WS281xOutputManager()
+{
+    for (size_t i = 0; i < _channels.size(); ++i)
+        ReleaseChannel(i);
+}
+
+bool WS281xOutputManager::RecreateChannel(size_t channelIndex, int8_t pin, size_t ledCount, String* errorMessage)
```
robertlipe, line 76: I think we're going to regret recreating this wheel. There's just a LOT of real-world rubber meeting the ground here.

I'm with you on voting this part of FastLED off the islance since the channel API is a dead end for us, but consider backing it with

https://components.espressif.com/components/espressif/led_strip/versions/2.5.4/examples/led_strip_spi_ws2812?language=en

tools/test_topology.py
```diff
@@ -0,0 +1,340 @@
+#!/usr/bin/env python3
+
+"""
+Smoke-test WS281x runtime topology against a live NightDriverStrip device.
```
robertlipe, line 4: Nice!

platformio.ini
```diff
@@ -19,7 +19,7 @@
 ; mesmerizer    HUB75 info panel with audio effects, weather, info, etc
 
 [platformio]
-default_envs =
+default_envs = mesmerizer
```
robertlipe: no

platformio.ini
```diff
@@ -30,10 +30,10 @@ extra_configs =
 ; Options that are used (or extended) in all device sections (and hence environments) are defined here
 
 [base]
-upload_port     =
-monitor_port    =
-build_type      = release
-upload_speed    = 2000000
+upload_port     = /dev/cu.usbserial-71112101
```
robertlipe: No.

Pro Tip: just use little scripts tha set the arguments for YOUR system and/or setenv PLATFORMIO_UPLOAD_PORT and friends and you can quit changing these things.

robertlipe (commented):
I have to run, but here's a start.

I'm looking forward to this. Exciting times.

davepl:
I’ll work through your list, but what supports more than 8 channels?  How could you have LED_PIN15?

Thanks!
Dave

include/deviceconfig.h
davepl: I agree it should be in here - please add it when we have support for rotation... or does SmartMatrix already give us that?

include/deviceconfig.h
davepl: Done on input pins, others should be done separately please!

include/deviceconfig.h
davepl: Perhaps, but not in this checkin

rbergen:
I'm just noting for now that CI on demo_v3 currently fails.

robertlipe:
On Sun, Apr 26, 2026 at 12:16 PM David W Plummer ***@***.***>
wrote:

> *davepl* left a comment (PlummersSoftwareLLC/NightDriverStrip#857)
> <https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/857#issuecomment-4322566868>
> I’ll work through your list, but what supports more than 8 channels? How
> could you have LED_PIN15


Lightcrafter16, which I've mentioned at least three times this week, though
I see I called it "LightDriver16" two of those times—is clearly the same
board. (sorry) I've been working with the developer.

https://moonmodules.org/MoonLight/moonbase/inputoutput/#lightcrafter16

I have one here. The board is real, and as mentioned, I have a fork of
Nightdriverled currently running on it, though I haven't pushed it because
it's irresponsible with the 80A (!) it can manage, and I don't want to
promote code that doesn't manage power states responsibly.

Beyond 16 channels, unique traits (that stretch NightDriver in new and
interesting ways) include on-board current sensors, thermometers, a need
for rotation, which I've implemented as these are often used to drive
reasonably short row of strips to provide HUB75-style displays, on rows of
shorter, brighter WS281x), and on-board W5500, and on-board FETs for
high-side switching of strips. It stretches things like RMT not working
because you have FOUR RMT channels and 16 strip channels; you either have
to re-arm them per frame or just use SPI/I2S.

That was the board on which I tested/fixed the new IR remote, because
original builds uniquely failed on S3.

I'm not asking you to develop support for the board. I'm asking you to
remember that it exists and to add bracing for boards like it in any code
you touch.

robertlipe:
https://github.com/pixelmatix/SmartMatrix/blob/af288cb536bd934d09be309d2346f9a8ffb51f4e/src/Layer.cpp#L26

platformio.ini
davepl: Mai ouis.

platformio.ini
davepl: I see the custom_*.ini but don't know what it does - do I just create a file called

custom_dave.ini

...and put my settings in there, or how does one name the file so its found and used?

robertlipe: yes. I end up with a dozen+ custom_$PROJ files: custom_lc16.ini, custom_tinyled.ini, custom_waveshare_s3_147.ini, custom_yd.ini, etc. for things that I, uniquely, care about but haven't tried to upstream because I don't want to impact build times for everyone else. See recent post on that very topic.

I never code upload_port because it's different on every combination of host and physical port, hub port choice, etc. It's why I have tiny scripts, like go-mes which also end up being notes about that project. This builds and runs mesmerizer and encodes the dumb commands I need 1% of the time that I'll otherwise forget.

As Linux user might replace ioreg with something someting /sys/class/tty.



```
#!/bin/sh

set -e

if [ ! -c $PLATFORMIO_UPLOAD_PORT ]
then
  export PLATFORMIO_UPLOAD_PORT=$(ioreg -l | awk -F\" '/IOCalloutDevice.*usbserial/ {P=$4}; END{print P}')
fi


E=mesmerizer

#~/.platformio/penv/bin/pio run  -e $E
#~/.platformio/penv/bin/pio run --target buildfs -e $E

pio run    --target upload -e $E
# pio device monitor -e mesmerizer
tio -a latest

```

rbergen:
The approach to fix the CI failure merged the key delta between #858 and this PR into this one, which makes treating them separately a moot point. I'm going to close this and move on to #858 - I've now basically reviewed #858 in its entirely already anyway because the beat detection stuff is now embedded in this PR.

PR #858, opened 2026-04-26
Newbeat

davepl:
## Description
Adds actually useful/accurate beat detection.  Updates the Stars effect to use it.  Adds a blue "near miss" and red "beat detect" in the center of the VU meter.  Not permanent, necessarily.

* [X] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [X] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [X] I selected `main` as the target branch.
* [X] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #859, opened 2026-04-28
Localui

davepl:
## Description
My two changes are at opposite ends of the abstraction spectrum - one is the new UI and plumbing, the other is a custom RMT driver.  I couldn’t get FastLED to play nice.  You can add controllers, but never remove one, it seems, so runtime config is challenging.  I really want to be able to change the basics live so people can set it up without a refresh (or even reboot, but less important).

The current changes are doing two main things.

First, they turn WS281x output into a genuinely runtime-configurable transport:

live matrix width/height/serpentine changes within compiled LED capacity
live WS281x pin, channel count, and color-order changes
one owned runtime output path instead of switching between compile-time and runtime transport modes
validation so impossible pins/settings are rejected before they break output
effect reinitialization after topology changes so effects do not keep stale geometry assumptions
Second, they replace the old frontend with a static embedded administration UI:

no React / Node runtime in the firmware UI path
NDSCPP-style look and feel
unified Effects / Settings / Statistics experience
frame preview over websocket
more structured settings editing, better controls, and better runtime feedback

A major sub-goal of the most recent work has been to stop depending on FastLED for live WS281x transport. FastLED still provides CRGB, color helpers, and effect-side pixel handling, but the runtime LED transmission path is being moved under NightDriver control so pin/output reconfiguration is stable and does not depend on FastLED’s internal ESP32 RMT controller behavior.

So the short version is:

live device/output reconfiguration
safer runtime WS281x driver ownership
a new static admin UI that replaces the old embedded web app


robertlipe:
As I've said repeatedly, I really think that FastLED in this generation is
a losing bet. You're going to HAVE to either replace the business end of it
with something like the Espressif-maintained code I cited, NeoPixel, or
wait for FastLED 4 which has the channels UI that supports dynamic
configuration, but that doesn't support ISO C++ - see the patch that I've
mentioned a couple of times that was rejected out of hand by the FastLED
maintainer. The breadth of strips observed in the wild that almost support
this or that timing and the growing array of ESP32 hardware, each with
better, but different controllers is going to be a pit of time. (That I'm
not going to to spend.) This is legitimately a project of its own to do
well. Pulling out 'scopes to support 3-5PX (remember WW/CW) and 2805 and
SK6812/SK6815 timings or UCS2904's and APA102's (do you have a place to
configure the clock pin) is no fun. It's actually a difficult space. and,
in my strong opinion, one I recommend we continue to outsource.

I dumped probably a hundred hours on this problem last summer and concluded
that using FastLED for pixel blending and such (there's, like 3 files that
we really need) and replacing the very bottom layer is the only sane way
forward. Since then FastLED's 4 channel API brought hope, but the maintiner
rejecting fixes because "his AI didn't like it" just made me leery of
hitching our wagon more to this and I gave up.

Maybe YOU can take FastLED4 and produce fixes that Zach's favorite AI likes
better than ISO C++.

If I were a maintainer, I wouldn't you replace the very top-level UI and
the very bottom level LED in the same PR.

Still, Let me grab a laptop and look at the code.

include/effects/matrix/PatternWeather.h
```diff
@@ -100,7 +100,10 @@ extern const uint8_t thunderstorm_night_end[]       asm("_binary_assets_bmp_thun
 
 static constexpr auto pszDaysOfWeek = to_array( { "SUN", "MON", "TUE", "WED", "THU", "FRI", "SAT" } );
 
-static std::map<const String, EmbeddedFile, std::less<const String>, psram_allocator<std::pair<const String, EmbeddedFile>>> weatherIcons =
+// This lookup table is tiny and constructed during global initialization, before the system has
+// completed its normal startup path. Keep it on the regular heap so boot does not depend on PSRAM.
```
robertlipe, line 104: No need for a comment to explain why to NOT do it in the contrived (and wrong - this very fix is part of our SIOF issues and PR).

Please consider thas a part of a speparate PR.

include/effects/matrix/spectrumeffects.h
```diff
@@ -122,6 +122,54 @@ class InsulatorSpectrumEffect : public EffectWithId<InsulatorSpectrumEffect>, pu
...
+
+        return now <= _indicatorUntilMs;
+    }
+
+    void DrawBeatIndicator(std::vector<std::shared_ptr<GFXBase>> & GFX, int yVU) const
+    {
+        if (!UpdateBeatIndicatorState())
+            return;
+
+        const int centerLeft = (GFX[0]->width() / 2) - 1;
```
robertlipe, line 165: What if this strip is on GFX[1]? Why not just look at the gfx we're on?

Better yet, why not make all these beat functions into subscribers listening on event observers so they're not passing around all this dirty laundry?

platformio.ini
```diff
@@ -485,27 +496,30 @@ board_build.partitions = config/partitions_custom_noota_v3.csv
...
-[env:m5demo]
+
+[m5demo]
 extends         = dev_m5stick_c
 build_flags     = ${dev_m5stick_c.build_flags}
 build_src_flags = ${dev_m5stick_c.build_src_flags}
                   -DM5DEMO=1
                   -DEFFECTS_DEMO=1
+                  -DUSE_WS281X=1
                   -DPROJECT_NAME="\"M5Demo\""
```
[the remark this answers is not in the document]
davepl: Fixed

src/remotecontrol.cpp
```diff
@@ -268,6 +282,11 @@ class RemoteControlImpl
     RingbufHandle_t _ringbuf = NULL;
     bool _begun = false;
 
+    static void EdgeISR(void* arg)
+    {
+        (void)arg;
+    }
```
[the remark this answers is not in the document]
davepl: Fixed

davepl:
Well, you got your wish, it’s a ground-up-fresh transport.  The only thing I’m using from FastLED, I think, are compile-time constants for how long the pulses should be.

BTW, please keep the review specific to things IN THIS change.  If the code is consistent with the rest of the project, and you want to change a practice across the whole projects, open it as a separate issue please!

- Dave

include/effects/strip/fireworkseffect.h
```diff
@@ -0,0 +1,214 @@
...
+            particle.velocity = random_range(-_maxSpeed * speedScale, _maxSpeed * speedScale);
+            particle.age = 0.0f;
+            particle.preignitionTime = _particlePreignitionTime;
+            particle.ignitionTime = _particleIgnitionTime;
+            particle.holdTime = _particleHoldTime;
+            particle.fadeTime = _particleFadeTime;
+            particle.size = std::max(1.0f, burstSize * random_range(0.8f, 1.2f));
+            particle.drag = _particleDrag * random_range(0.85f, 1.15f);
+
+            _particles.push_back(particle);
```
robertlipe, line 121: Do a reserve and then
_particles.emplace_back();
auto& particle = _particles.back();
and you can construct this right in the copy inside the _particles vector

include/effects/strip/fireworkseffect.h
```diff
@@ -0,0 +1,214 @@
...
+
+        const size_t burstCount = std::clamp<size_t>(
+            1 + (beat.major ? 1U : 0U) + (beat.strength >= 2.1f ? 1U : 0U),
+            1,
+            3);
+        const size_t particleCount = std::clamp<size_t>(
+            static_cast<size_t>(_baseBurstParticles + beat.confidence * 10.0f + beat.strength * 6.0f + (beat.major ? 6.0f : 0.0f)),
+            4,
+            12);
+        const float speedScale = std::clamp(0.5f + beat.strength * 0.32f + beat.confidence * 0.18f, 0.25f, 1.75f);
```
robertlipe, line 183: Avoid all the magic constants sprinkled around.

include/deviceconfig.h
```diff
@@ -124,6 +125,50 @@
...
 class DeviceConfig : public IJSONSerializable
 {
+  public:
+    enum class OutputDriver : uint8_t
+    {
+        WS281x,
+        HUB75
+    };
+
+    enum class WS281xColorOrder : uint8_t
```
robertlipe, line 135: What about UCS2904, 2805, SK6812/SK6815, APA102 and all the other things that are now supported? (And working...)

include/deviceconfig.h
```diff
@@ -124,6 +125,50 @@
...
+    {
+        RGB,
```
robertlipe, line 137: What are your plans for WW/CW?

include/deviceconfig.h
```diff
@@ -191,6 +242,14 @@ class DeviceConfig : public IJSONSerializable
     static constexpr const char * GlobalColorTag = NAME_OF(globalColor);
     static constexpr const char * ApplyGlobalColorsTag = NAME_OF(applyGlobalColors);
     static constexpr const char * SecondColorTag = NAME_OF(secondColor);
+    static constexpr const char * MatrixWidthTag = "matrixWidth";
```
robertlipe, line 246: We've already had this "conversation". It's faster and takes less memory to make these char[].

Multiple, overlapping PRs are just frustrating for everyone.

include/deviceconfig.h
```diff
@@ -257,4 +318,76 @@ class DeviceConfig : public IJSONSerializable
...
+    {
+        #if USE_HUB75
+            return false;
+        #else
+            return true;
+        #endif
+    }
+    static constexpr size_t GetCompiledLEDCount() { return NUM_LEDS; }
+    static constexpr size_t GetCompiledChannelCount() { return NUM_CHANNELS; }
+    static constexpr int GetCompiledAudioInputPin() { return AUDIO_INPUT_PIN; }
```
robertlipe, line 335: Clarify as AnalogAudioInputPin as it doesn't handle I2S

include/deviceconfig.h
```diff
@@ -257,4 +318,76 @@ class DeviceConfig : public IJSONSerializable
...
+        #endif
+    }
+    bool SupportsLiveAudioInputReconfigure() const { return false; }
+    String GetAudioInputModeName() const
+    {
+        #if !ENABLE_AUDIO
+            return "disabled";
+        #elif USE_M5
+            return "m5_internal";
+        #elif USE_I2S_AUDIO || ELECROW
```
robertlipe, line 377: Isn't the goal to get rid of code like USE_I2S_AUDIO and make it runtime selectable?

include/gfxbase.h
```diff
@@ -147,7 +150,9 @@ class GFXBase : public Adafruit_GFX
     String _currentPaletteName;
 
     #if USE_NOISE
-        std::unique_ptr<Noise> _ptrNoise;
+        // I was this many years old when I learned about std::once
```
robertlipe, line 153: N/A

include/globals.h
```diff
@@ -693,20 +696,20 @@ extern const int g_aRingSizeTable[];
 // The M5 mic is on Pin34, but when I wire up my own microphone module I usually put it on pin 36.
 
 #if ENABLE_AUDIO
-    #ifndef INPUT_PIN
+    #ifndef AUDIO_INPUT_PIN
```
robertlipe, line 699: If you're going to rename it (which I strongly support), please put "ANALOG" in the name, e.g.AUDIO_ANALOG_PIN

include/interfaces.h
```diff
@@ -93,6 +93,23 @@ inline bool operator==(const psram_allocator<T>&, const psram_allocator<U>&) { r
 template <typename T, typename U>
 inline bool operator!=(const psram_allocator<T>&, const psram_allocator<U>&) { return false; }
 
+template<typename T>
+struct psram_array_deleter
```
robertlipe, line 98: Funny. I'm trying to delete psram_allocator because the system knows where memory comes from and you're adding to it. 

I'd much rather we fix our call to heap_caps_malloc_extmem_enable(), and use heap_caps_malloc(...MALLOC_CAP_8BIT) in the places that matter. We fix a lot of SIOF and general complexity that way.

include/soundanalyzer.h
```diff
@@ -527,6 +585,11 @@ class SoundAnalyzerBase : public ISoundAnalyzer
     void SampleAudio();
     void UpdateVU(float newval);
     void ComputeBandLayout();
+    void ResetFrameState();
+    void ResetBeatDetection();
+    void UpdateBeatDetection();
+    void RecordBeat(uint32_t now, float confidence, float strength, float bass, float mid, float treble, float flux, bool simulated);
```
robertlipe, line 591: With this many args, just consider passing around a struct so yo don't blow out the number of args in the calls which then reverts to a struct behind your back anyway.

src/deviceconfig.cpp
```diff
@@ -42,6 +45,61 @@
...
+        #endif
+        #if NUM_CHANNELS >= 6
+        LED_PIN5,
+        #endif
+        #if NUM_CHANNELS >= 7
+        LED_PIN6,
+        #endif
+        #if NUM_CHANNELS >= 8
+        LED_PIN7,
+        #endif
```
robertlipe, line 99: 16. Reminding me that much of this is part of antoher code review already in progress.

src/deviceconfig.cpp
```diff
@@ -66,8 +124,79 @@ void DeviceConfig::SaveToJSON() const
...
+DeviceConfig::WS281xColorOrder DeviceConfig::GetCompiledWS281xColorOrder()
+{
+    return kCompiledWS281xColorOrder;
+}
+
+String DeviceConfig::GetColorOrderName(WS281xColorOrder colorOrder)
+{
+    switch (colorOrder)
+    {
+        case WS281xColorOrder::RGB: return "RGB";
```
robertlipe, line 154: WW/CW

src/deviceconfig.cpp
```diff
@@ -420,6 +674,44 @@ void DeviceConfig::SetSecondColor(const CRGB& newSecondColor)
     SetAndSave(secondColor, newSecondColor);
 }
 
+DeviceConfig::ValidateResponse DeviceConfig::ValidateAudioInputPin(int pin) const
```
robertlipe, line 811: Isn't this std::expected throughout?

src/deviceconfig.cpp
```diff
@@ -558,3 +850,122 @@ void DeviceConfig::ApplyColorSettings(std::optional<CRGB> newGlobalColor, std::o
...
 }
+
+DeviceConfig::ValidateResponse DeviceConfig::ValidateTopology(uint16_t width, uint16_t height, bool serpentine) const
+{
+    if (width == 0 || height == 0)
+        return { false, "matrix dimensions must be greater than zero" };
+
+    // The strip path sizes its live buffers from total LED capacity, not the original compile-time aspect ratio.
+    // That keeps reshaping flexible while still refusing requests that would outgrow the compiled backing store.
+    if (static_cast<size_t>(width) * height > GetCompiledLEDCount())
```
robertlipe, line 995: If you're not using the bottom-end of FastLED, why keep ones of its worst traits, the compile-in LED limits at all?

If you're going to rip it out, make it all dynamic.

src/gfxbase.cpp
```diff
@@ -1377,34 +1377,63 @@ GFXBase::GFXBase(int w, int h) : Adafruit_GFX(w, h),
...
+        FillGetNoise();
+    });
+}
+#endif
 
 // Dirty hack to support FastLED, which calls out of band to get the pixel index for "the" array, without
 // any indication of which array or who's asking, so we assume the first matrix. If you have trouble with
 // more than one matrix and some FastLED functions like blur2d, this would be why.
-uint16_t XY(uint8_t x, uint8_t y)
+uint16_t XY(uint16_t x, uint16_t y)
```
robertlipe, line 1434: See other review.

src/main.cpp
```diff
@@ -490,9 +491,15 @@ void setup()
...
-        #if INPUT_PIN
-        if (INPUT_PIN >= 0)
-            pinMode(INPUT_PIN, INPUT);
+        // USE_M5 implies we are using M5Unified which manages the mic pins itself, so we 
+        // skip manual pin setup in that case.  For other boards, we set the audio input pin
+        // according to the current config, which allows the boot-applied pin to match what 
+        // settings report and for settings changes to take effect immediately.
+
+        #if !USE_M5
+        const auto audioInputPin = g_ptrSystem->GetConfiguredAudioInputPin();
```
robertlipe, line 500: We have almost this code somewhere else in this ver PR, I think.  Can we centralize ALL the sound initialization, perhaps into soudmanager's construction?

src/main.cpp
```diff
@@ -524,7 +531,9 @@ void setup()
     #elif USE_M5
 
         M5.begin();
-        // M5 specific setup is now inside M5Screen constructor
+        // M5Unified boots the panel in portrait. Set landscape before we size the Screen wrapper so
+        // the screen task and layout code agree on width/height from the start.
+        M5.Lcd.setRotation(1);
```
robertlipe, line 536: Can this hunk go in Screen::Screen?

src/remotecontrol.cpp
```diff
@@ -218,15 +218,27 @@ static const RemoteColorCode RemoteColorCodes[] =
...
 // ---------------------------------------------------------
 
-#define NEC_DECODE_MARGIN 200  // Tolerance in microseconds
+#define NEC_DECODE_MARGIN 300  // Tolerance in microseconds
 #define RMT_RESOLUTION_HZ 1000000
 
+namespace
+{
+    #if USE_WS281X
+    // Reserve legacy RMT channel 7 for IR receive on WS281x builds. LED output uses channels from 0 upward,
```
robertlipe, line 227: Except that IR doesnt' always GO on RMT 7 and there aren't always any 7 free IR channels for TX, as discussed in the other PR.

On S3, for example, half are only for TX and some are only for RX.

Newer parts have different rules

src/remotecontrol.cpp
```diff
@@ -240,6 +252,8 @@ class RemoteControlImpl
         config.rx_config.filter_ticks_thresh = 100; // Ignore pulses shorter than 100us
         config.rx_config.idle_threshold = 20000;    // 20ms idle = end of frame
 
+        attachInterruptArg(_pin, &RemoteControlImpl::EdgeISR, this, CHANGE);
```
robertlipe: This code hasn't needed an ISR so far. Stuff appears in buffers and we poll them when we care, eliminating a load of synchronization issues.

src/remotecontrol.cpp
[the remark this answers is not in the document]
robertlipe: Yeah, this.

src/soundanalyzer.cpp
```diff
@@ -210,10 +295,13 @@ float SoundAnalyzerBase::BeatEnhance(float amt)
...
-#if INPUT_PIN < 0
-    debugI("Audio: INPUT_PIN < 0, skipping hardware initialization. SimBeat only.");
-    return;
-#endif
+    const auto inputPin = GetConfiguredAudioInputPin();
+
+    if (inputPin < 0)
+    {
+        debugI("Audio: input pin < 0, skipping hardware initialization. SimBeat only.");
+        return;
```
robertlipe, line 303: Mauybe THIS is where that code to set input mode should go. Makes more sense than main.

src/soundanalyzer.cpp
```diff
@@ -287,6 +375,30 @@ void SoundAnalyzerBase::SetPeakDecayRates(float r1, float r2)
...
+
+    return _peak1Decay[band];
+}
+
+float SoundAnalyzerBase::Peak2Decay(int band) const
+{
+    if (band < 0 || band >= NUM_BANDS)
+        return 0.0f;
+
+    return _peak2Decay[band];
```
robertlipe, line 391: Isn't that a design error if called? I'd rather make that into an assert and fi the callers.

src/soundanalyzer.cpp
```diff
@@ -287,6 +375,30 @@ void SoundAnalyzerBase::SetPeakDecayRates(float r1, float r2)
...
+}
+
+unsigned long SoundAnalyzerBase::LastPeak1Time(int band) const
+{
+    if (band < 0 || band >= NUM_BANDS)
+        return 0;
```
robertlipe, line 397: Likewise.

src/soundanalyzer.cpp
```diff
@@ -296,10 +408,182 @@ void SoundAnalyzerBase::SetPeakDataFromRemote(const PeakData &peaks)
...
+        }
+        else
+        {
+            treble += band;
+        }
+
+        flux += delta;
+    }
+
+    bass /= static_cast<float>(bassBands);
```
robertlipe, line 499: Are you confident that'll never be zero?

src/soundanalyzer.cpp
```diff
@@ -296,10 +408,182 @@ void SoundAnalyzerBase::SetPeakDataFromRemote(const PeakData &peaks)
...
+    const size_t trebleBands = (NUM_BANDS > bassBands + midBands) ? (NUM_BANDS - bassBands - midBands) : 0;
+    if (trebleBands > 0)
+        treble /= static_cast<float>(trebleBands);
+
+    flux /= static_cast<float>(NUM_BANDS);
+    lowFlux /= static_cast<float>(bassBands);
+    const float beatBandGroups = static_cast<float>((bassBands > 0 ? 1 : 0) + (midBands > 0 ? 1 : 0) + (trebleBands > 0 ? 1 : 0));
+    const float beatVu = (bass + mid + treble) / std::max(1.0f, beatBandGroups);
+
+    const float score = bass * 0.55f + lowFlux * 1.50f + flux * 1.20f + mid * 0.15f + beatVu * 0.25f;
```
robertlipe, line 512: Extract magic constants at least to named constants throughout.

src/soundanalyzer.cpp
```diff
@@ -447,12 +746,12 @@ void SoundAnalyzerBase::InitI2S_Legacy()
 
     pinMode(I2S_BCLK_PIN, OUTPUT);
     pinMode(I2S_WS_PIN, OUTPUT);
-    pinMode(INPUT_PIN, INPUT);
+    pinMode(inputPin, INPUT);
```
robertlipe, line 749: This is the THIRD occurrence of this.

 grep  "pinMode.*INPUT_PIN" src/* inc*/* | wc -l
       3

src/ws281xoutputmanager.cpp
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    const DRAM_ATTR rmt_item32_t kBitZero = MakeRmtItem(NsToRmtTicks(kWs2812T0HighNs), NsToRmtTicks(kWs2812T0LowNs));
+    const DRAM_ATTR rmt_item32_t kBitOne  = MakeRmtItem(NsToRmtTicks(kWs2812T1HighNs), NsToRmtTicks(kWs2812T1LowNs));
+
+    // The translator is called by the legacy RMT driver as it needs more items.
+    // It consumes raw GRB/RGB/etc. bytes and expands each bit into one timing
+    // item, so the higher layers only need to provide packed color bytes.
+
+    void IRAM_ATTR WS2812ByteTranslator(const void* src, rmt_item32_t* dest, size_t srcSize, size_t wantedNum, size_t* translatedSize, size_t* itemNum)
```
robertlipe, line 93: I've said that I think this is a bad direction, but you're going on with this anyway. You're either going to end up taking on a whole to of strip support issues or just not support them.

src/ws281xoutputmanager.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+    // Runtime color order is a settings concern rather than a template concern,
+    // so we materialize the final byte order explicitly into the per-channel
+    // output buffer before handing bytes to the RMT driver.
+
+    void WriteColorOrderedBytes(uint8_t* dest, const CRGB& color, DeviceConfig::WS281xColorOrder colorOrder)
+    {
+        switch (colorOrder)
+        {
+            case DeviceConfig::WS281xColorOrder::RGB:
```
robertlipe, line 121: Unswitching this would be faster. It's an optimization  already implemented in other options.

src/ws281xoutputmanager.cpp
```diff
@@ -0,0 +1,416 @@
...
+                *errorMessage = "failed to allocate WS281x byte buffer";
+            return false;
+        }
+
+        std::fill_n(nextOutputBytes.get(), byteCount, 0);
+        state.outputBytes = std::move(nextOutputBytes);
+        state.byteCount = byteCount;
+    }
+
+    // The IDF4-compatible helper seeds a TX config for the requested channel/GPIO.
```
robertlipe, line 224: This will all need replaced on IDF5.

Current is 6.

platformio.ini
```diff
@@ -32,8 +32,8 @@ extra_configs =
 [base]
 upload_port     =
 monitor_port    =
-build_type      = release
-upload_speed    = 2000000
+build_type      = debug
+upload_speed    = 4000000
```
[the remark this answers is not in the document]
robertlipe: See previous reviews.

robertlipe (commented):
This is at least 4 different PRs. Rutger would throw me out if I submitted this.
1) Beat
2) UI
3) Dynamic
4) Replace LED with homebrew.

This overlaps greatly with reviews I've already submitted, so I can't tell what was actually implemented and fixed and just not in this tree vs. what's just being clicked 'ignore'.

Unless review comments are going to be taken seriously, I can't keep re-reviewing the same files.

 #4 is either going to work because people are only using the simplest of strips or run people off who are using funky strips. It's a support problem. There are way more ESP32 and strip combinations than are obvious.

You can debug the IDF5 path today with any of our _v3 builds.

rbergen:
The "stacked PR" aspect is something I've sent Dave a separate message about. The way #857, #858 and #859 are now related basically means we need to review and merge (with the combination of those two being an "atomic" operation) each of these PRs in that order to prevent duplication of work. 

About the FastLED thing, my question is what our options really are if we do want to implement the features Dave is now aiming to add, and we agree FastLED is not compatible with our path forward. (And I'll be brutally honest, a maintainer allowing the current generation of token predictors to have the last word about _anything_ is a clear red flag for me in itself.)
There are clear downsides in terms of at least breadth-of-device support to the homebrew option, but are there other truly viable options we can use?

robertlipe:
Only the ones I've listed several times in approximate increasing order of
viability.
A) Patch FastLED 3.
B) Adafruits Neopixel library. Meh. Not great ESP32 support.
C) Use FastLED 4, which has the new channels API meant for exactly this ,
but finding some way past my head clash with Zach in the iso C++ thing I've
mentioned several times. With that out of the way, dynamic configuration is
O(hundred lines)... mostly FastLED is fixed, perhaps patching in the fix I
already produced. The long-term future of FastLED is a concern, but it has
inertia.
D) the Espressif LED library which, unsurprisingly, has great support for
esp32 before chips are launched and less than great for all the novel led
types I'm trying to defend here.
E) Makuna Neopixel library is the direction that WLED went when FastLED was
leaderless. Hard to argue against their success. Great esp32 support. Great
strip support.

WW/CW, Pucks, clocked, manufacturing variances, etc. all exist and "work"
today,.albeit with some manual configuration. All stop working with this PR.

It's common to mix the pixel math library of FastLED (it's separate) for
the pixel math and outsource the DMA bashing. That's how WLED went. That's
how we've used Smartmatrix forever. It's just the lingua franca of
blinkies. I've experimented with three of the above (that FastLED patch
exists for a reason) and found they're viable, but didn't get as far as
something worth submitting.

There are many options beyond vibe coding this with 'works on my strip and
my ESP32') and taking on support for the matrix of dozens of led types,
dozens of manufacturers, and a dozen different esp32s. S31 is inevitable
for us. P4 and P4X are already here.

I'm qualified to comment on our problematic third-party libs in this
project and really think that for both types of actual pixel libs, strip
and matrix, we should continue to outsource those to *externally actively
maintained* projects. Smartmatrix's pretty clear successor is Mr.
Codetastics lib and it allows dynamic reconfiguration. Panel support is at
least as messy as strip support. Look at their respective bug queues in
these projects.

Re: stacking this number of unrelated changes, that just Rules for Thee. I
get it his project, so the default is for me to lose this argument, but how
do you expect other contributors that have to pass reviews to react while
we have to pass gauntlets for contributions? The one new contributor we've
had in the last, what, two years went quiet because they were (reasonably)
asked to split an overreaching PR.

davepl:
Keep in mind that the architectural change here was the separation of transport from rendering… and I only implemented WS2812B, but we can add them for any strips we ever wish to support.

It’s just an output driver now, and we have two (HUB75 and WS2812B).

In fact, you COULD defer to FastLED4 for particular types.  Let it do the Show(), I don’t care… but the base code can/should/must preserve the ability to change on the fly, ideally.

I’m confident I could do the same for HUB75 but my knowledge is limited enough that I’d probably get it working with the specific type of panel I have, and maybe not others.  As odd as SmartMatrix can be, I think they’ve got a lot of historical compatibility work in there.

WS2812B, not so much, it’s pretty easy.

- Dave

robertlipe:
>  defer to FastLED4 f

If you use FastLED 4, use the new Channels API which allows for runtime
construction and dstruction of these, shim our current ADD_CHANNEL, add a
contruent DESTROY_CHANNEL, and we're done. Most of this thrash becomes
completely unnecessary. It's literally under a hundred lines of code and I
can spot you about a third of those if I could find the tree I used. See:
https://github.com/FastLED/FastLED/blob/master/src/fl/channels/README.md It
already knows about all the DMA engines (four useful ones), not just one
and all the LED types. FastLED 4 is in late stages of development and not
without sharp edges. See https://github.com/FastLED/FastLED/issues/2167 and
https://github.com/FastLED/FastLED/pull/2196

I think Moonlight (one of the high-performance WLED forks) has committed to
this.

Staying on FastLED also keeps the door open for non-ESP32 ports, notably
Teensy.


We've already agreed that dynamic reconfiguration of panels is less in
demand than dynamic configurationf or strips. I'm content to focus on doing
_somethhing_ about FastLED3 in this PR. People don't plug and chug panels
the way they do strips.

> It’s just an output driver now, and we have two (HUB75 and WS2812B).

That's what we have after this PR. We have TODAY, with minimal futzing of
ADD_CHANNELS, support for all these:
https://github.com/FastLED/FastLED/wiki/Chipset-reference

> WS2812B, not so much, it’s pretty easy.

Well, even that one type is really not so easy.
There's the host/DMA side, with a plethora of needed DMACs with different
tradeoffs of performance and memory size.
There's the LED chipset side. There are about a dozen different chipsets in
use. Most of what people call WS2812B arent' even WS2812B and there are a
ton of clones out there.  WS2811s (pretty much everything that isn't 5V,
which means almost everything of any distance)  of a certain age use an
annoyingly different timing envelope. (e.g.
https://forums.adafruit.com/viewtopic.php?t=43060) There are still B-not's
in circulation. WS2812C's deliver better color saturation at low
intensities and is used by some vendors  WS2813 (I get it. WS2813 !=
WS2812) can drive at 4Khz instead of 800Hz. A random strip from Ali or even
BTF might contain any of these and if it's 12 or 24V is almost certainly
2813 or 2815 or maybe 2805. If it's outdoor puck, then probably it's none
of these. SK6812 (also duly noting that SK6812 != WS2812B) has a longer
reset pulse and is 32-bits delivering less chroma crushing and less chunky
steps between colors. (And, like APA102, are beautiful, and eliver most of
the value without adding a clock signal.! Wish they weren't so darned
expensive.)

See
https://github.com/FastLED/FastLED/blob/master/cookbook/reference/led-types.md
for how deep this river runs.

Please Resist the urge to haggle with any of these specific examples. (C's
and B's are very software compatible, unlike the V4 and V5 of B's. V6 has
been shipping for a while.) The underlying point is that there's a complete
world to each half of this "pretty easy" point that I don't think we should
just flip away.

In case anyone is actually reading this,
https://github.com/PabloCastellano/awesome-ws2812 is a decent resource,
though it's getting a bit stale.
https://aleaengineering.com/2020/01/06/experiments-in-implementing-the-ws2812b-protocol-in-avr-mcus-without-bit-banging/
is nice, too.

"WS2812b" is as colloquial as "x86" was. The kernel and compiler guys knew
they were similar, but not "easy".

Thank you for  '830. Without re-reviewing it all, that's helpful, though I
think most of those comments I made on a different, smaller, PR from this
morning I've lost track of which PR[s] have what review comments applied,
pending, or dismissed, though.

davepl:
How many programmers does it take to change a light bulb?  It works in my office!  

So do all the LEDs, at least so far, and I’ve got quite a mix.  Let me know if you run into anything that should work but doesn’t!

When I do the spiral lamp rebuilt, it’s RGBW, so I’ll hit that.  But I cheated last time by injecting the extra bytes in FastLED, and it’ll be a lot easier this time with full control.

Cheers
Davce

robertlipe:
I just learned that the email bridge damages links. If you find yourself
screaming "citation needed, <https://xkcd.com/285/>" please see the actual
GH issue.


> How many programmers does it take to change a light bulb? It works in my
> office!


Given your background, I can't believe I have to press the "it works on my
machine, ship it" issue. How well represented are any of the aforementioned
chipsets in your lab? How do those APA102's look?

>  I’ve got quite a mix.

Do you *really* have the diversity I'm trying to represent here or do you
just have a lot of the same thing?

Now at this point, we each must retreat into our respective corners and say
that given the way this project is structured and the support traffic we
have, we simply don't KNOW what percentage of our users are using SK6812,
APA102, FW1906, etc. Neither one of us can tangibly argue in data. You
recently posited that Espressif's retail division, M5
<https://www.espressif.com/en/news/Espressif_Acquires_M5Stack>, was
dominant. In your office, that's likely true. In my office, I just skimmed
receipts and have upward of 120 ESP32 mutants (Nightdriver isn't my only
way to stay amused...) and have exactly one M5 product...because you sent
it to me. (Thank you again. It's been helpful....custom_m5tab.ini was one I
meant to call out in that other PR) We we can each point to our favorite
"benchmarks" that underscore our respective positions, but we dont'
_know_.  We agree that MOST users are on something that's close enough to
WS2812 that their lives will be simplified by being runtime
configurable. That this isn't a good goal isn't even the the contention.
Right now, a developer wanting APA102 or SK2812 WWA mode can build with a
one-line change and get great results. After this (family of) PR's, I don't
see how they can.

When I was shopping for a core blinky patform for projects during late
COVID and saw that NightDriver used a home-rolled DMA/Chipset solution, I'd
have kept scrolling. (You're free at this moment to wish that I had done
so.😇)  so 100% of Daves would be happy and 100% of Roberts would have
dismissed it before ever building.  We'll surely agree that "market truth"
is somewhere between those points, becuase that's the range of math. I've
built projects for friends using APA102 and Nightdriver and was going to
deploy
<https://gree-leds.com/ic-protocol/fw1906-ic-protocol-and-datasheet.html>
FW1906 <https://github.com/wled/WLED/pull/3810> (TM1906?) 40V pucks with
NightDriverLED under my awning this fall until life intervened. So the
number of people deploying NightDriver on funky strips may not even need
two hands to count, but it's not zero. We just don't know. Developer/user
discussions in this project have always been frighteningly quiet.

When I do the spiral lamp rebuilt, it’s RGBW, so I’ll hit that. But I
> cheated last time by injecting the extra bytes in FastLED, and it’ll be a
> lot easier this time with full control.
>

Indeed, the times i've needed WS2805 or WS2813 support, there's been some
required (trivial) compile-time futzing to get all our CRGB[W|WW] ducks in
a row, but we should be moving toward making that easier, not "write your
own controller layer". We're alread requiring that futzing to get the pins
and counts lines up (eliminating that is the point: got it) but it's not
been THAT much harder for a hobbyist to configure until now.  Note that
WS2805 and WS2815, and one of the SK siblings have good WW/CW support, but
it's up to the app (hey, that's us!) to decide if we want to handle the
white channels via additive or subtractive blending.
https://kno.wled.ge/features/cct/

It's OOB for this PR, but in case any of your short-term projects involve a
ladder (and nightdriver - I can't help you with spiders or leaky shingles
:-) talk to me about upcoming partiion table changes in the hope we can
save you from doing it twice. The upcomign Arduino3 shift WILL require an
incompatible partition table (NVS *has* to grow. Might as well ditch the
deprecated SPIFFS and and move to wear-leveling and directory-supporting
<https://github.com/RalphBacon/203-SPIFFS-vs-LITTLEFS> (Hello, DOS 2.0!)
LittleFS while wer're at it....and add crash dump partions - let's break
all the parittion table issues when we break them) before you do. This is
another five-move chess maneuver I've been planning (nerd disclaimer: I
don't play chess) that I've studied the hell out of and have just concluded
is unavoidable. It's just a question of how many times we do it. YOu've
already chastised me about things not directly in this PR, but if you DO
decide to push what's under your awning, I'd rather you be annoyed at me
for telling you ten[2] times than for not telling you at all.)

It's increasingly clear that I can't talk you out of this move. This
represents the worst combination of device
<https://github.com/FastLED/FastLED/blob/master/src/platforms/readme.md#popular-boards--where-to-look>
support (ESP32 and otherwise), homebrew dynamic configuration, multilane
<https://github.com/FastLED/FastLED/blob/master/src/platforms/readme.md#multilane-output>,
clocked
<https://github.com/FastLED/FastLED/blob/master/src/platforms/readme.md#2-spi-clocked>,
and chipset
<https://github.com/FastLED/FastLED/blob/master/src/platforms/readme.md#t1t2t3-vs-t0ht0lt1ht1l>
support. I'll either get over it or I'll fork. I won't keep defending it,
but I won't quite table-flip out, either. I think we can still collaborate
on the rest.

[1] I have a PR somewhere that shoves all our file I/O through ONE
abstraction, making it trivial to conditionalize.
[2] For some value of "ten"

davepl:
I was making a joke, but there’s some truth to it.  Shipping is a feature, Robert.  Don’t make the perfect the enemy of the good - pick a feature set, get it stable, ship it.

I’d rather ship a fast and robust WS2812B and HUB75 mix that covers 98% of hardware than be tied to a platform that precludes the very feature I’m try to achieve.

I’m more than happy to add (back) support for any mainstream tech that people are actually using, but supporting formats for the sake of format count isn’t in my nature.  But I certainly would not want to break and undo any work done here previously.

That said, losing formats that only worked as a side-effect of using FastLED when I didn’t even know we supported them is not a big regression to me.  

If there’s a format we supported last week that I “broke” please open an issue and if appropriate I will handle it!

Thanks,
Dave

PS: I guess my unstated goal/maxim here is “We’ve built an incredibly powerful alternative to WLED but its too hard for regular humans to use.  Let’s turn what we have now into something we can get into the hands of hobbyists.  Let’s make it a tool OTHER people can make videos about.

If you want to light 30 LEDs, WLED will always be easier.  But that’s not a reason not to lower the barrier to entry with NDS, which is really what I’m trying to do.

rbergen:
I took a first look at this - focusing on the actual web UI code -, and I see some things that I think are functionally incorrect, as well as some "sloppiness". None of that is surprising for a vibe-coded UI, but it does warrant a more in-depth review. That will take some time, which may well make my full review feedback land during the weekend.

site/app.js
```diff
@@ -0,0 +1,1787 @@
...
+    els.prevEffectButton.addEventListener("click", () => postForm("/previousEffect").then(loadEffectsOnly));
+    els.nextEffectButton.addEventListener("click", () => postForm("/nextEffect").then(loadEffectsOnly));
+    els.refreshEffectsButton.addEventListener("click", () => loadEffectsOnly());
+    els.saveIntervalButton.addEventListener("click", applyEffectInterval);
+    els.reloadSettingsButton.addEventListener("click", () => loadSettingsOnly());
+    els.applySettingsButton.addEventListener("click", () => applyDeviceSettings(false));
+    els.applySettingsRebootButton.addEventListener("click", () => applyDeviceSettings(true));
+    els.rebootButton.addEventListener("click", () => resetDevice({ board: 1 }));
+    els.resetDeviceConfigButton.addEventListener("click", () => resetDevice({ board: 1, deviceConfig: 1 }));
+    els.resetEffectsConfigButton.addEventListener("click", () => resetDevice({ board: 1, effectsConfig: 1 }));
```
rbergen, line 153: I think the payload for these calls should match the events. For resetDeviceConfigButton, only reset the board, for resetEffectsConfigButton, only reset the effectsConfig. Also, use `true` instead of `1` for the flag you actually want to set.

site/app.js
```diff
@@ -0,0 +1,1787 @@
...
+    els.applyEffectDialogButton.addEventListener("click", applyEffectSettings);
+    window.addEventListener("resize", () => {
+      updateActiveTabCutout();
+      drawPreviewFrame();
+    });
+  }
+
+  function initializeStaticShell() {
+    els.hostValue.textContent = window.location.hostname || "--";
+    els.webPortValue.textContent = window.location.port || (window.location.protocol === "https:" ? "443" : "80");
```
rbergen, line 179: Use of https is currently purely hypothetical, in the sense that on-board webserver has no support for it.

site/app.js
```diff
@@ -0,0 +1,1787 @@
...
+      const gripCell = document.createElement("td");
+      gripCell.className = "grip-cell";
+      const grip = document.createElement("span");
+      grip.className = "drag-grip";
+      grip.draggable = true;
+      grip.title = "Drag to reorder";
+      grip.addEventListener("dragstart", (event) => handleEffectDragStart(event, index, tr));
+      grip.addEventListener("dragend", () => clearEffectDragState());
+      gripCell.appendChild(grip);
+      tr.appendChild(gripCell);
```
rbergen, line 450: I think we should use the dedicated (table) DOM APIs JavaScript provides to put the table's contents together, instead of injecting raw HTML/elements.

I know Claude can help with this clean-up - and explain why it is a better approach.

This comment extends to the entire renderEffects function, meaning beyond these lines, and all other render* functions.

site/app.js
```diff
@@ -0,0 +1,1787 @@
...
+
+    for (let index = 0; index < Math.max(compiledMaxChannels, configuredPins.length); index += 1) {
+      specs.push({
+        name: `${SYNTHETIC_WS281X_PIN_PREFIX}${index}`,
+        friendlyName: `WS281x pin ${index + 1}`,
+        description: `GPIO assigned to WS281x channel ${index + 1}.`,
+        type: settingType.Integer,
+        minimumValue: -1,
+        maximumValue: 48
+      });
```
rbergen: Why is this collection of setting specs not generated and returned at the server side, like all the others? (To be clear, that is to say I think they should be.)

site/app.js
```diff
@@ -0,0 +1,1787 @@
...
+
+  function compareSettingSpecs(left, right) {
+    const rankDiff = settingPriority(left.name) - settingPriority(right.name);
+    if (rankDiff !== 0) {
+      return rankDiff;
+    }
+    return String(left.friendlyName || left.name).localeCompare(String(right.friendlyName || right.name));
+  }
+
+  function settingPriority(name) {
```
rbergen: This function, and a bunch of others that follow, apply a whole heap of client-side knowledge about the effect settings that are supposed to be dynamic and defined at the server side.

I strongly believe this needs to be reimplemented so that all of the knowledge about what settings exist resides at the server side, which I acknowledge up-front will translate to an extension of the server-side setting(Spec) related logic and APIs.

rbergen (changes requested):
I started my in-depth review and came about halfway through app.js before I decided to stop. As my last comment in the set expresses, this UI embeds a LOT of front-end knowledge about settings that are supposed to be dynamically defined at the server side. 

This makes the current implementation brittle to the point I don't think it's responsible to merge this. A next iteration should move all knowledge about what settings exist to the back-end, and extend existing  C++ and JSON APIs to implement and publish this.

davepl:
Are you suggesting this as a future work item for the API, or are you impugning the client code for not using the available functionality?

I would agree it's better not to bake knowledge into the client (though I confess I do not care, and only care about the server side).  But is that something we support now that it should have taken advantage of, or a new work item you’d like to see on the API?

I will reiterate that I do not care about the internals of app.js, which could be half Egyptian Hieroglyphs and half LISP.  I do care how it impacts the server API, though.  So describe your proposed changes in the context of the C++ API.

- Dave

rbergen:
What I'm saying is that _if_ we consider the UI a disposable artifact in essence, which I believe is the premise, then we should allow API consumers to reliably replicate its functionality without having to incorporate logic that is specific to the throw-away.

My conclusion is that the current combination of front-end and back-end doesn't implement this. This front-end, and my interpretation of what you're saying about what matters, boils down to "it matters if it's already in the back-end and can't be hashed together in the front-end" which makes it a statement that is effectively meaningless from a broader perspective, in my opinion. 

I think this conversation, and the project, would benefit from a clear statement on what matters in terms of stability, consistency and predictability on the one hand, and what doesn't on the other. What matters we should treat accordingly, what doesn't we shouldn't waste our time on and allow AI to run away with.

davepl:
My stance is thus:  “What happens in app.js stays in app.hs.  It can hardcode binary offsets, I don’t care, don’t want to know.  Its mutable and generatable at any time with no recourse other than git history.”

There shouldn’t be dependent functionality on the ESP32 side unless it’s (perhaps partial support) for the API and is useful to ALL consumers that would consume it for UI.

I’m still not clear on the app.js side or where your comments are or where to find them.  I did not receive any PR comments beyond distaste for the slop, which are not actionable.  I need commentary on the CPP side to fix anything.

- Dave

rbergen:
The key point I was making was about what should be in app.js and what shouldn't - and yes, I phrased that at a level that I felt was in line with the nature of the code I was reviewing. I'm happy to go into more detail about it, if we agree it's something that should be fixed, at a conceptual level.

In my view, the topic we're discussing is an architectural one to its core, and one that directly relates to the actual scope of the back-end that's on flash. Whatever falls in that scope I think we do agree on is stuff that matters.

site/app.js
robertlipe: It's worth actively avoiding. Certificate and related overhead in the onboard OS is worth turning off in cases where we care about code size.

rbergen:
As indicated, my main issue is that the front-end embeds a lot of knowledge about the settings (setting specs) it's getting from the back-end, which makes the overall implementation brittle.
As you asked for specifics, I've asked Sonnet to extract the embedded knowledge that I feel should be moved to the back-end to adhere to the premise of dynamic setting definition.

I hope this helps!

## Embedded Settings Knowledge in app.js

### 1. Section assignment (`settingSectionKey`, lines ~640–680)

The front-end hardcodes which setting name belongs to which UI section. This is a complete, named mapping:

| Setting names | Assigned section |
|---|---|
| `matrixWidth`, `matrixHeight`, `matrixSerpentine` | `topology` |
| `outputDriver`, `ws281xChannelCount`, `ws281xColorOrder`, `ws281xPin*` | `output` |
| `location`, `locationIsZip`, `countryCode`, `timeZone` | `location` |
| `use24HourClock`, `useCelsius`, `ntpServer`, `openWeatherApiKey` | `clock` |
| `audioInputPin` | `audio` |
| `brightness`, `globalColor`, `secondColor`, `applyGlobalColors`, `clearGlobalColor`, `showVUMeter`, `rememberCurrentEffect`, `effectInterval` | `appearance` |
| Everything else | `system` |

The section title and description strings are also hardcoded in `getSettingsSections()`.

---

### 2. Display ordering (`settingPriority`, lines ~610–635)

Explicit priorities for how settings are ordered within their section:

| Setting | Priority |
|---|---|
| `matrixWidth` | 0 |
| `matrixHeight` | 1 |
| `matrixSerpentine` | 2 |
| `ws281xPin0..N` | 3 + pin index |
| `outputDriver` | 10 |
| `ws281xChannelCount` | 11 |
| `ws281xColorOrder` | 12 |
| All others | 100 |

---

### 3. Special input control overrides (`buildSettingField`, lines ~760–830)

Named settings that get a non-default input widget, with business logic embedded per name:

- **`brightness`** — Slider with a custom percentage scale (5–100%) mapped from a raw 0–255 integer. The conversion functions `rawBrightnessToDisplay` / `displayBrightnessToRaw` and the clamped range are hardcoded in the front-end.
- **`effectInterval`** — Custom composite control: a boolean toggle ("Rotate effects") + a seconds number input. The semantics (0 = pinned, positive value = milliseconds, display as seconds) and clamping range (1–2,147,483 sec) are hardcoded. The help text addition "Disable rotation to keep the current effect active indefinitely." is appended unconditionally.
- **`ws281xColorOrder`** — `<select>` with color order options.
- **`timeZone`** — `<select>` populated from an external `/timezones.json` file.
- **`countryCode`** — `<select>` populated by exhaustively enumerating all ISO 3166-1 alpha-2 codes via `Intl.DisplayNames`.
- **`outputDriver`** — `<select>` with driver options.
- **`ws281xChannelCount`** — `<select>` with channel count options, and its draft value is coerced to `Number` while all other selects stay as strings.

---

### 4. Output driver display labels (`getOutputDriverOptions`, line ~1660)

```js
value === "ws281x" ? "WS281x" : value === "hub75" ? "HUB75" : value
```
The driver identifiers `ws281x` / `hub75` and their human-readable labels are hardcoded. The spec should carry a `friendlyName`.

---

### 5. `hostname` requires-reboot knowledge (`applyDeviceSettings`, lines ~1145–1155)

The code checks specifically for `hostname` in the draft and shows a reboot warning. The fact that this setting requires a reboot is embedded in the front-end. It should be expressed via a `requiresReboot` flag on the setting spec.

---

### 6. Synthetic WS281x pin settings (`getOrderedDeviceSettingSpecs`, lines ~565–600)

The front-end **invents** specs for `ws281xPin0..N` that do not come from `/settings/specs`. It hardcodes:
- `friendlyName`: `"WS281x pin N"`
- `description`: `"GPIO assigned to WS281x channel N."`
- `type`: `Integer`
- `minimumValue`: `-1`
- `maximumValue`: `48`

The valid GPIO range (−1 to 48) and the whole concept of these as individually-addressable integer settings are front-end inventions.

---

### 7. Channel count upper-bound fallback (`getChannelCountOptions`, line ~1667)

```js
const maxChannels = Math.max(1, Math.min(8, compiledMax));
```
The hardcoded ceiling of **8 channels** is a fallback that supplements whatever `compiledMaxChannels` the schema provides. If the schema always carries this, the `8` here is redundant but still brittle.

---

### 8. Fallback color order list (line ~101)

```js
const COLOR_ORDER_OPTIONS = ["RGB", "RBG", "GRB", "GBR", "BRG", "BGR"];
```
Used when `unifiedSchema.outputs.ws281x.allowedColorOrders` is absent. The valid color orders are domain knowledge that belongs in the back-end spec.

---

### 9. API endpoint routing per setting (`splitDeviceDraftPayloads`, lines ~1680–1720)

The code knows which settings go to the legacy `/settings` endpoint and which go to `/api/v1/settings`, and how their names map between the two representations:

| Legacy setting name | Unified API path |
|---|---|
| `matrixWidth` | `topology.width` |
| `matrixHeight` | `topology.height` |
| `matrixSerpentine` | `topology.serpentine` |
| `outputDriver` | `outputs.driver` |
| `ws281xChannelCount` | `outputs.ws281x.channelCount` |
| `ws281xColorOrder` | `outputs.ws281x.colorOrder` |
| `ws281xPin*` | `outputs.ws281x.pins[N]` |

All of this mapping logic is hardcoded in the front-end. Each setting spec should carry a field indicating its API target and its canonical path within the unified schema.

---

### Summary for the PR author

The setting spec definitions in the back-end need to be extended to carry at least:

1. **`section`** — which UI group the setting belongs to (topology, output, location, clock, audio, appearance, system)
2. **`displayOrder`** — explicit ordering within the section
3. **`inputType`** or **`widgetHint`** — for settings that need non-default controls (slider with percentage scale, toggle+number composite, select from enum)
4. **`allowedValues`** — for enum-style selects (`colorOrder`, `outputDriver`, `channelCount`); the front-end should not need to know valid values
5. **`friendlyName`** / **`description`** — already partially done, but WS281x pins are currently synthesized entirely in the front-end
6. **`requiresReboot`** — flag on `hostname` (and potentially others) so the front-end doesn't need to special-case by name
7. **`apiTarget`** — which endpoint and JSON path the setting is persisted to, so `splitDeviceDraftPayloads` can be driven by spec data rather than a hardcoded mapping
8. **`scale`** / **`displayMin`** / **`displayMax`** — for `brightness`, the 0–255 ↔ 5%–100% conversion is embedded knowledge that should be carried by the spec
9. **`semantics`** — for `effectInterval`, the "0 means disabled" convention and the milliseconds-to-seconds display conversion are implicit domain rules the spec should express

rbergen:
(I forgot to mention that I feel less strongly about 9. The effect interval is a bit of a special case across back and front-end.)

davepl:
The special settings endpoint has already been removed and replaced with proper calls to the settings API, but I deferred that work to a separate checkin yesterday.  So all that was long since fixed, but not in that PR.

I’ll go through the remainder of the comments to check if anything remains, fix as appropriate, and will get you a PR. Unfortunately it’s going to be, as always, based off the latest tree (SetupUI).

- Dave

davepl:
Actually, I’ve delayed learning git long enough that AI is pretty good at it now, so was able to figure out how to make the fixes right in the PR branch for your comfort.  Will advise when they’re ready for re-review!

Thanks for the detailed list!

- Dave

rbergen (approved):
LGTM

robertlipe:
It's 05:50c. NSSCPP has switched to "sunset" mode. I reserve the right to
be wrong. Please forgive me if i'm cranky or misguided.

If there's stil a path for review feedback to be not lost, great. That
would render most of this moot.

The only way I know now to review these (education welcome if I'm
particularly stupid) is do diff the C++ parts against the proposed merge
base. I'll explicitly D/C against the C++ side abd naybe the web/JS/UI side
hides a non-obvious change from sight. If so, I'll re-apologize in advance.

I also reserve the right to be grumpy about all the things I fussed about
but were seemingly merged anyway, despite the 'dismissed' clicksand the
promise theyd be addressed in other PRs. Maybe they are and maybe they
aren't.IMO, breaking this line of feedback is like kicking the dog six
weeks after it wet the floor. We HAVE to tighten up the
comment-to-integration feedback loop here. Maybe you touched some code and
changed it because of X, but if a reviewer thinks hard about that code (and
if you think they're a bozo, please say so THERE), please either play the
"The project belongs to the project owner who will do as they do" card (I
respect that; I'm a project owner myself...and sometimes I stomp on
feedback.) car OR the "review comments don't care. IDC. Don't bother."
card. Honestly, either would be OK with me. I'd have just hard-forked
sooner instead of later.  But while the editors are open and we have
mutltiple people thinking hard aout that code _anyway_ we hould probably
just deal with it. If you absolutely, positively, can't/won't deal with it
while its open, at least open a ticket, assigned to the reviewer (who
whould be free to open the "you broke i, *you* fix it" rule, if
approparite).

THIS merged PR stil has icky PSRAM stuff, which I thought we'd all agred to
neuter.I thought
https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/859/changes#diff-d2a180980be138a3ffd943382dd9707cdc255a9e37705fe4f430ac9c10cda49cR98,
for example, we agreed to delet, not  add.

We DID get a lot of input on modern STL, in particular,  integrated. Thank
you.

spectrumeffects.h:
segment 0 is a thing with geometry.
segment 1 is another thing eith yet other geometry.
Maybe thats my pin with the Octagon before you lay down another GFX entry
for the second half of a mirrored strip. Now you code:
        const int centerLeft = (GFX[0]->width() / 2) - 1;
We're pretty sure that GFX doesn't mean what you think it means.

(This is why clicking 'resolved' withot explanations is bad.)

Re: firework effect. Still should reserve all the particles and then
emplace bck tin to them instead of doing so individually.

>   const float speedScale = std::clamp(0.5f + beat.strength * 0.32f +
beat.confidence * 0.18f, 0.25f, 1.75f);

That's still not very tasty, IMO.

Draw() can still separate the “update + remove” phase from the “draw” phase
using std::erase_if. That removes the iterator bookkeeping entirely and
tends to read much more clearly.. Something like
void Draw() override {
  fadeAllChannelsToBlackBy(_frameFade);

  const float deltaSeconds =
  static_cast<float>(g_Values.AppTime.LastFrameTime());

  if (deltaSeconds <= 0.0f)
    return;

  std::erase_if(_particles, [&](auto& particle) {
    particle.Update(deltaSeconds);

    return !particle.IsAlive() ||
    particle.position < -particle.size ||
    particle.position > static_cast<float>(_cLEDs) + particle.size;
  });

  for (const auto& particle : _particles) {
    setPixelsFOnAllChannels(
    particle.position - particle.size * 0.5f,
    particle.size,
    particle.CurrentColor(),
    true);
  }
}

or a single-pass version if locality demands it. erase_if will still help
on most ESP32s.

Re:
https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/859/changes#r3151435749
(this won't go down well), you might as well document that that "anything
not on Dave's bench won't work. Here's his phone number..." :-)  This has
at least three high performance boards I've been working on that this will
kill.  If the project decides it doesn't care, I get that, but we should
probably make a formal, hard branchpoint of where we expect things to quite
working and this seems like it should be one.

Let me try to sound it both quietly (deep inside a PR comment) and loudly:
I'll fork/replace the project before I replace the LED layer with this
amount of thought. It's ALREADY causing issues between IDF4 and IDF5 (IDF6
is current) and I now have two on my bench with ESP32-S31 (and thus,
ESP-IDF6....) landing on my bench this month.

We should consider if that WS281xColorOrder[] order is actually undefined,
too, but that's not a major fuss here.

Ignoring the CW/WW quesition doens't make it go away.  If we're definiing
an arch for the nex tfive years , this is the time to do it. Strip one is
RGBW/WW/CW. Striop2 is RGBw. Strip3 is RGBW/WW. Strip 4 is RGB/CW ( you DID
tie color order so stride, didn't you?)  That's not an unreassonable
configuration for Joe Six Pack's holiday lights. How do you handle this in
the new dyanic world? ("We didn't" int he static world is true, butit's
what we're trying to escape.

static constexpr const char * MatrixWidthTag = "matrixWidth";
static constexpr const char MatrixWidthTag[] = "matrixWidth";
Another
<https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/859/changes#r3151442339>
reasont to NOT introduce anotherlevel of pointer indirection here instead
of simplicly using a label is so that sizeof() will work as a compile-time
constant:
<https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:3,endLineNumber:6,positionColumn:3,positionLineNumber:6,selectionStartColumn:3,selectionStartLineNumber:6,startColumn:3,startLineNumber:6),source:'static++const+char*+starMatrixWidthTag+%3D+%22matrixWidth%22%3B%0Astatic++const+char+arrayMatrixWidthTag%5B%5D+%3D+%22matrixWidth%22%3B%0A%0Aint+main()+%7B%0A++static_assert(sizeof(arrayMatrixWidthTag)+%3E+4)%3B%0A++static_assert(sizeof(starMatrixWidthTag)+%3E+4)%3B%0A%0A++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:esp32g20241119,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'1',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B20',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+Xtensa+ESP32+gcc+14.2.0+(20241119)+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'Xtensa+ESP32+gcc+14.2.0+(20241119)',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+Xtensa+ESP32+gcc+14.2.0+(20241119)+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4>

> struct psram_array_deleter

Now we just added things to the logic that I thought we agreed to delete.
(Not tis line, a LOT in interfaces.h, which I'm now skipping over.) This is
about a quarter or a third into this PR., but I'm hitting the hay.

These PRs are too large and, if this is indicative, are being merge with
too mny outstanding and casually dismissed disagreements.

IMO, this was merged prematurely.

These overlapping/repetitive PRs are torture for everyone involved. Maybe
this merge is just the middle of some process I can't yet set, but ...
we'll we'll talk about options later.



If you're going to insist on contributors (including me) to submit small,,
self-cntained, coherent, reviewable PRs, the rules need to apply
internally, too.


I think NDSCPP has now goine back to SUNUP mode.

RJL

rbergen:
I'm going to keep it short here: I'm trying to get some sanity back into the overall state of affairs, and I did feel merging this one was part of that process. As the full picture extends way beyond this PR, I'm going to send a separate email about it.

PR #860, opened 2026-04-28
Fine tune stacks

davepl:
## Description
Reduces stack usage, frees up about 10K of base RAM.  I tested all scenarios I could think of, including debugger commands like clock, etc....

Add instrumentation to the main printout for now so we can monitor stack

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -700,6 +700,9 @@ void loop()
 
             const auto& taskManager = g_ptrSystem->GetTaskManager();
             strOutput += str_sprintf("CPU: %03.0f%%, %03.0f%%, FreeDraw: %4.3lf", taskManager.GetCPUUsagePercent(0), taskManager.GetCPUUsagePercent(1), g_Values.FreeDrawTime);
+            const auto stackUsage = taskManager.GetStackUsageSummary();
+            if (stackUsage.length() > 0)
+                strOutput += str_sprintf(" Stack: %s", stackUsage.c_str());
 
             debugI("%s", strOutput.c_str());
         }
--- a/src/taskmgr.cpp
+++ b/src/taskmgr.cpp
@@ -19,6 +19,25 @@
 
 #include <esp_task_wdt.h>
 
+namespace
+{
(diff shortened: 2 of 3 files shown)


src/taskmgr.cpp
```diff
@@ -234,6 +263,30 @@ void NightDriverTaskManager::NotifyNetworkThread()
     xTaskNotifyGive(_taskNetwork);
 }
 
+String NightDriverTaskManager::GetStackUsageSummary() const
+{
+    String output = TaskManager::GetStackUsageSummary();
```
robertlipe, line 268: As bad as Arduino String tends to be, find some typical value of the length of this (little penalty if it's wrong by a few percent) and .reserve() the space to reduce allocation thrash.

src/main.cpp
```diff
@@ -700,6 +700,9 @@ void loop()
 
             const auto& taskManager = g_ptrSystem->GetTaskManager();
             strOutput += str_sprintf("CPU: %03.0f%%, %03.0f%%, FreeDraw: %4.3lf", taskManager.GetCPUUsagePercent(0), taskManager.GetCPUUsagePercent(1), g_Values.FreeDrawTime);
+            const auto stackUsage = taskManager.GetStackUsageSummary();
+            if (stackUsage.length() > 0)
```
robertlipe, line 704: Is it ever NOT > 0? Looks like it always writes to output.

src/taskmgr.cpp
```diff
@@ -19,6 +19,25 @@
 
 #include <esp_task_wdt.h>
 
+namespace
+{
+    void AppendTaskStackUsage(String& output, const char* label, TaskHandle_t handle, size_t stackSizeBytes)
```
robertlipe, line 24: This might be clearer if it just returned output instead of constantly reallocating and writing into output.

Now that RVO is guaranteed the calling convention can be quite nice.

robertlipe:
Oh, and if we need to sandbag these on RISCV-V where the calling convention is just more stack heavy because we don't have register windows like Xtensa does, we'll come back and fix these then.

PR #861, opened 2026-04-28
Faster Visualizer, Pulsars Effect update

davepl:
## Description
Improves the performance of the visualizer.  Adds a delay between frames in the client.   Updates the Pulsar effect to take advantage of the new Beat Detection.

This moves the COLORDATA server to CORE0 so it's not competing with Draw. 

### DIFF
--- a/include/effects/matrix/PatternPulse.h
+++ b/include/effects/matrix/PatternPulse.h
@@ -152,29 +152,24 @@ class PatternPulsar : public BeatEffectBase, public EffectWithId<PatternPulsar> 
     static constexpr int kMaxNewStarsPerFrame = 8;
     int diff;
 
-    static constexpr float Clamp01(float value)
-    {
-        return value < 0.0f ? 0.0f : (value > 1.0f ? 1.0f : value);
-    }
-
     static constexpr size_t ComputeBurstCount(const BeatInfo& beat)
     {
-        const float normalizedConfidence = Clamp01(beat.confidence / 1.5f);
-        const float normalizedStrength = Clamp01(beat.strength / 2.5f);
-        const float normalizedBass = Clamp01(beat.bass);
+        const float normalizedConfidence = std::clamp(beat.confidence / 1.5f, 0.0f, 1.0f);
+        const float normalizedStrength = std::clamp(beat.strength / 2.5f, 0.0f, 1.0f);
+        const float normalizedBass = std::clamp(beat.bass, 0.0f, 1.0f);
         const float majorBonus = beat.major ? 0.25f : 0.0f;
-        const float burstScore = Clamp01(
+        const float burstScore = std::clamp(
             normalizedConfidence * 0.40f +
             normalizedStrength * 0.35f +
             normalizedBass * 0.15f +
-            majorBonus);
+            majorBonus, 0.0f, 1.0f);
 
         return 1U + static_cast<size_t>(burstScore >= 0.60f);
     }
 
     static constexpr int ComputeMaxSteps(const BeatInfo& beat)
     {
-        const float normalizedStrength = Clamp01(beat.strength / 2.5f);
+        const float normalizedStrength = std::clamp(beat.strength / 2.5f, 0.0f, 1.0f);
         const int minSteps = beat.major ? 7 : 5;
         const int maxSteps = beat.major ? 13 : 10;
         return minSteps + static_cast<int>(normalizedStrength * static_cast<float>(maxSteps - minSteps));
--- a/include/effects/matrix/PatternPulse.h
+++ b/include/effects/matrix/PatternPulse.h
@@ -134,15 +134,16 @@ class PatternPulsar : public BeatEffectBase, public EffectWithId<PatternPulsar> 
 
     std::vector<PulsePop> _pops;
 
-    float fadeRate = 0.9;
+    static constexpr float kFadeRate = 0.9f;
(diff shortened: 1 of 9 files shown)


src/ledviewer.cpp
```diff
@@ -106,44 +106,53 @@ int LEDViewer::CheckForConnection()
...
+    const auto bytesSent = send(socket, pData, cbSize, MSG_DONTWAIT);
+
+    if (bytesSent < 0)
     {
-        debugE("Could not write to color data socket\n");
-        return false;
+        if (errno == EAGAIN || errno == EWOULDBLOCK)
+            return SendResult::WouldBlock;
+
+        debugE("Could not write to color data socket: %s", strerror(errno));
```
robertlipe, line 147: Should this be in the WouldBlock block or is it too common to log?

src/network.cpp
```diff
@@ -859,35 +872,53 @@ void IRAM_ATTR ColorDataTaskEntry(void *)
...
-            webSocketServer.SendColorData(leds, activeLEDCount);
-#endif
-        }
+                previewPacket->header = COLOR_DATA_PACKET_HEADER;
+                previewPacket->width  = graphics.GetMatrixWidth();
+                previewPacket->height = graphics.GetMatrixHeight();
+                memcpy(previewPacket->colors, leds, sizeof(CRGB) * activeLEDCount);
 
+                // Prefer the websocket preview transport used by the local UI. The legacy raw TCP
+                // preview socket remains available for external tools, but we do not drive both
```
robertlipe: Good. nightdriver_client.py --live-view depends on that.

src/network.cpp
```diff
@@ -859,35 +872,53 @@ void IRAM_ATTR ColorDataTaskEntry(void *)
...
-        wsListenersPresent = webSocketServer.HaveColorDataClients();
+                if (wsListenersPresent)
+                {
+                    webSocketServer.SendColorData(leds, activeLEDCount);
+                }
+                else
 #endif
+                if (socket >= 0)
+                {
+                    debugV("Sending color data packet");
```
robertlipe: I know what 'verbose' means, but isn't that a LOT of traffic?

include/effects/matrix/PatternPulse.h
```diff
@@ -134,9 +134,51 @@ class PatternPulsar : public BeatEffectBase, public EffectWithId<PatternPulsar>
...
+    {
+        return value < 0.0f ? 0.0f : (value > 1.0f ? 1.0f : value);
+    }
+
+    static constexpr size_t ComputeBurstCount(const BeatInfo& beat)
+    {
+        const float normalizedConfidence = Clamp01(beat.confidence / 1.5f);
+        const float normalizedStrength = Clamp01(beat.strength / 2.5f);
+        const float normalizedBass = Clamp01(beat.bass);
+        const float majorBonus = beat.major ? 0.25f : 0.0f;
```
robertlipe, line 160: Throughout this function, you sprinkle a ton of magic numbers. Prefer named constants that give a human reader some clue that "13", ".025", and "0.035" actually DO.

kFadeRate is an example of this done well.

include/effects/matrix/PatternPulse.h
```diff
@@ -134,9 +134,51 @@ class PatternPulsar : public BeatEffectBase, public EffectWithId<PatternPulsar>
 
     std::vector<PulsePop> _pops;
 
-    float fadeRate = 0.9;
+    static constexpr float kFadeRate = 0.9f;
+    static constexpr int kMaxNewStarsPerFrame = 8;
     int diff;
 
+    static constexpr float Clamp01(float value)
```
robertlipe: Why a dedicated function to reimplement
std::clamp(value, 0.0f, 1.0f); ?

include/effects/matrix/PatternPulse.h
```diff
@@ -134,9 +134,51 @@ class PatternPulsar : public BeatEffectBase, public EffectWithId<PatternPulsar>
...
+    {
+        const size_t burstCount = ComputeBurstCount(beat);
+        const int maxSteps = ComputeMaxSteps(beat);
+
+        for (size_t i = 0; i < burstCount; ++i)
+        {
+            PulsePop pop;
+            pop.maxSteps = maxSteps + random_range(0, 3);
+            _pops.push_back(pop);
+        }
```
robertlipe, line 188: Prefer
for (size_t i = 0; i < burstCount; ++i)
{
    _pops.emplace_back();
    _pops.back().maxSteps = maxSteps + random_range(0, 0.3f);
}

to avoid a stack-constructed value followed by a move...after ensuring that _pops is reserved() outside the loop: _pops.reserve(_pops.size() + burstCount);


Kick around the idea of giving it a ctor so we can just spring it into existence correctly and construct it right into the parent container. I can't run with it right now, but something like:

struct PulsePop {
    int maxSteps;

    PulsePop(int steps) : maxSteps(steps) {}
};
for (size_t i = 0; i < burstCount; ++i)
{
    _pops.emplace_back(maxSteps + random_range(0, 3));
}

Have I written this before? With so much overlap, I'm not sure, which is why I'm fussing.

robertlipe (changes requested):
The only thing absolutely between me and an LGTM on the C++ parts of this is the named constants thing and the unnecessary re-implementation of clamp. The others make for better code, IMO, but I wouldn't withhold approval over them.

rbergen:
Closing in favor of #869

PR #862, opened 2026-04-29
Optimize WB2812B Draw Path

davepl:
## Description
This make the locking more fine-grained on the render path, and it optimizes the brightness calc by doing it on the fly rather than as a seperate pass ever Show() call.

### DIFF
--- a/src/drawing.cpp
+++ b/src/drawing.cpp
@@ -285,12 +285,12 @@ void IRAM_ATTR DrawLoopTaskEntry(void *)
         uint16_t localPixelsDrawn   = 0;
         uint16_t wifiPixelsDrawn    = 0;
         double frameStartTime       = g_Values.AppTime.FrameStartTime();
-        auto& graphics = *g_ptrSystem->GetDevices()[0];
 
         {
             // Hold the render mutex for the frame pipeline so runtime topology/output
             // changes cannot reconfigure the active buffers mid-frame.
             std::lock_guard<std::recursive_mutex> renderGuard(g_render_mutex);
+            auto& graphics = *g_ptrSystem->GetDevices()[0];
 
             graphics.PrepareFrame();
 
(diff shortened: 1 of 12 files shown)


src/drawing.cpp
```diff
@@ -226,8 +226,8 @@ void ShowOnboardRGBLED()
             ledcWrite(2, 255 - c.g);
             ledcWrite(3, 255 - c.b);
         #else
-            int iLed = g_ptrSystem->GetEffectManager().g().GetLEDCount() / 2;
-            const auto& graphics = g_ptrSystem->GetEffectManager().g();
+            const auto& graphics = *g_ptrSystem->GetDevices()[0];
```
robertlipe, line 229: It's it possible that we have a sparse Devices array and 0 isn't defined?

I'm not sure that preview should jump through hoops to find a a valid Devices to preview, but we can avoid a crash if it's undefined/empty.

src/ledstripeffect.cpp
```diff
@@ -233,8 +233,14 @@ void LEDStripEffect::fillSolidOnAllChannels(CRGB color, int iStart, int numToFil
 
     for (auto& device : _GFX)
     {
-        for (int i = iStart; i < iStart + numToFill; i+= everyN)
-            device->setPixel(i, color);
+        if (everyN == 1)
```
robertlipe, line 236: Unswitch this for performance:
if (everyN == 1)
{
    for (auto& device : _GFX)
        fill_solid(device->leds + iStart, numToFill, color);
}
else
{
    for (auto& device : _GFX)
    {
        auto* p = device->leds + iStart;
        auto* end = p + numToFill;

        for (; p < end; p += everyN)
            *p = color;
    }
}

src/ledstripeffect.cpp
```diff
@@ -233,8 +233,14 @@ void LEDStripEffect::fillSolidOnAllChannels(CRGB color, int iStart, int numToFil
...
+        {
+            fill_solid(device->leds + iStart, numToFill, color);
+            continue;
+        }
+
+        for (int i = iStart; i < iStart + numToFill; i += everyN)
```
robertlipe, line 242: DOn't recompute that on every iteration:
int end = iStart + numToFill;
for (int i = iStart; i < end; i += everyN)
    device->leds[i] = color;

PR #865, opened 2026-05-06
Dynamic allocation of services, new ITask, IService models

davepl:
NightDriverStrip service lifecycle and audio/input cleanup
This check-in moves several long-running background loops behind explicit service lifecycles, tightens shutdown/restart behavior, and finishes the recent audio-input reconfiguration work. It also updates the RGB Music Blend Stars effect to use beat events directly, improves the web settings UI so edited interval values are not overwritten by refreshes, and adds validation/cleanup around the settings paths that can now apply changes live.

The biggest change here is that I removed the TaskManager-owned task startup path because it had become a loose collection of global task entry points with inconsistent ownership and shutdown behavior. Moving those loops into IService / ITaskService puts each task’s lifecycle next to the state it owns: construction, start, shutdown signaling, socket or hardware teardown, and destructor cleanup all live on the service object. That makes stop/restart paths much safer, avoids friend/global entry-point plumbing, gives every task a common Start() / Stop() contract, and lets services break out of blocking calls cleanly before the object or its resources are destroyed.

Some of the service migrations were straightforward, but others (notably the network service) required a bit of refactoring to break circular dependencies and move state around. The web server was a special case because of the AsyncTCP listener that is registered inside LWIP and cannot be fully reset from this fork; I added an _everStarted flag to warn if we try to start the web server more than once, which would leak or double-bind the listening socket.

Changes:

Added IService / ITaskService and migrated task-owned loops onto service objects with consistent Start(), Stop(), IsRunning(), shutdown signaling, and destructor discipline.
Moved audio sampling into AudioService, with runtime reconfigure support for the persisted audio input pin and safe silent fallback behavior when audio is stopped or unavailable.
Updated web settings handling so audio input pin changes use the shared live-apply helper and report failure instead of silently accepting a reverted setting.
Hardened socket/telnet/network service shutdown by closing descriptors from the stop path, avoiding double-close races, and cleaning up mDNS on network task exit.
Made RemoteControl::end() stop the IR polling task, so OTA/start-stop hooks actually shut down the service.
Converted color preview streaming into ColorStreamerService, including frame listener cleanup and socket cleanup on shutdown.
Converted render, JSON writer, debug console, websocket, socket server, network reader, screen, and audio serial work to the new service model where applicable.
Updated RGB Music Blend Stars to use OnBeat / OnNearBeat instead of VU ratio, with blue stars for near beats, green for beats, and red for strong beats.
Fixed the effect interval editor so background refreshes no longer immediately overwrite a value while it is being edited or saved.
Added web UI validation/error handling improvements for settings and topology changes.
Removed stale task-entry declarations and cleaned up includes after the service migration.
Added documentation around service destructor discipline and cleaned up small housekeeping items such as header comments, trailing whitespace, and the minimal drawing.h placeholder.

### DIFF
--- a/include/audioservice.h
+++ b/include/audioservice.h
@@ -1,0 +1,183 @@
+#pragma once
+
+//+--------------------------------------------------------------------------
+//
+// File:        audioservice.h
+//
+// NightDriverStrip - (c) 2026 Plummer's Software LLC.  All Rights Reserved.
+//
+// This file is part of the NightDriver software project.
+//
+//    NightDriver is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    NightDriver is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU General Public License for more details.
+//
+//    You should have received a copy of the GNU General Public License
+//    along with Nightdriver.  It is normally found in copying.txt
+//    If not, see <https://www.gnu.org/licenses/>.
+//
+// Description:
+//
+//    AudioService owns the audio sampler task and the I2S/ADC hardware
+//    lifecycle. It can be started, stopped, restarted, and reconfigured at
+//    runtime so changes to the audio input pin no longer require a reboot.
+//
+//    The compile-time audio mode (analog ADC, I2S external mic, M5 onboard)
+//    is still selected at build time because the SoundAnalyzer template
+//    binding and ESP-IDF driver paths are bound to that mode. The
+//    runtime-mutable surface is the input pin and the enabled flag.
+//    AudioConfig::FromCompileDefaults() reproduces the existing macro-driven
+//    behavior so first boot looks exactly like before.
+//
+//    The global g_Analyzer instance remains the in-process source of audio
+//    telemetry; AudioService just controls when its hardware is installed
+//    and when its sampler task is running. When audio is stopped or
+//    unavailable, AudioService::Source() returns a silent stub so effects
+//    see VU/peaks/beats == zero rather than stale or undefined values.
+//
+// History:     May-04-2026         Davepl      Created
+//
+//---------------------------------------------------------------------------
+
+#include "globals.h"
+
+#include "itaskservice.h"   // includes iservice.h
+#include "soundanalyzer.h"
+
+// IAudioSource is a stable read-only interface effects can hold instead of
+// reaching for the g_Analyzer global. It's a type alias of the existing
+// ISoundAnalyzer interface so the migration is mechanical when we get to it.
+
+using IAudioSource = ISoundAnalyzer;
+
+// AudioConfig
+//
+// Runtime-mutable audio configuration. The mode/sampleRate/I2S pin fields
+// are diagnostic and will not change live in this build target since the
+// driver path is bound at compile time, but they're carried in the config
+// so the SetupUI and Reconfigure() reflect a consistent picture.
+
+struct AudioConfig
+{
+    enum class Mode
+    {
+        Disabled,           // ENABLE_AUDIO is 0 in this build
+        AnalogADC,          // I2S-ADC built-in path (ESP32 only)
+        I2SExternal,        // External digital mic (INMP441 etc.)
+        M5Onboard           // M5Unified-managed mic
+    };
+
+    bool   enabled       = false;
+    Mode   mode          = Mode::Disabled;
+    int    inputPin      = -1;     // ADC pin, or I2S DIN pin in I2S mode. -1 disables hardware.
+    int    bclkPin       = -1;     // I2S only
+    int    wsPin         = -1;     // I2S only
+    int    dataPin       = -1;     // I2S only (typically equal to inputPin)
+    int    sampleRate    = 24000;  // SoundAnalyzerBase::SAMPLING_FREQUENCY
+
+    // Build a config that mirrors the existing compile-time defaults.
+
+    static AudioConfig FromCompileDefaults();
+
+    // Overlay the persisted audio input pin (if a DeviceConfig is reachable)
+    // onto a compile-defaults base. Used by Start() so the very first audio
+    // session honors the user's last-saved pin without a manual Reconfigure.
+    
+    static AudioConfig FromCurrentSettings();
+
+    // Human-readable mode name. Mirrors DeviceConfig::GetAudioInputModeName().
+    
+    const char* ModeName() const;
+};
+
+// AudioService
+//
+// Lifecycle manager for the audio engine. Construction is cheap; nothing is
+// allocated or installed until Start() is called.
+//
+// Two complete class definitions live below, gated by ENABLE_AUDIO at file
+// scope. The shape mirrors the SoundAnalyzer pattern in soundanalyzer.h: the
+// public surface is identical between the two variants so callers don't need
+// to #if their use sites, but each variant's method bodies are unconditional
+// rather than dotted with inline preprocessor branches.
+
+#if ENABLE_AUDIO
+
+class AudioService : public ITaskService
+{
+  public:
+    AudioService();
+    ~AudioService() override;
+
+    AudioService(const AudioService&) = delete;
+    AudioService& operator=(const AudioService&) = delete;
+
+    // IService::Name
+    const char* Name() const override { return "AudioService"; }
+
+    // Service-specific. Stops the audio engine if running, applies the new
+    // config, and restarts if newConfig.enabled is true.
+    bool Reconfigure(const AudioConfig& newConfig);
+
+    // Audio-enabled builds always report available, regardless of running state.
+    bool IsAvailable() const { return true; }
+
+    const AudioConfig& Config() const { return _config; }
+
+    // Read-only audio source. Always non-null. When audio is stopped or
+    // unavailable, the returned source reports zero VU / silent peaks /
+    // no beats so effects can run safely.
+    IAudioSource& Source() const;
+
+  protected:
+    // ITaskService hooks
+    TaskConfig GetTaskConfig() const override;
+    bool OnBeforeStart() override;
+    void Run() override;
+    void OnAfterStop() override;
+
+  private:
+    AudioConfig _config{};
+};
+
+#else // !ENABLE_AUDIO
+
+// Stub variant: same public surface, but every operation is a safe no-op and
+// Source() returns the silent fallback. Callers that ask "is audio available?"
+// get false; callers that try to Start get a clean false back rather than a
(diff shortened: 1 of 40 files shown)


site/app.js
```diff
@@ -784,7 +784,7 @@ $$$$$$$b   *u    ^$L            $$  $$$$$$$$$$$$u@       $$  d$$$$$$
     const setFieldError = (hasError, message) => {
       const help = wrapper.querySelector(".field-help");
       if (hasError) {
-        errorSet.add(key);
+        errorSet.set(key, message);
         help.textContent = message;
         help.classList.add("field-error");
       } else {
```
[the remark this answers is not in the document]
davepl: @copilot apply changes based on [this feedback](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/865#discussion_r3197101935)

src/audioservice.cpp
```diff
@@ -0,0 +1,392 @@
...
+        int   AudioFPS() const override { return 0; }
+        int   SerialFPS() const override { return 0; }
+        bool  IsRemoteAudioActive() const override { return false; }
+
+        // --- VU Metrics ---
+        float VU() const override { return 0.0f; }
+        float VURatio() const override { return 0.0f; }
+        float VURatioFade() const override { return 0.0f; }
+        float PeakVU() const override { return 0.0f; }
+        float MinVU() const override { return 0.0f; }
```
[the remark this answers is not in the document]
davepl: @copilot apply changes based on [this feedback](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/865#discussion_r3197101888)

site/app.js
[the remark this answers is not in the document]
davepl: @copilot apply changes based on [this feedback](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/865#discussion_r3197101935)

include/audioservice.h
```diff
@@ -0,0 +1,183 @@
...
+// are diagnostic and will not change live in this build target since the
+// driver path is bound at compile time, but they're carried in the config
+// so the SetupUI and Reconfigure() reflect a consistent picture.
+
+struct AudioConfig
+{
+    enum class Mode
+    {
+        Disabled,           // ENABLE_AUDIO is 0 in this build
+        AnalogADC,          // I2S-ADC built-in path (ESP32 only)
```
robertlipe, line 71: "Original" here to reduce ambiguity

include/audioservice.h
```diff
@@ -0,0 +1,183 @@
...
+        I2SExternal,        // External digital mic (INMP441 etc.)
+        M5Onboard           // M5Unified-managed mic
+    };
+
+    bool   enabled       = false;
+    Mode   mode          = Mode::Disabled;
+    int    inputPin      = -1;     // ADC pin, or I2S DIN pin in I2S mode. -1 disables hardware.
```
robertlipe, line 78: analogInputPin, please

include/audioservice.h
```diff
@@ -0,0 +1,183 @@
...
+
+    // Overlay the persisted audio input pin (if a DeviceConfig is reachable)
+    // onto a compile-defaults base. Used by Start() so the very first audio
+    // session honors the user's last-saved pin without a manual Reconfigure.
+    
+    static AudioConfig FromCurrentSettings();
+
+    // Human-readable mode name. Mirrors DeviceConfig::GetAudioInputModeName().
+    
+    const char* ModeName() const;
```
robertlipe, line 96: [[nodiscard]] Since it's always string literals, can it be a const ref?

include/debugconsole.h
```diff
@@ -0,0 +1,72 @@
...
+    void OnBeforeWaitForStop() override;
+
+  private:
+    
+  // Both fds are set by Run() and may be cleared by OnBeforeWaitForStop()
+    // running on a different task. atomic<int> gives us a lock-free
+    // exchange so we never call close() twice on the same fd.  I can't 
+    // believe POSIX doesn't define an invalid handle constant, but here we are.
+
+    std::atomic<int> _listenFd{-1};
```
robertlipe, line 68: file I/O predates POSIX by almost 20 years. The C++ way to handle this is to make it a std::optional.

include/deviceconfig.h
```diff
@@ -366,7 +366,17 @@ class DeviceConfig : public IJSONSerializable
...
-    bool SupportsLiveAudioInputReconfigure() const { return false; }
+
+    // True when AudioService can install/uninstall the audio driver at runtime.
+    // Today this matches the SupportsConfigurableAudioInputPin set: M5 onboard
+    // audio is managed by M5Unified itself, and the analog ADC path uses a
+    // fixed channel, so neither benefits from a live restart.
+
+    bool SupportsLiveAudioInputReconfigure() const
+    {
+        return SupportsConfigurableAudioInputPin();
```
robertlipe, line 377: This seems like an implicit conversion silently downcasting an optional int? (Maybe time to byte the bullet and make PinNumber a real class that can be checked for uniqueness, etc. throughout) to a bool.

At least make it ...Pin() > 0 or something to force it back to the bool space.

include/deviceconfig.h
```diff
@@ -366,7 +366,17 @@ class DeviceConfig : public IJSONSerializable
...
+    }
+
     String GetAudioInputModeName() const
```
robertlipe, line 380: If there's no reason for these to be stupid Arduino Strings, just make them const char[]. Keep constructors out of this.

include/webserver.h
```diff
@@ -170,6 +182,14 @@ class CWebServer
...
     static bool ValidateUnifiedDeviceSettings(JsonObjectConst device, String* errorMessage = nullptr);
     static bool SetSettingsIfPresent(AsyncWebServerRequest * pRequest, String* errorMessage = nullptr);
+
+    // Apply a new audio input pin to DeviceConfig and, when the build supports
+    // a live reconfigure, push it through AudioService::Reconfigure() without
+    // requiring a reboot. Reverts the persisted pin on failure. Safe to call
+    // whether or not g_ptrSystem / AudioService are present. Returns true if
+    // either the pin was unchanged or the live reconfigure succeeded.
+    
+    static bool ApplyAudioInputPinChange(int oldPin);
```
robertlipe, line 192: is this newPin?

include/webserver.h
```diff
@@ -222,13 +242,30 @@ class CWebServer
...
 
-    // begin - register page load handlers and start serving pages
+    ~CWebServer() override { Stop(); }
+
+    // IService lifecycle. Start delegates to begin() (which registers the
+    // route handlers and calls AsyncWebServer::begin); Stop ends the server.
+    // Idempotent on both sides.
+    bool Start() override;
+    void Stop() override;
+    bool IsRunning() const override   { return _running.load(); }
```
robertlipe, line 252: There's verb disagreement here. I'd expect isRunning to be a lightweight call to see if we're running. The naming convention, at least, implies that load() is a verb doing Real Work.

Maybe this method began life one way and evolved, but this seems weird.

src/jsonserializer.cpp
```diff
@@ -294,23 +329,22 @@ void IRAM_ATTR JSONWriterTaskEntry(void *)
             // Wait until we're woken up by a writer being flagged, or until we've reached the hold point
             ulTaskNotifyTake(pdTRUE, notifyWait);
 
-            if (!g_ptrSystem->HasJSONWriter())
-                continue;
-
-            auto& jsonWriter = g_ptrSystem->GetJSONWriter();
+            // Stop requested: drop straight into a final flush pass below.
+            if (ShouldShutdown())
```
robertlipe, line 333: In others, you changed the for(;;) inwo while (!ShouldShutdown). I actually like that form a little better as it makes the loop exit condition clearer.

src/network.cpp
```diff
@@ -535,12 +566,20 @@ namespace nd_network
...
                     entry.lastReadMs.store(millis());
                 }
             }
             notifyWait = pdMS_TO_TICKS(nextEventMs);
         }
+
+        // Tear down mDNS so a subsequent Stop()/Start() cycle doesn't double-
+        // register the "esp32" hostname (the underlying mdns component will
+        // refuse to re-init while it still holds an active responder).
+        // I'm not convinced MDNS even works, but this is the polite thing to do if it does.
```
robertlipe, line 580: I spent a few hours digging into it, as discussed.

I'm confident it DOES NOT work. I know what it takes to fix it, but since nobody has noticed in the last N years that it doesn't work, it's not that important. It takes a non-trivial amount of flash that we could otherwise use (like for this code...)

I'll jettison mDNS into the sun soon, so don't get too ambitious about dancing around it.

src/network.cpp
```diff
@@ -836,40 +865,26 @@ bool ProcessIncomingData(std::unique_ptr<uint8_t[]> &payloadData, size_t payload
     #endif
 }
 
-#if INCOMING_WIFI_ENABLED
-// SocketServerTaskEntry
+// SocketServerTaskEntry was migrated into SocketServer::Run() (ITaskService).
+
+#if COLORDATA_SERVER_ENABLED
```
robertlipe, line 870: If we're going to call it ColoStreamer in some places, let's eradicate 'COLORDATA" everywhere.

src/network.cpp
```diff
@@ -883,16 +898,31 @@ void IRAM_ATTR ColorDataTaskEntry(void *)
...
+    {
+        EffectManager&        mgr;
+        IFrameEventListener&  listener;
+        ~ListenerGuard() { mgr.RemoveFrameEventListener(listener); }
+    } listenerGuard{ effectManager, frameEventListener };
+
+    while (!ShouldShutdown())
     {
-        while (!WiFi.isConnected())
+        while (!WiFi.isConnected() && !ShouldShutdown())
```
robertlipe, line 920: Are all three of these checks really necessary? You do this in a few places...

src/soundanalyzer.cpp
```diff
@@ -103,28 +103,72 @@ SoundAnalyzerBase::SoundAnalyzerBase()
...
     {
-        i2s_channel_disable(_rx_handle);
-        i2s_del_channel(_rx_handle);
+        const esp_err_t disableErr = i2s_channel_disable(_rx_handle);
+        if (disableErr != ESP_OK)
+            debugW("Audio: i2s_channel_disable returned %d", (int)disableErr);
+        const esp_err_t delErr = i2s_del_channel(_rx_handle);
+        if (delErr != ESP_OK)
+            debugW("Audio: i2s_del_channel returned %d", (int)delErr);
+        _rx_handle = nullptr;
```
robertlipe, line 131: I recently got into trouble with this kind of code. (And this code is mine...)

If EITHER the _disable or the del_channel fails, rx_handle will still be in play and if we just toss it and null rx_handle, we don't have a recovery path.

Of course, if either of them failed, you're probably headed for a reboot _anyway_ but it might be worth stacking their success so that TeardownAudioInput() can fail an keep rx_handle in play. 

repeats for adc_handle and i2s_stop

robertlipe (commented):
Down to mostly style nits. Things like store(true) seem like they should have better naming, but I'm OK with all the logic here.

PR #866, opened 2026-05-06
PSRAM strategy reversal. Switched to PSRAM-default routing via heap_c…

davepl:
Make PSRAM the default for allocations >32, remove all of the "special casing" of PSRAM, properly allocate DMA memory where needed


include/gfxbase.h
```diff
@@ -188,10 +189,12 @@ class GFXBase : public Adafruit_GFX
 
     CRGB *leds = nullptr;
     #if MATRIX_HEIGHT > 1
-        std::unique_ptr<Boid[], psram_array_deleter<Boid>> _boids;
+        std::unique_ptr<Boid[]> _boids;
     #endif
 
-    using PolarMapArray = PolarMap[kMatrixWidth][kMatrixHeight];
+    // std::array nesting (rather than a raw C 2D array) so std::make_unique<PolarMapArray>()
```
robertlipe, line 195: Maybe not in this PR, but I'm in favor of std::array essentially everywhere. If we need to fix PolarMap, let's do it.

std::array is better than a C 2D array in every way I can think of.

include/interfaces.h
```diff
@@ -30,155 +30,122 @@
...
 //
-// A C++ allocator that allocates from PSRAM instead of the regular heap. Initially
-// I had just overloaded new for the classes I wanted in PSRAM, but that doesn't work
-// with make_shared<> so I had to provide this allocator instead.
+// For the rare cases that genuinely need internal SRAM (DMA targets, ISR-
+// accessed state, hot per-frame buffers that don't tolerate PSRAM latency),
+// reach for make_unique_internal / make_unique_dma below — those bypass the
+// default routing and explicitly request internal-only memory.
 
+// internal_allocator<T>
```
robertlipe, line 51: You could probably use operator placement new and eliminate these, but this is such an improvement that I'll take it!

robertlipe (approved):
There are some unrelated changes here, but this is a change I wholeheartedly welcome.

Committing this probably unblocks my work on the SIOF PR, helping the modern C++ work.

Simplicity wins!

include/interfaces.h
```diff
@@ -30,155 +30,122 @@
...
-inline bool operator==(const psram_allocator<T>&, const psram_allocator<U>&) { return true; }
-
+inline bool operator==(const internal_allocator<T>&, const internal_allocator<U>&) { return true; }
 template <typename T, typename U>
-inline bool operator!=(const psram_allocator<T>&, const psram_allocator<U>&) { return false; }
+inline bool operator!=(const internal_allocator<T>&, const internal_allocator<U>&) { return false; }
 
-template<typename T>
-struct psram_array_deleter
+// dma_allocator<T>
```
robertlipe, line 84: I patched this into a tree so I could see bigger picture.

I can't find that make_{shared,unique]_{dma,internal} ar actually used anywhere.

Plus 'shared_dma' is kind of weird concept. Those really should be monogomous.

Your solution in WS281xGFX - which is probably where you intended to use these - is actually much more tasteful than this allocator stuff anyway.

I think you raised the blade to make the change but didn't quite cut as deeply as you could have. Please whack this entire block for all four allocators.

You could also have just gone:
class WS281xBuffer {
 public:
    void* operator new(size_t size) {
      return heap_caps_malloc(size, MALLOC_CAP_DMA| MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
    }
    void operator delete(void* ptr) {
      heap_caps_free(ptr);
    }
}

auto buf = std::make_unique<WS281xBuffer>(blah);

...but I actually don't mind the C-ness of just callign heap_caps_alloc() yourself here - it's just honest. It's simple.

But please DO just whack all (most?) of this stuff in interfaces.h and let's never debug this code again.

include/gfxbase.h
davepl: Done

include/interfaces.h
davepl: I'll nuke it seperately, but agree

rbergen:
Closing in favor of #869

PR #870, opened 2026-05-09
Refactor SettingSpec construction; add ApiPath-aware dispatch; update…

rbergen:
## Note

This PR aims to update the head branch of #859 (localui), _not_ main. The reason for this is that this allows my proposed changes to the local UI to be reviewed separately before the mentioned PR is merged into the main tree. This means that after this PR has been merged, #859 still needs to be merged into main for the local UI to end up in the main tree

## Description

This PR refactors `SettingSpec` construction, corrects the webserver's setting dispatch logic, moves pin spec generation to `DeviceConfig`, aligns the web UI's select-option handling with the firmware's serialization format, and documents the unified settings API endpoints.

### `SettingSpec` constructor families (`include/interfaces.h`, `src/types.cpp`)

Adds typed constructor families (A–F) that call `FinishAndValidateInitialization()` directly, so most call sites no longer need a `FinishGuard` scope. `FinishGuard` is retained only where post-construction mutation is genuinely necessary: runtime-determined `RequiresReboot`, `EmptyAllowed` overrides, `OptionValues`/`OptionLabels` overrides, and the synthesized per-channel pin specs.

### Call-site conversions (`src/deviceconfig.cpp`, `src/ledstripeffect.cpp`, `include/effects/matrix/PatternSubscribers.h`)

All `SettingSpec` construction sites are converted to the new constructors where possible, replacing the previous pattern of constructing a bare spec and then assigning UI metadata fields one by one in a `FinishGuard` block.

### Pin spec generation moved to `DeviceConfig` (`src/deviceconfig.cpp`, `include/deviceconfig.h`, `src/webserver.cpp`, `include/webserver.h`)

The per-channel `ws281xPin{N}` setting specs were previously synthesized in `CWebServer`. They are now generated in `DeviceConfig::GetSettingSpecs()`, where the channel count, naming convention, `apiPath` format, and priority range are already available. A `pinSpecStrings` member is added to `DeviceConfig` to provide stable `const char*` backing. `CWebServer::LoadDeviceSettingSpecs()` is simplified accordingly and the `synthesizedPinSpecs`/`synthesizedPinSpecStrings` static members are removed from `CWebServer`.

### ApiPath-aware dispatch (`src/webserver.cpp`)

Both name-based dispatch loops — `ApplyEffectSettings` and `ValidateAndSetSetting` — now skip specs that carry an `ApiPath`. Settings with an `ApiPath` are addressed through the structured unified-settings path and cannot be matched by POST parameter name; previously the code would silently do nothing for them.

### Web UI select-option handling (`site/app.js`)

`getWidgetSelectOptions()` is refactored to replace the `labelOverrides` map (a client-side concept) with a unified `applyLabelMap()` helper that reads from the parallel `values`/`labels` arrays already serialized by the firmware. This aligns the UI with the wire format used by `SendSettingSpecsResponse` for all three option sources (`schemaPath`, `externalTimeZones`, and inline), removing a layer of client-side knowledge about how individual values should be labelled.

### REST API documentation (`REST_API.md`)

Documents the previously undocumented `GET /api/v1/settings/schema`, `GET /api/v1/settings`, and `POST /api/v1/settings` endpoints, including the full set of accepted fields and the transactional apply semantics for topology and output changes. Also fixes a pre-existing typo in the validated-setting endpoint description.

## Contributing requirements

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -96,7 +96,7 @@
 #define DEVICE_CONFIG_FILE          "/device.cfg"
 #define NTP_SERVER_DEFAULT          "0.pool.ntp.org"
 #ifndef BRIGHTNESS_MIN
-    #define BRIGHTNESS_MIN          uint8_t(10)
+    #define BRIGHTNESS_MIN          uint8_t(13)
 #endif
 #ifndef BRIGHTNESS_MAX
     #define BRIGHTNESS_MAX          uint8_t(255)
@@ -192,6 +192,7 @@ class DeviceConfig : public IJSONSerializable
 
     std::vector<SettingSpec, psram_allocator<SettingSpec>> settingSpecs;
     std::vector<std::reference_wrapper<SettingSpec>> settingSpecReferences;
+    std::vector<String> pinSpecStrings;
     size_t writerIndex;
 
     void SaveToJSON() const;
@@ -249,7 +250,7 @@ class DeviceConfig : public IJSONSerializable
     static constexpr const char * WS281xChannelCountTag = "ws281xChannelCount";
     static constexpr const char * WS281xPinsTag = "ws281xPins";
     static constexpr const char * WS281xColorOrderTag = "ws281xColorOrder";
-    static constexpr const char * AudioInputPinTag = "audioInputPin";
+    static constexpr const char * AudioInputPinTag = NAME_OF(audioInputPin);
 
     DeviceConfig();
 
--- a/include/effects/matrix/PatternStocks.h
+++ b/include/effects/matrix/PatternStocks.h
@@ -338,12 +338,18 @@ private:
         // Lazily load this class' SettingSpec instances if they haven't been already
         if (mySettingSpecs.size() == 0)
         {
-            mySettingSpecs.emplace_back(NAME_OF(stockServer), "Stock server location",
-                                        "The host and port of the service that provides stock data.",
-                                        SettingSpec::SettingType::String);
-            mySettingSpecs.emplace_back(NAME_OF(tickerSymbols), "Ticker symbols",
-                                        "Comma-separated list of ticker symbols to show stock data for.",
-                                        SettingSpec::SettingType::String);
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(stockServer),
+                .FriendlyName = "Stock server location",
+                .Description  = "The host and port of the service that provides stock data.",
+                .Type         = SettingSpec::SettingType::String
+            }));
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(tickerSymbols),
+                .FriendlyName = "Ticker symbols",
+                .Description  = "Comma-separated list of ticker symbols to show stock data for.",
+                .Type         = SettingSpec::SettingType::String
+            }));
         }
 
         return &mySettingSpecs;
--- a/include/effects/matrix/PatternStocks.h
+++ b/include/effects/matrix/PatternStocks.h
@@ -347,7 +347,7 @@ private:
             mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
                 .Name         = NAME_OF(tickerSymbols),
                 .FriendlyName = "Ticker symbols",
-                .Description  = "Comma-separated list of ticker symbols to show stock data for.",
+                .Description  = "Comma-separated list of ticker symbols for which to retrieve and display stock data.",
                 .Type         = SettingSpec::SettingType::String
             }));
         }
--- a/include/effects/matrix/PatternSubscribers.h
+++ b/include/effects/matrix/PatternSubscribers.h
@@ -147,30 +147,32 @@ class PatternSubscribers : public EffectWithId<PatternSubscribers>
         // Lazily load this class' SettingSpec instances if they haven't been already
         if (mySettingSpecs.size() == 0)
         {
-            mySettingSpecs.emplace_back(
-                NAME_OF(youtubeChannelGuid),
-                "YouTube channel GUID",
-                "The <a href=\"http://tools.tastethecode.com/youtube-sight\">YouTube Sight</a> GUID of the channel for which "
-                "the effect should show subscriber information.",
-                SettingSpec::SettingType::String
-            );
-            mySettingSpecs.emplace_back(NAME_OF(backgroundColor), "Background Color",
-                                        "Color for the background",
-                                        SettingSpec::SettingType::Color);
-            mySettingSpecs.emplace_back(NAME_OF(borderColor), "Border Color",
-                                        "Color for the border around the edge", SettingSpec::SettingType::Color);
-            {
-                SettingSpec::FinishGuard spec(mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
-                                             "The name of the channel for which the effect should show subscriber information.",
-                                             SettingSpec::SettingType::String));
-                spec->EmptyAllowed = true;
-            }
-            {
-                SettingSpec::FinishGuard spec(mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
-                                             "The name of the channel for which the effect should show subscriber information.",
-                                             SettingSpec::SettingType::String));
-                spec->EmptyAllowed = true;
-            }
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(youtubeChannelGuid),
+                .FriendlyName = "YouTube channel GUID",
+                .Description  = "The <a href=\"http://tools.tastethecode.com/youtube-sight\">YouTube Sight</a> GUID of the channel for which "
+                                "the effect should show subscriber information.",
+                .Type         = SettingSpec::SettingType::String
+            }));
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(backgroundColor),
+                .FriendlyName = "Background Color",
+                .Description  = "Color for the background",
+                .Type         = SettingSpec::SettingType::Color
+            }));
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(borderColor),
+                .FriendlyName = "Border Color",
+                .Description  = "Color for the border around the edge",
+                .Type         = SettingSpec::SettingType::Color
+            }));
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(youtubeChannelName),
+                .FriendlyName = "YouTube channel name",
+                .Description  = "The name of the channel for which the effect should show subscriber information.",
+                .Type         = SettingSpec::SettingType::String,
+                .EmptyAllowed = true
+            }));
         }
 
         return &mySettingSpecs;
--- a/include/effects/matrix/PatternSubscribers.h
+++ b/include/effects/matrix/PatternSubscribers.h
@@ -159,10 +159,18 @@ class PatternSubscribers : public EffectWithId<PatternSubscribers>
                                         SettingSpec::SettingType::Color);
             mySettingSpecs.emplace_back(NAME_OF(borderColor), "Border Color",
                                         "Color for the border around the edge", SettingSpec::SettingType::Color);
-            mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
-                                         "The name of the channel for which the effect should show subscriber information.",
-                                         SettingSpec::SettingType::String)
-                                         .EmptyAllowed = true;
+            {
+                SettingSpec::FinishGuard spec(mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
+                                             "The name of the channel for which the effect should show subscriber information.",
+                                             SettingSpec::SettingType::String));
+                spec->EmptyAllowed = true;
+            }
+            {
+                SettingSpec::FinishGuard spec(mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
+                                             "The name of the channel for which the effect should show subscriber information.",
+                                             SettingSpec::SettingType::String));
+                spec->EmptyAllowed = true;
+            }
         }
 
         return &mySettingSpecs;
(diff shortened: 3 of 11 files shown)


include/effects/matrix/PatternSubscribers.h
```diff
@@ -159,10 +159,18 @@ class PatternSubscribers : public EffectWithId<PatternSubscribers>
...
+                                             "The name of the channel for which the effect should show subscriber information.",
+                                             SettingSpec::SettingType::String));
+                spec->EmptyAllowed = true;
+            }
+            {
+                SettingSpec::FinishGuard spec(mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
+                                             "The name of the channel for which the effect should show subscriber information.",
+                                             SettingSpec::SettingType::String));
+                spec->EmptyAllowed = true;
+            }
```
robertlipe: Is this intentionally doubled?
Could it be less doubled?

include/deviceconfig.h
```diff
@@ -96,7 +96,7 @@
 #define DEVICE_CONFIG_FILE          "/device.cfg"
 #define NTP_SERVER_DEFAULT          "0.pool.ntp.org"
 #ifndef BRIGHTNESS_MIN
-    #define BRIGHTNESS_MIN          uint8_t(10)
+    #define BRIGHTNESS_MIN          uint8_t(13)
```
robertlipe, line 99: Not mentioned int he changelog, so this looks accidental.

include/interfaces.h
```diff
@@ -200,8 +200,7 @@ struct SettingSpec
         Boolean,
         String,
         Palette,
-        Color,
-        Slider
+        Color
```
robertlipe, line 203: Isn't Slider new?

include/interfaces.h
```diff
@@ -220,8 +219,7 @@ struct SettingSpec
         Default,           // type-driven default (Integer -> number input, Boolean -> checkbox, etc.)
         Slider,            // numeric slider; honors DisplayScale and DisplaySuffix
         Select,            // dropdown sourced from inline Options or via OptionsSource
-        IntervalToggle,    // boolean-on + numeric value composite (effectInterval-style)
-        Color              // color picker; raw value is an integer 0xRRGGBB
+        IntervalToggle     // boolean-on + numeric value composite (effectInterval-style)
```
robertlipe, line 222: And wasn't Color new?

include/interfaces.h
```diff
@@ -318,10 +316,16 @@ struct SettingSpec
...
-    std::vector<const char*> InlineOptionLabels = {};
+    // Parallel arrays of raw values and friendly labels for Select options.
+    // For OptionsSource::Inline: the full option list. OptionLabels may be
+    //   empty, in which case values double as labels.
+    // For OptionsSource::SchemaPath: optional label overrides for schema-
+    //   derived values (e.g. "ws281x" -> "WS281x"). Either both arrays are
+    //   empty (no overrides) or both are non-empty and the same length.
+    // For OptionsSource::ExternalTimeZones: optional label overrides for
+    //   specific time zone identifiers. Same empty-or-matched-length rule.
+    std::vector<const char*> OptionValues = {};
```
robertlipe, line 327: Instead of parallel arrays can they be single arrays of std::pair? That might spare us a world of syncrhonization pain. Maybe it's a pair of const char* and a std::optional(const char*).

include/interfaces.h
```diff
@@ -332,28 +336,83 @@ struct SettingSpec
...
     SettingSpec(const char* name, const char* friendlyName, SettingType type);
 
-    // Constructor that sets both minimum and maximum values
+    // Base constructors with min/max range
     SettingSpec(const char* name, const char* friendlyName, const char* description, SettingType type, double min, double max);
-
-    // Constructor that sets both minimum and maximum values
     SettingSpec(const char* name, const char* friendlyName, SettingType type, double min, double max);
 
+    // ---- Constructors for UI-positioned specs (type + section + apiPath) ----
```
robertlipe: Yikes. It's OK to have setters and modifiers. Does eveything HAVE to be in the construction list? Even with completion, these will be overwhelming to a user.

src/deviceconfig.cpp
```diff
@@ -384,324 +384,265 @@ const std::vector<std::reference_wrapper<SettingSpec>>& DeviceConfig::GetSetting
...
+            spec->Section = kSectionSystem;
+            spec->RequiresReboot = true;
+            spec->ApiPath = "device.hostname";
         }
         {
-            auto& spec = settingSpecs.emplace_back(
+            SettingSpec::FinishGuard spec(settingSpecs.emplace_back(
                 PowerLimitTag,
                 "Power limit",
                 "The maximum power in mW that the matrix attached to the board is allowed to use. As the previous sentence implies, this "
```
robertlipe: wait. MINIMUM power? That doesn't make sense. If a device is unplugged, how do you enforce that?

Oh. this is sort of already in the code...but it's not USED that I can tell.

This might be a dead code cleanup opportunity. Maybe.

src/deviceconfig.cpp
```diff
@@ -384,324 +384,265 @@ const std::vector<std::reference_wrapper<SettingSpec>>& DeviceConfig::GetSetting
...
-            spec.Section = kSectionTopology;
-            spec.Priority = 2;
-            spec.ApiPath = "topology.serpentine";
-        }
+        settingSpecs.emplace_back(
+            MatrixWidthTag,
+            "Matrix width",
+            "Active matrix width. WS281x builds validate this by total LED capacity, so width * height must stay within the compiled LED budget.",
+            SettingSpec::SettingType::PositiveBigInteger,
+            1.0, (double)GetCompiledLEDCount(),
```
robertlipe: Embrace new casts..or led editors fix it...or work it out so that the "right" type is just used. Do we need a GetCompileLedCountFloat()?

Also, strongly prefer float to double where we can. Software emulation is awful.

src/deviceconfig.cpp
```diff
@@ -384,324 +384,265 @@ const std::vector<std::reference_wrapper<SettingSpec>>& DeviceConfig::GetSetting
...
-                "Byte order used when streaming RGB values to WS281x LEDs. This applies live on strip builds and is ignored on HUB75 builds.",
-                SettingSpec::SettingType::String
-            );
-            spec.Section = kSectionOutput;
-            spec.Priority = 12;
-            spec.ApiPath = "outputs.ws281x.colorOrder";
-            spec.Widget = SettingSpec::WidgetKind::Select;
-            spec.Options = SettingSpec::OptionsSource::SchemaPath;
-            spec.OptionsSchemaPath = "outputs.ws281x.allowedColorOrders";
+            const auto& nameStr        = pinSpecStrings.emplace_back(str_sprintf("ws281xPin%zu", i));
```
robertlipe, line 667: We could avoid a wrapper and just use asprintf() for these. It should detect the non-constness and take ownership of that allocated object.

src/types.cpp
```diff
@@ -142,6 +167,141 @@ SettingSpec::SettingSpec(const char* name, const char* friendlyName, SettingType
...
+{
+    FinishAndValidateInitialization();
+}
+
+SettingSpec::SettingSpec(const char* name, const char* friendlyName, SettingType type,
+                         const char* section, const char* apiPath, const char* optionsSchemaPath, std::optional<int> priority)
+    : SettingSpec(name, friendlyName, nullptr, type, section, apiPath, optionsSchemaPath, priority)
+{}
+
+// Constructor E: positioned Select widget with non-Inline source
```
robertlipe: Any time we need "ctor E", which has six indistinguishable values, it's time to question our life choices.

Did you consider simpler constructors and named setters for these?

Can we better use the type system? Maybe there's a FloatSetting {float min, float max, bool displayed} or 
enum Valdation{UseValidation = true, DontValidate = false}
or DisplayCategory(const char* name, const chart* apiPath} or... well, _anything_ to break up these long arg lists here and above.

Evenif the classes compile away (and the compiler IS good at this) and you jsut have
SettingSpec (.... NamedPath("Blinkies", "/blinkies/options/all"), which completely dissolves at buid time, it's still way more readable.

If you want to talk through options, lets.

src/webserver.cpp
```diff
@@ -782,36 +779,29 @@ void CWebServer::SendSettingSpecsResponse(AsyncWebServerRequest * pRequest, cons
                 auto options = widget["options"].to<JsonObject>();
                 options["source"] = spec.OptionsSourceName();
 
+                auto addOptionArrays = [&]()
+                {
+                    auto valuesArr = options["values"].to<JsonArray>();
+                    for (auto entry : spec.OptionValues)
+                        valuesArr.add(entry ? entry : "");
```
robertlipe, line 786: This loses our "optional"-ness of these. Can you preserver that up into JsonLand, or is "" just as good as it gets?

robertlipe (changes requested):
Standing n/c on the part aht isn't C++.

I'm concerned abut the explosion of arguments in ctors. Let's find ways to NOT have ctors that take six char*s.

I've also constrained to this PR specifically. I'll admit that it feels like Dave is playing shell games with reviews, merging them into other PRs while there's are still unresolved issues, and then resubmmitting them, so this one being a PR atop of that PR and still having tiny things like changing mininums that looks like they might be accidental makes me nervous. I can no longer tell what's coming and going because we have multiple PRs both adding and deleting the same code. 

Most of my beefs here are that we're not takign the best advantage of the typing system and we're letting everything decay to char*, String (ugh) and bools when better types should be used.

include/effects/matrix/PatternSubscribers.h
```diff
@@ -159,10 +159,18 @@ class PatternSubscribers : public EffectWithId<PatternSubscribers>
                                         SettingSpec::SettingType::Color);
             mySettingSpecs.emplace_back(NAME_OF(borderColor), "Border Color",
                                         "Color for the border around the edge", SettingSpec::SettingType::Color);
-            mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
-                                         "The name of the channel for which the effect should show subscriber information.",
-                                         SettingSpec::SettingType::String)
-                                         .EmptyAllowed = true;
+            {
```
davepl: This code is so nice you called it twice, but did you mean to?

rbergen: Nope. Not sure what happened there.

include/effects/matrix/PatternSubscribers.h
rbergen: No. Yes.

include/deviceconfig.h
rbergen: It isn't. We had a define set to one value (10) when the actual value was changed to something else (13). This fixes that.

include/interfaces.h
rbergen: We now declare a data type and a widget type to edit it. Slider occurred in both, and I decided it was a widget type.

include/interfaces.h
rbergen: Same answer as for Slider, except Color is a true data type.

include/interfaces.h
rbergen: That doesn't translate to JSON, which is something we do have to consider here.

include/interfaces.h
rbergen: This is a valid point. I'm changing the approach on this one in an upcoming commit.

src/types.cpp
rbergen: I'm switching to designated initializers that are passed through a static function for validation purposes.

src/webserver.cpp
rbergen: JsonLand doesn't do optionness, which is exactly the reason for this approach. Sometimes one has to grudgingly go with the least common denominator.

src/deviceconfig.cpp
rbergen: As I recall it, trying to ask a board to operate under a sensibly minimum amount of power is pointless, so we don't pretend otherwise.

In any case, this is not newly introduced in this PR, so I'd consider it a separate discussion.

src/deviceconfig.cpp
rbergen: I'm basically inheriting this from the PR I'm trying to improve. I'll defer to Dave to answer the question on what the appropriate data type is.

src/deviceconfig.cpp
rbergen: Again, I'm inheriting this from the PR I'm augmenting. This is not part of what I'm aiming to revise.

include/interfaces.h
robertlipe: Oh. Gross. It's tempting to treat JSON's problems on the boundaries of JSON, but given our limitations her, that's a reasonable choice. OK.

include/effects/matrix/PatternStocks.h
```diff
@@ -338,12 +338,18 @@ class PatternStocks : public EffectWithId<PatternStocks>
...
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(stockServer),
+                .FriendlyName = "Stock server location",
+                .Description  = "The host and port of the service that provides stock data.",
+                .Type         = SettingSpec::SettingType::String
+            }));
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(tickerSymbols),
+                .FriendlyName = "Ticker symbols",
+                .Description  = "Comma-separated list of ticker symbols to show stock data for.",
```
robertlipe: Nit: Drop the for. Or
"Comma-separated list of ticker symbols of stock data sources"

include/effects/matrix/PatternSubscribers.h
```diff
@@ -147,30 +147,32 @@ class PatternSubscribers : public EffectWithId<PatternSubscribers>
...
-                spec->EmptyAllowed = true;
-            }
-            {
-                SettingSpec::FinishGuard spec(mySettingSpecs.emplace_back(NAME_OF(youtubeChannelName), "YouTube channel name",
-                                             "The name of the channel for which the effect should show subscriber information.",
-                                             SettingSpec::SettingType::String));
-                spec->EmptyAllowed = true;
-            }
+            mySettingSpecs.push_back(SettingSpec::Validate(SettingSpec{
+                .Name         = NAME_OF(youtubeChannelGuid),
```
robertlipe, line 151: Named initializers I can get behind as a reasonable approach. Good call!

robertlipe (approved):
Nice approaches. Thanx.

include/deviceconfig.h
robertlipe: No further action required, but ... I remember adding this. BRIGHTNESS_MIN mattered to me because Tiny at even 10% is uncomfortably bright when next to a keyboard while developing. My builds used "2" successfully, but I don't recall if i've interacted with the javascript slider. But there IS a use for a lower range here.

PR #873, opened 2026-05-12
Add RGBWW support, SK6812 support, warm glow test effect

davepl:
Check-in: RGBCCW (4/5-channel addressable LED) support

Summary

Adds support for 4-channel SK6812 RGBW strips (and the framework for future 5-channel SM16825/WS2805 RGBCCW chips) to NightDriverStrip without touching any existing CRGB-based effects. Existing 3-channel WS2812 builds are unchanged at the source level and pay zero runtime or memory cost for the new code paths. New ships as a working spirallamp_rgbw env targeted at an M5StickC Plus 2 driving two 144-LED SK6812 strips on GPIO 32 and 33.
What's new
CRGBW pixel type (include/crgbw.h) — a 2-byte {cw, ww} struct used as a parallel "whites plane" alongside the existing FastLED CRGB leds[] array. Internal representation always carries both cool-white and warm-white intent; the PixelFormat for each chip decides how to map those onto the strip's actual physical channels.
PixelFormat strategy class (include/pixelformat.h) — the chip-specific policy that converts the (leds, whites) framebuffer pair into the byte stream the wire expects. Abstract base with two concrete subclasses:

Ws2812Format — 3 bytes/pixel, whites ignored. Reproduces the previous PackChannelPixelsForColorOrder behavior exactly.
Sk6812Format — 4 bytes/pixel. Auto-synthesizes the W channel from RGB shared-portion (W = min(R,G,B)) with a configurable extraction ratio (see below), plus saturating-adds any effect-explicit whites and ambient white floor.

The white plane lives on GFXBase as a sibling pointer to leds. Allocated by WS281xGFX only when USE_SK6812=1; left null otherwise, so 3-channel builds incur no memory cost. Reallocated in ConfigureTopology and freed in the destructor.

New API on GFXBase:

setPixelWhite(x, y, cw, ww) — explicit dual-white intent
setPixelCCT(x, y, kelvin, brightness) — Kelvin gets split between CW/WW via SplitByCct helper in crgbw.h

Both no-op when whites is null, so the same source compiles and runs on plain WS2812 builds without conditional branches.

WarmGlowEffect (include/effects/strip/cct_effects.h) — reference CCT-aware effect with a CRGB-fallback path. On builds with a whites plane it drives the dedicated W LED(s) via setPixelCCT and zeros RGB. On builds without one it falls back to a Kelvin-tinted CRGB approximation so warm and cool presets are visibly distinct. Registered into both EFFECTS_DEMO and EFFECTS_SPIRALLAMP at 2700K and 6500K.

What changed in existing code

WS281xOutputManager now holds a unique_ptr<PixelFormat> picked at construction by USE_SK6812. Per-channel byteCount is format->BytesPerPixel() * ledCount instead of ledCount * 3. The Show() per-channel pack loop calls format->Pack(device->leds, device->whites, …) instead of PackChannelPixelsForColorOrder directly. The old template'd PackChannelPixels and color-order switch have been deleted; that logic moved into Ws2812Format::Pack.
WS281xGFX::PostProcessFrame now zeros the tail of whites[] beyond pixelsDrawn the same way it already zeroed the tail of leds[].

EffectManager::StartEffect now zeros the whites plane on every graphics device at every effect transition. GFXBase::Clear, LEDStripEffect::fillSolidOnAllChannels, and LEDStripEffect::setAllOnAllChannels similarly zero the matching range of whites when clearing leds. This fixes a cross-effect bleed where a CCT effect (WarmGlow) would leave W LEDs lit underneath subsequent RGB effects that wrote only to leds[]. Leds[] is intentionally not blanket-cleared on effect switch — effects like FireEffect deliberately persist leds[] state across frames for heat-decay loops.

Configurable white-extract ratio

The shared-portion extraction in Sk6812Format is now a whiteExtractRatio parameter (0..255). Default 128 (50%): half the shared white routes to the dedicated W LED, half stays in RGB. The previous "full extraction" (255) behavior reliably washed out near-white content because the W LED on most SK6812 SKUs is 2-3× brighter per digital count than RGB. Effect-explicit whites from setPixelWhite/setPixelCCT are NOT scaled by this ratio — those route through unchanged.
Overrideable at build time via -DSK6812_WHITE_EXTRACT_RATIO=N in any env's build_src_flags, alongside NIGHTDRIVER_DEFAULT_CCT_KELVIN, NIGHTDRIVER_DEFAULT_AMBIENT_CW, NIGHTDRIVER_DEFAULT_AMBIENT_WW. Plumbing the same knobs through DeviceConfig + SetupUI for runtime control is a deliberate follow-up — easier to design those once we have real RGBCCW hardware to compare against.

New env

[env:spirallamp_rgbw] — extends dev_m5stick_c_plus2. 2 channels of SK6812 RGBW @ 144 LEDs each, pins 32 and 33, USE_SK6812=1, EFFECTS_DEMO=1. Uses EFFECTS_DEMO instead of EFFECTS_SPIRALLAMP because FireFanEffect in the spirallamp set writes through FastLED's global controller registry (FastLED[i][x] = color), which is empty on the NightDriver runtime-managed strip path and null-derefs on the first draw. That's a pre-existing bug in FireFanEffect — it was latent until now because the original spirallamp env uses EFFECTS_MINIMAL. Worth fixing as a separate change before switching back.
What it doesn't do (deliberate)

No DeviceConfig / SetupUI plumbing for the CCT/ambient/extract-ratio knobs yet. Compile-time defaults are tuned reasonably; runtime knobs are the next commit's problem once the SetupUI design has real RGBCCW hardware feedback to design against.
No SM16825Format yet. The abstraction's there; the format is a ~50-line add when we have actual 5-channel hardware in hand. WS2805 is similarly straightforward.
No tone-curve or LUT-based CCT split. SplitByCct is linear between 2700K (pure WW) and 6500K (pure CW), clamped at both ends. On 4-channel SK6812 the linearity doesn't matter because both halves route to the single W LED anyway; matters more for 5-channel chips where CW and WW are physically separate phosphors.

### DIFF
--- a/include/effects/strip/cct_effects.h
+++ b/include/effects/strip/cct_effects.h
@@ -31,43 +31,31 @@
 // unless RGB fallback is being used. Future 5-channel formats can render the
 // CW/WW split directly.
 //
-// On builds without a whites plane, falls back to filling RGB with a coarse
-// Kelvin approximation so warm and cool presets are visibly distinct.
+// This effect is intended as a maximum-output white scene. It fills RGB with
+// the brightest RGB approximation of the requested CCT and, when a whites
+// plane exists, adds the brightest CCT-correct CW/WW mix the hardware can
+// represent.
+//
+// Check your power supplies and power limits!  Monitor temperatures and 
+// current draw when running this effect, especially at high brightness levels.  
+// It can easily draw more current than the strip or power supply can handle, 
+// which can cause damage.  Start with low brightness and work your way up 
+// while monitoring the strip's behavior and temperature.
 
 class WarmGlowEffect : public EffectWithId<WarmGlowEffect>
 {
 protected:
 
     uint16_t _kelvin     = 4000;     // default: neutral-ish single-white intent
-    uint8_t  _brightness = 200;      // 0..255 white intensity
+    uint8_t  _brightness = 255;      // 0..255 scene intensity
 
     static String NameForKelvin(uint16_t kelvin)
     {
         if (kelvin <= 3000)
             return "Warm White Glow";
-        if (kelvin >= 6000)
+        if (kelvin >= 5000)
             return "Cool White Glow";
         return "White Glow";
-    }
-
-    static CRGB ApproximateRgbForKelvin(uint16_t kelvin, uint8_t brightness)
-    {
-        constexpr uint16_t kKelvinWarm = 2700;
-        constexpr uint16_t kKelvinCool = 6500;
-        constexpr uint16_t kKelvinSpan = kKelvinCool - kKelvinWarm;
-
-        const uint16_t clamped = std::min<uint16_t>(std::max<uint16_t>(kelvin, kKelvinWarm), kKelvinCool);
-        const uint16_t coolPart = static_cast<uint16_t>(clamped - kKelvinWarm);
-        const uint16_t warmPart = static_cast<uint16_t>(kKelvinSpan - coolPart);
-
-        const auto mix = [&](uint8_t warm, uint8_t cool) -> uint8_t {
-            const uint32_t channel = (static_cast<uint32_t>(warm) * warmPart)
-                                   + (static_cast<uint32_t>(cool) * coolPart)
-                                   + (kKelvinSpan / 2);
-            return static_cast<uint8_t>((channel / kKelvinSpan) * brightness / 255);
-        };
-
-        return CRGB(mix(255, 205), mix(180, 225), mix(100, 255));
     }
 
 public:
@@ -83,7 +71,7 @@ public:
     WarmGlowEffect(const JsonObjectConst& jsonObject)
         : EffectWithId<WarmGlowEffect>(jsonObject),
           _kelvin    (jsonObject["kelvin"]     | 4000),
-          _brightness(jsonObject["brightness"] |  200)
+          _brightness(jsonObject["brightness"] |  255)
     {
     }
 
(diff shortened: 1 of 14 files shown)


src/ws281xoutputmanager.cpp
```diff
@@ -672,7 +641,47 @@ void WS281xOutputManager::Show(const std::vector<std::shared_ptr<GFXBase>>& devi
...
+        #ifndef NIGHTDRIVER_DEFAULT_AMBIENT_WW
+            #define NIGHTDRIVER_DEFAULT_AMBIENT_WW 0
+        #endif
+        
+        // SK6812_WHITE_EXTRACT_RATIO: 0..255, fraction of shared-portion
+        // white pulled into the dedicated W LED
+        
+        #ifndef SK6812_WHITE_EXTRACT_RATIO
+            #define SK6812_WHITE_EXTRACT_RATIO 96
+        #endif
```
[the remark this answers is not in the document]
davepl: @copilot apply changes based on [this feedback](https://github.com/PlummersSoftwareLLC/NightDriverStrip/pull/873#discussion_r3223607493)

rbergen:
> That's a pre-existing bug in FireFanEffect — it was latent until now because the original spirallamp env uses EFFECTS_MINIMAL. Worth fixing as a separate change before switching back.

Should we maybe create an issue for this?

include/effects/strip/cct_effects.h
```diff
@@ -31,45 +31,33 @@
...
     uint16_t _kelvin     = 4000;     // default: neutral-ish single-white intent
-    uint8_t  _brightness = 200;      // 0..255 white intensity
+    uint8_t  _brightness = 255;      // 0..255 scene intensity
 
     static String NameForKelvin(uint16_t kelvin)
     {
         if (kelvin <= 3000)
             return "Warm White Glow";
-        if (kelvin >= 6000)
+        if (kelvin >= 5000)
```
davepl: I actually consider anything over 3500 to be "cool white" but I don't think most folks do, so....

rbergen: I do think temperatures over ~4000K are generally considered cool white? It so happens I recently bought some fluorescent lights that were 4100K, and they were definitely sold as "cool white". Towards the softer end of that range probably, but still.

src/ws281xgfx.cpp
```diff
@@ -46,6 +46,116 @@
...
+             + ((green * kPowerGreenMw) >> 8)
+             + ((blue * kPowerBlueMw) >> 8)
+             + (kPowerDarkMw * ledCount);
+    }
+
+    uint8_t SaturatingAdd(uint8_t a, uint8_t b)
+    {
+        const uint16_t sum = static_cast<uint16_t>(a) + static_cast<uint16_t>(b);
+        return sum > 255 ? 255 : static_cast<uint8_t>(sum);
+    }
```
rbergen, line 96: This function is also included in pixelformat.h as PixelFormatHelpers::SaturatingAdd. As this file (ws281xgfx.cpp) includes that one, I think we should drop this copy and adopt PixelFormatHelpers::SaturatingAdd throughout.

rbergen: I'm going to resolve this conversation. I've decided I'm going to put a PR together with some follow-up fixes to the recent merges, this being one of them.

rbergen (approved):
LGTM

PR #883, opened 2026-05-29
Bugfixes - PSRAM and Networking

davepl:
Cleaned up the network-fix delta and tightened the WebUI form controls.

Core fixes

Added explicit PSRAM/internal allocation helpers and allocator aliases so large buffers are allocated intentionally instead of through scattered raw allocator calls.
Replaced manual PSRAM allocation/reset patterns with safer unique_ptr helper construction.
Kept socket decompression scratch memory in internal RAM where required.
Reduced lock-order risk in effect manager paths by using std::scoped_lock for dual-lock sections.
Fixed websocket preview packet handling to send from copied color data rather than a potentially unstable source.
Cleaned up several effect/pattern buffer ownership sites to use the new allocation helpers consistently.
WebUI fixes

Added stable id/name values to static and generated form controls.
Removed native input type="color" controls and replaced them with swatch buttons plus numeric color values to avoid browser-generated hidden fields.
Removed the native range slider input and stale range CSS after identifying it as another likely source of generated hidden inputs.
Verified all input/select/textarea elements in the WebUI source now have an id or name.
Cleanup

Removed accidental whitespace/NUL-style churn.
Kept the changes focused around allocation correctness, network stability, and the WebUI autofill warning.


src/webserver.cpp
```diff
@@ -56,12 +57,42 @@
...
+            const unsigned long parsed = strtoul(text.c_str(), &end, 10);
+            if (end && *end == '\0' && parsed <= std::numeric_limits<uint16_t>::max())
+            {
+                target = static_cast<uint16_t>(parsed);
+                return true;
+            }
+        }
+
+        return false;
+    }
```
rbergen: This function is not used anywhere - nor should it be. Which means you're (re)introducing dead code, which I propose we don't.


```suggestion
```

src/webserver.cpp
```diff
@@ -947,21 +978,29 @@ bool CWebServer::BuildSettingSpecsJson(String& json, const std::vector<std::refe
 
 bool CWebServer::EnsureDeviceSettingSpecsJson()
 {
+    std::lock_guard<std::recursive_mutex> guard(g_settingSpecsCacheMutex);
```
rbergen: Use function template argument deduction for brevity.

```suggestion
    std::lock_guard guard(g_settingSpecsCacheMutex);
```

src/webserver.cpp
```diff
@@ -947,21 +978,29 @@ bool CWebServer::BuildSettingSpecsJson(String& json, const std::vector<std::refe
...
     }
 
+    deviceSettingSpecsJson = json;
     debugI("WebServer: cached device setting specs JSON (%zu bytes).", (size_t)deviceSettingSpecsJson.length());
     return true;
 }
 
 const std::vector<std::reference_wrapper<SettingSpec>> & CWebServer::LoadDeviceSettingSpecs()
 {
+    std::lock_guard<std::recursive_mutex> guard(g_settingSpecsCacheMutex);
```
rbergen: Use function template argument deduction for brevity.


```suggestion
    std::lock_guard guard(g_settingSpecsCacheMutex);
```

rbergen (commented):
This feels a bit like a partial revert of the "rely on standard malloc to figure it out" approach we adopted a few weeks ago. Don't get me wrong: if this fixes WiFi drop-outs, I'm all for it.

I made 3 low-impact suggestions, including concrete code change proposals. You can batch and commit them in the PR's "Files changed" tab on GitHub for sure; I've never tried that workflow from within VSCode, but it may also be possible there.

rbergen (approved):
LGTM!

PR #884, opened 2026-05-31
Bugfixes - PSRAM and JSON

davepl:
Changed JSON persistence to use internal RAM for save-time JSON documents. The normal JSON allocator can use PSRAM, but SPIFFS writes can disable cache, so the save path now keeps its working JSON document in internal RAM. This reduces the chance of touching PSRAM during flash/cache-disabled windows.

Wrapped config file writes/removes with the render mutex and, for HUB75 builds, wait for any pending SmartMatrix buffer swap before touching SPIFFS. This covers /device.cfg, /effects.cfg, and /current.cfg. The goal is to avoid flash writes racing with the draw loop or panel refresh state.

Changed JSON/current-effect logging so “bytes written” is logged only after flush/close succeeds. Previously the log could appear before the write was fully settled, which made the crash timing look more directly tied to JSONWriter than it really was.

Added a timeout/yielding wait for SmartMatrix layer swaps. SmartMatrix’s swap path can otherwise spin while waiting for swapPending to clear, starving Idle1 and tripping the task watchdog. The wait now calls delay(1) and times out with a warning instead of hard-spinning forever.

Moved long-lived system services and JSON/network registration entries to internal allocation. These objects contain task/concurrency state and are touched across service boundaries, so they are safer in internal RAM than PSRAM under cache-disabled or ISR-adjacent conditions.

Changed the legacy ESP-IDF 4.x RMT remote driver install to stop using ESP_INTR_FLAG_IRAM. The legacy RMT RX driver creates its ringbuffer with default malloc, which can land in PSRAM after PSRAM-default routing is enabled. If that IRAM ISR fires while cache is disabled, it can crash in xRingbufferSendFromISR(). Remote input can tolerate dropped frames during flash writes, so keeping the ISR out of cache-disabled windows is the safer tradeoff.

### DIFF
--- a/src/effectmanager.cpp
+++ b/src/effectmanager.cpp
@@ -663,8 +663,16 @@ void WriteCurrentEffectIndexFile()
     std::lock_guard renderPause(g_render_mutex);
 
     #if USE_HUB75
-        HUB75GFX::WaitForMatrixSwap();
-    #endif
+    // Capture the current effect index without holding g_render_mutex to avoid nested lock order issues
+    size_t currentIndex = g_ptrSystem->GetEffectManager().GetCurrentEffectIndex();
+
+    {
+        std::lock_guard renderPause(g_render_mutex);
+
+        #if USE_HUB75
+            HUB75GFX::WaitForMatrixSwap();
+        #endif
+    }
 
     SPIFFS.remove(CURRENT_EFFECT_CONFIG_FILE);
 
@@ -676,7 +684,7 @@ void WriteCurrentEffectIndexFile()
         return;
     }
 
-    auto bytesWritten = file.print(g_ptrSystem->GetEffectManager().GetCurrentEffectIndex());
+    auto bytesWritten = file.print(currentIndex);
 
     file.flush();
     file.close();
--- a/src/effectmanager.cpp
+++ b/src/effectmanager.cpp
@@ -651,11 +651,21 @@ void RemoveEffectManagerConfig()
 {
     RemoveJSONFile(EFFECTS_CONFIG_FILE);
     // We take the liberty of also removing the file with the current effect config index
+    std::lock_guard renderPause(g_render_mutex);
+    #if USE_HUB75
+        HUB75GFX::WaitForMatrixSwap();
+    #endif
     SPIFFS.remove(CURRENT_EFFECT_CONFIG_FILE);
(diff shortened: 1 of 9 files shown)


src/effectmanager.cpp
```diff
@@ -667,11 +677,12 @@ void WriteCurrentEffectIndexFile()
     }
 
     auto bytesWritten = file.print(g_ptrSystem->GetEffectManager().GetCurrentEffectIndex());
-    debugI("Number of bytes written to file %s: %zu", CURRENT_EFFECT_CONFIG_FILE, (size_t)bytesWritten);
 
     file.flush();
     file.close();
 
+    debugI("Number of bytes written to file %s: %zu", CURRENT_EFFECT_CONFIG_FILE, (size_t)bytesWritten);
```
rbergen: I took a look at the Copilot comment, and if we want to prevent crashes by edge cases then I think we should also try to prevent introducing deadlocks in the process. 
I asked Codex 5.3 for some input, and it came up with the following alternative implementation, which looks very similar to the one in `RemoveEffectManagerConfig()`.

```suggestion
    // Capture the current effect index without holding g_render_mutex to avoid nested lock order issues
    size_t currentIndex = g_ptrSystem->GetEffectManager().GetCurrentEffectIndex();

    {
        std::lock_guard renderPause(g_render_mutex);

        #if USE_HUB75
            HUB75GFX::WaitForMatrixSwap();
        #endif
    }

    SPIFFS.remove(CURRENT_EFFECT_CONFIG_FILE);

    File file = SPIFFS.open(CURRENT_EFFECT_CONFIG_FILE, FILE_WRITE);

    if (!file)
    {
        debugE("Unable to open file %s for writing!", CURRENT_EFFECT_CONFIG_FILE);
        return;
    }

    auto bytesWritten = file.print(currentIndex);

    file.flush();
    file.close();

    debugI("Number of bytes written to file %s: %zu", CURRENT_EFFECT_CONFIG_FILE, (size_t)bytesWritten);
```

rbergen (commented):
Left one comment with an alternative approach for the locking in `WriteCurrentEffectIndexFile()`, in response to Copilot's comment.

I do hope we manage to get back to a stable runtime state with this.

rbergen (approved):

PR #886, opened 2026-05-31
Activeline _ Add a WiFi activity output signal option, allow for zero effects in main table

davepl:
Adds a new optional WIFI_ACTIVITY pin that goes HIGH during WiFi drawing.  Also allows projects to contain zero active effects such that they render no output (and signal no activity) until wifi drawing goes active.

* [X] I read the contribution guidelines in [CONTRIBUTING.md](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/CONTRIBUTING.md).
* [X] I understand the BlinkenPerBit metric, and maximized it in this PR.
* [X] I selected `main` as the target branch.
* [X] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).

PR #887, opened 2026-06-01
Activeline _ Add a WiFi activity output signal option, allow for zero effects in main table

rbergen:
Adds a new optional WIFI_ACTIVITY pin that goes HIGH during WiFi drawing. Also allows projects to contain zero active effects such that they render no output (and signal no activity) until wifi drawing goes active.

This PR replaces #886 with a clean branch rebased from current upstream/main and only the intended commits replayed, to avoid re-reviewing already-merged shared history.


rbergen (commented):
LGTM!

PR #888, opened 2026-06-02
Refactor ownership boundaries and split large compile units

rbergen:
## Summary

This PR is a maintainability refactor focused on ownership clarity, consistency, and compile-unit hygiene.  
It is intended to be behavior-preserving.

The core goals were:

1. Move DeviceConfig domain knowledge out of webserver-centric code and into DeviceConfig-owned code.
2. Make result reporting patterns consistent across settings/update paths.
3. Make JSON setting-value parsing and validation flow consistent.
4. Split large compile units and oversized headers into focused modules.
5. Keep runtime behavior stable while reducing coupling and improving code navigation.

## What Changed

1. DeviceConfig ownership refactor
   - Consolidated configuration-domain logic into DeviceConfig-focused compilation units.
   - Reduced cross-layer leakage from web transport/controller code into configuration internals.
   - Added clearer structure for unified settings handling, validation, and settings spec ownership.

2. Standardized result reporting
   - Aligned internal result/status handling across related update/configuration paths.
   - Reduced one-off reporting patterns in favor of a more uniform approach.

3. Standardized JSON settings parsing
   - Consolidated JSON value parsing/validation behavior and error pathways.
   - Reduced duplicated parsing logic and mismatched edge-case handling.

4. Large CU decomposition
   Split previously large implementation files into responsibility-focused units, including:
   - Effect manager lifecycle/access/playback/persistence responsibilities.
   - Webserver effects/settings/unified handling responsibilities.
   - Gfx/noise/palette/transforms responsibilities.
   - Sound analyzer input init/sampling responsibilities.

5. fan_effects header decomposition
   - Split the large fan effects header into logical sub-headers by concern.
   - Improved readability and maintainability while keeping behavior intact.

6. Follow-on structural alignments
   - Updated adjacent interfaces and includes needed to support the new ownership boundaries.
   - Included small platform/config adjustments needed to keep all targets building.

## Why This Should Not Change Behavior

- The changes are primarily relocations and reorganizations of existing logic.
- Functional pathways were preserved while responsibilities were moved to more appropriate modules.
- No intentional feature changes were introduced.

## Validation

1. Local verification
   - Local PlatformIO build of mesmerizer environment passed during the refactor workflow.

2. CI verification
   - Branch CI completed successfully for this head commit.
   - Matrix builds passed.

## Risk Assessment

- Main risk is structural regression from code movement (linkage/include/order issues), mitigated by local build + full CI matrix pass.
- No known intentional runtime behavior changes.

## Reviewer Notes

- This PR is large by line movement because of CU/header splits.
- Recommended review strategy:
  1. Review by subsystem (DeviceConfig, EffectManager, webserver split, gfx split, fan effects split).
  2. Prefer semantic/ownership checks over raw line churn.
  3. Focus on boundary correctness and consistency patterns rather than feature deltas.

## Contributing requirements

* [x] All code herein is subjected to the license terms in [COPYING.txt](http://github.com/PlummersSoftwareLLC/NightDriverStrip/blob/main/COPYING.txt).* [ ]


config/partitions_custom_feather.csv
rbergen, line 1: Needed because some of the feather builds just exceeded the app partition size.

config/timezones.json
rbergen, line 1: (Ir)regular update of the time zone file, reflecting changes in the contents of the Ubuntu Linux time zone tree.

config/partitions_custom_feather.csv
```diff
@@ -0,0 +1,7 @@
+# Standard 4M partition layout with OTA, tuned for Feather S3 firmware size
```
robertlipe, line 1: Note that this breaks OTA. 

We (I) have a plan to break the entire world in this regard in the near future. We should try to minimize the number of times we do this. 

Note that A3 requures a minimum of three sectors for nvs, so bumping that 0x2000 to 0x3000 might minimize some churn.

Consider if we really want MORE tables, per board, or if this should just be how all 4M boards are.

include/effects/strip/fan_geometry.h
```diff
@@ -0,0 +1,290 @@
...
+        case Sequential:
+        default:
+            return fanBase + fanPos;
+    }
+}
+
+// Clears pixels logically into a fan bank in a selected direction.
+inline void ClearFanPixels(float fPos, float count, PixelOrder order = Sequential, int iFan = 0)
+{
+    fPos += iFan * FAN_SIZE;
```
robertlipe, line 119: We and/or FastLED have optimized methods elsewhere for setting a range.

include/effects/strip/fan_geometry.h
```diff
@@ -0,0 +1,290 @@
...
+{
+    return g_aRingSizeTable[iRing];
+}
+
+inline int GetFanIndex(float fPos)
+{
+    return fPos / FAN_SIZE;
+}
+
+inline int GetRingIndex(float fPos)
```
robertlipe, line 139: Some of these could be small maps instead of computed on every call.

include/deviceconfig.h
```diff
@@ -394,11 +418,23 @@ class DeviceConfig : public IJSONSerializable
...
+    SuccessResultWithMessage ValidateRuntimeConfig(const RuntimeConfig& config) const;
+    SuccessResultWithMessage ParseAndValidateUnifiedSettings(JsonObjectConst root, UnifiedSettingsRequest& out) const;
+    SuccessResultWithMessage SetRuntimeConfig(const RuntimeConfig& config, bool skipWrite = false);
+    SuccessResultWithMessage ApplyUnifiedDeviceSettings(const UnifiedSettingsRequest& request);
+    void SerializeUnifiedSettings(JsonObject root) const;
+    void SerializeUnifiedSettingsSchema(JsonObject root) const;
     void SetAudioInputPin(int newAudioInputPin);
+
+  private:
+    static ResultWithMessage<std::optional<int>> ResolveUnifiedAudioInputPin(JsonObjectConst device);
```
robertlipe, line 434: If all the class data is static, should the instance just be a single tone?

include/improvserial.h
```diff
@@ -190,29 +196,55 @@ class ImprovSerial
...
         }
 
         void log_write(std::vector<uint8_t>& data)
         {
+            // Hex dump output is significantly larger than input bytes.
+            RotateImprovLogIfNeeded((data.size() * 4) + 64);
+
             auto file = SPIFFS.open(IMPROV_LOG_FILE, FILE_APPEND);
+            if (!file)
+                return;
```
robertlipe, line 247: Is this worth a logE?

src/deviceconfig_internal.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace
+{
+    constexpr const char* kRecompileNeededMessage = "recompile needed";
+
+    // Maps the compile-time COLOR_ORDER macro to DeviceConfig runtime enum values.
+    constexpr DeviceConfig::WS281xColorOrder ToRuntimeColorOrder(EOrder order)
+    {
+        switch (order)
+        {
+            case EOrder::RGB: return DeviceConfig::WS281xColorOrder::RGB;
```
robertlipe, line 26: A map might be a better date structure

src/effectmanager_access.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+    return *_gfx[iChannel];
+}
+
+bool EffectManager::IsEffectEnabled(size_t i) const
+{
+    std::lock_guard effectGuard(g_effect_manager_mutex);
+    if (i >= _vEffects.size())
+    {
+        debugW("Invalid index for IsEffectEnabled");
```
robertlipe, line 96: debugE

src/effectmanager_access.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // debug, and render tasks can otherwise hold an iterator/reference while
+    // another task reorders or deletes effects.
+    std::lock_guard effectGuard(g_effect_manager_mutex);
+    return _vEffects;
+}
+
+std::shared_ptr<LEDStripEffect> EffectManager::EffectAt(size_t index) const
+{
+    std::lock_guard effectGuard(g_effect_manager_mutex);
+    return index < _vEffects.size() ? _vEffects[index] : nullptr;
```
robertlipe, line 114: Isn't this .at()

src/effectmanager_persistence.cpp
```diff
@@ -0,0 +1,285 @@
...
+    {
+        if (file.size() > 0)
+        {
+            debugI("Attempting to read file %s", CURRENT_EFFECT_CONFIG_FILE);
+
+            auto valueString = file.readString();
+
+            if (!valueString.isEmpty())
+            {
+                index = strtoul(valueString.c_str(), nullptr, 10);
```
robertlipe, line 85: Consider fromchars

src/effectmanager_persistence.cpp
```diff
@@ -0,0 +1,285 @@
...
+void SaveEffectManagerConfig()
+{
+    if (l_EffectManagerInitializing)
+        return;
+
+    debugV("Saving effect manager config...");
+    g_ptrSystem->GetJSONWriter().FlagWriter(l_EffectsManagerJSONWriterIndex);
+}
+
+void RemoveEffectManagerConfig()
```
robertlipe, line 114: Do you want my pr that centralizes all the fs into one class (I did it for littlefs) so you can centralize this halt the blinkies logic? It's not complicated..

src/effectmanager_runtime.cpp
```diff
@@ -0,0 +1,645 @@
...
+    }
+}
+
+void EffectManager::DisableEffect(size_t i, bool skipSave)
+{
+    std::scoped_lock guard(g_render_mutex, g_effect_manager_mutex);
+
+    if (i >= _vEffects.size())
+    {
+        debugW("Invalid index for DisableEffect");
```
robertlipe, line 329: That's an error

src/effectmanager_runtime.cpp
```diff
@@ -0,0 +1,645 @@
...
+
+// Creates a copy of an existing effect in the list. Note that the effect is created but not yet added to the effect list;
+//   use the AppendEffect() function for that.
+std::shared_ptr<LEDStripEffect> EffectManager::CopyEffect(size_t index)
+{
+    std::scoped_lock guard(g_render_mutex, g_effect_manager_mutex);
+
+    if (index >= _vEffects.size())
+    {
+        debugW("Invalid index for CopyEffect");
```
robertlipe, line 423: Error

src/effectmanager_runtime.cpp
```diff
@@ -0,0 +1,645 @@
...
+}
+
+// Deletes an effect from the effect list. Note that core effects cannot be deleted.
+bool EffectManager::DeleteEffect(size_t index)
+{
+    std::scoped_lock guard(g_render_mutex, g_effect_manager_mutex);
+
+    if (index >= _vEffects.size())
+    {
+        debugW("Invalid index for DeleteEffect");
```
robertlipe, line 461: Check code throughout

robertlipe (approved):
It's really hard to review something this size that's a mixture of new code and moved code, so I kind of phoned it in with a mixture of close scrutiny and skimming. Ultimately it seems ok.

config/partitions_custom_feather.csv
rbergen: > Note that this breaks OTA.

How is that?
 
> We (I) have a plan to break the entire world in this regard in the near future. We should try to minimize the number of times we do this. 
> Note that A3 requures a minimum of three sectors for nvs, so bumping that 0x2000 to 0x3000 might minimize some churn.
> Consider if we really want MORE tables, per board, or if this should just be how all 4M boards are.

My reasoning was that a board-specific table would limit the impact on what I believe to be more common boards for now, while we work to the more massive overhaul you're referring to.

As for the three sectors for nvs, the current 0x2000 is a carbon copy of the "regular" 4M layout that was used for the Feather before. Does that mean you're saying nvs hasn't been working with that layout either?

include/effects/strip/fan_geometry.h
rbergen: I made a very deliberate choice to only move code where that is enough for things to end up in its right place, and make minimal changes beyond that where consistency demands it (which should be the limit of the "new code" you mentioned earlier). 

As I explained in the description, this is a PR focused at improving maintainability - and that more specifically from an overall maintainer's perspective -, and with that it is explicitly steering clear from also implementing other improvements. 

As you stated yourself, the PR is involved enough as it is without adding that layer on top of what is already here.

I acknowledge that does not comply with the Boy Scout Rule, and I make that decision consciously.

For the rest of my responses to your review, I will mark all comments that I believe this decision applies to as N/A.

include/deviceconfig.h
rbergen: I'm not sure I understand the "all the class data is static" remark. These are not member variables but static member functions, and they are part of the DeviceConfig class, which certainly does include instance variables.

include/improvserial.h
rbergen: Possibly. Then again, with a log file with only one purpose (temporary low-level debugging) not being opened, the process the file is generated for should surface quite quickly that that didn't happen.

src/deviceconfig_internal.cpp
rbergen: N/A

src/effectmanager_access.cpp
rbergen: N/A, and I disagree. This is a message notifying about a checked input being invalid (which would normally come in via the REST API and should therefore be expected to be junk), which is not an internal error condition.

src/effectmanager_access.cpp
rbergen: N/A

src/effectmanager_persistence.cpp
rbergen: N/A

src/effectmanager_persistence.cpp
rbergen: N/A based on earlier considerations, with the added side note that I am really only copying code Dave introduced literally one or two PRs ago. 

That said, I think the answer is "yes" from a follow-up perspective.

src/effectmanager_runtime.cpp
rbergen: No, for the same reason as before (API input, so junk is the basic expectation).

src/effectmanager_runtime.cpp
rbergen: Once more, no for reasons of source of parameter.

src/effectmanager_runtime.cpp
rbergen: Decision will be consistent with what I've indicated before.

rbergen (commented):
> It's really hard to review something this size that's a mixture of new code and moved code, so I kind of phoned it in with a mixture of close scrutiny and skimming. 

I know. And after working through a range of PRs that suffered from the same fundamental quality I figured I'd make some big changes that would benefit overall maintainability, fast. I accept arguments against that exist.

> Ultimately it seems ok.

Thank you!

include/effects/strip/fan_geometry.h
rbergen: N/A

config/partitions_custom_feather.csv
robertlipe: Many _other_ builds are so close to the edge (O(hundreds of bytes)) that any upstream change may put them over the edge.

It's not a reason to NOT change it; it's a reason to change everything in lockstep.

config/partitions_custom_feather.csv
robertlipe: I wasn't clear. I reviewed that on a phone. (Which was a truly terrible idea...)

Changing the partition table breaks OTA. There are cases that might work, but there are cases that don't. It's not a reason to NOT change it, but it's worth stating that this is a breaking change.
As long as you're on the unsupported/abandoned Arduino2/ESP-IDF4 path, 0x2000 works. Move to Arduino3 and it simply won't boot any more because NVS needs to be able to erase/flash a page at a time and having two pages to work with is problematic. That inevitable other growth is one of the reasons we've earlier discussed what parts of the generated binaries are -actually- dead weight and can be potentially throw overboard.

My suggest is that if you're going to break it, break it in a way more consistent with future directions. It's not like the data partition in any of our strip builds has any non-trivial size to it. Many of our builds are teetering on the edge. I added  in an experiment a few months ago and pretty much every build broke.

Having the luxury of not caring about installed base or OTA from this path, I've change nvs size AND filesystem  to the better performing, actively maintained LittleFS.

include/effects/strip/fan_geometry.h
robertlipe: Yeah, and that's exactly why I gave up on actually reading (and I mean "intellectualizing", not just parsing as valid c++) this review. Maybe I should have given up sooner because you DID warn about this.

I get the spot you're in, but figuring out what was new and what was being moved wasn't practical, so actually participating in a non-clerical review was wasting everyone's time.

This would never ever be accepted into the tree from anyone without write permissions. YOu'd have made a submitter bust this up onto about half a dozen different PRs, with most of them including the pinky swear "this only moves code" and maybe two of them actually soliciting input.

SPlitting up fan effects.
TZ
Field Access
WebServer
Sound
DeviceConfig
ErrorHandling

I understand that the BSR is really under challenge here these days, but how many times did you encourage me to set better examples for other submitters?

robertlipe (approved):

config/partitions_custom_feather.csv
rbergen: I see your point. But what you're also describing is that this requires a far more in-depth approach, with way farther-reaching impact, to make it truly future-proof - and that it requires knowledge about A3, for one, that I simply don't (yet) have.
That definitely goes against the premise that this PR tries to be comprehensive, but only concerning a pretty tightly-defined conceptual scope.

I think my only sin against that premise is the timezones.json update. I plead guilty, and I only did that because it was way overdue.

I made this particular change to make 2 failing Feather builds work, which aligns with the PR's premise "of applying focus on a couple of things, with minimal supporting changes".

We can "promote" this change to all 4M devices, but then we break OTA for all of them. I'm not sure that makes more sense.

include/effects/strip/fan_geometry.h
rbergen: It's fair pushback, Robert. To some extent you can consider it an act of corrective desperation. 

One other aspect is that I know the smaller PRs that would be the result of a split of this one would continue to play hop-skip with other PRs that are being opened; I already had to rebase this twice in the few days I was working on it, and now #889 is confirming the pattern.

I do believe for both sides of those merge conflicts it's better to move the sheets of ice under our feet once, and then move on.

rbergen:
Mainly for the record, some clarifying comments about the PR scope:

- The first three points in the core goals mentioned are really one: sanitize the distribution of responsibilities across DeviceConfig and CWebServer, and make the implementation signatures of those responsibilities consistent.
- The fourth is related to that, in the sense that while working on the first three, I found navigating some source files that were involved (primarily deviceconfig.cpp and webserver.cpp, but actually also effectmanager.cpp) became problematic due to sheer size. I did decide to then extend the scope of the split-up to all files with more than 1000 lines in principle - I picked that magic number because we humans still have 10 fingers. This pulled in gfxbase.cpp, soundanalyzer.cpp, and faneffects.h.
  I'll note that it had already occurred to me before that these files had become too large from a standpoint of maintainability and showed unhealthy intermingling of concerns in certain cases - I never before experienced a sufficiently pressing need to do something about it.

As already mentioned in an earlier comment, my sin and guilty pleasure in this context is timezones.json. Mea culpa.

robertlipe:
I could type a bunch (more) stuff that you already know but will try to
keep that to a minimum because I know we've all reached a breaking point
and focus mostly on the single remaining technical point.

> an act of corrective desperation.

In a war of escalating merge sizes, there are no winners.

I don't consider my own almost two-month-old crash fix PR—originally a
victim of and later a contributor to this style of escalation—to be a
winner in this game, either.




If the TZ table was what pushed the size over the limit, ask if it's worth
breaking OTA for those users. What if that user is "Dave on a ladder"—a
case that used to be important to appear in my design docs for better OTA?
(I have no idea if Dave has THESE in his awning, but we know he likes
Feathers...and we shouldn't be building for *just* Dave.) These are going
to HAVE to be broken again "soon."  I've been sitting on the NVS resize
until I can batch it with the filesystem change specifically to reduce the
number of potoential times up the ladder. (I have ladder  configurations
myself...)

If that's it, is it worth the thrash? Remember that "it's ain't broken"
discussion? *Nobody* is complaining about not honoring last year's TZ
changes.

TBC: An OTA writes to the inactive partition and once it successfully
boots[1] , it changes the otadata partition (not the partition table, which
is written only via direct connect) to reflect that the new partition is
good; The partition table is not changed by the OTA. Your change won't
actually change the size of the partitions, and the writes will either fail
or (gasp) overwrite the other partition or sail off the end of storage.
Neither case is great.

If you value OTA (and, hey, IDC if you don't - I know where my ladder
is...) {
  Find the straw on the camel's back (sounds like it's TZData) and ask if
it's worth it or if there are other areas - like the ones we discussed -
that can be shaved out.
  Still no dice? Document the heck out of it and be prepared for the
possibility of hate mail from people that don't notice it and will break
their OTAs.
  As counterpoint, we reluctantly broke Mesmerizer a few times based on the
premise that those were paying customers who were probably NOT building
from source, and people didn't come after us (here) with torches.
}

As an OS guy, I fight pretty hard to avoid rendering devices unbootable,
but this fight isn't mine.

[1] My 'ladder' fork changes the definition of 'successful' because I
introduced a hang and it failed to roll back

PR #891, opened 2026-06-04
Adapt core #837 refactors to current main with safe SIOF fixes, audit integration, and build-flag cleanup

rbergen:
## Summary

This PR ports the safe, high-value intent of #837 onto the current main branch state.

It includes five focused commits (followed by a clean-up one) that:
- apply static-initialization-order safety improvements in webserver and effect paths,
- carry forward include/dependency hygiene updates that are compatible with current main,
- integrate coding-standard audits into the PlatformIO build flow,
- add an opt-in strict audit mode for local and CI workflows,
- and apply the build-flag cleanup from #837.

## Relationship to #837: What Was Implemented

The following concepts from #837 are implemented in this PR, adapted to the current main codebase:

1. Static initialization safety improvements
   - Webserver and polar-map initialization paths were refactored to avoid static initialization order hazards.
   - Effect metadata/decoder initialization was also refactored to remove startup-order risks.

2. Build-time coding-standard auditing
   - The audit script is wired into PlatformIO pre-build execution so audits run consistently with builds.

3. Include/dependency hygiene and build-flag cleanup
   - The include-order and dependency-order cleanup that remains valid on current main was applied.
   - The base debug-flag arrangement from #837 was applied (moving stronger debug info to source flags and removing the forced global string.h include).

## Implementation Deviations from #837 (Beyond Branch Drift)

These are intentional deviations from #837 itself, not just consequences of old-base drift:

1. Effect resource ownership model
   - Instead of moving heavy effect resources to per-instance class members, this PR uses shared function-local statics for those resources.
   - Rationale: preserves SIOF safety while avoiding duplicate heavy allocations across instances.

2. Audit enforcement policy
   - #837 enforced audits as hard-fail by default.
   - This PR keeps default behavior as soft-fail and introduces opt-in hard-fail via local configuration.
   - Rationale: enables gradual enforcement and avoids immediately breaking existing workflows while still supporting strict mode when desired. The code in this PR passes audit and strict mode will become the CI policy on the short term.

## What Was Not Included from #837, and Why

1. Allocator-policy cleanup changes
   - Changes that remove/customize allocator pathways and JSON allocation behavior were not merged.
   - Rationale: recent experiences have shown that these are behavior-sensitive memory-policy decisions that can lead to significant functional breakages and need specific and deep validation.

2. Additional runtime behavior tweaks bundled with janitorial edits
   - Non-essential behavioral tuning (for example, threshold-level runtime warning changes) was not carried over.
   - Rationale: out of scope for this port and not required to realize the core safety/build intent.

3. Old-base environment/config churn
   - Broad environment/config differences present due old branch ancestry were not imported.
   - Rationale: unrelated to the core intent and risky to bring forward without dedicated review.

## Validation

- Local builds were successfully run on the adapted changes on envs mesmerizer, spectrum, m5demo
- A full CI run was successfully executed on the PR head branch

## Notes
- Supersedes #826 and #837
- Closes #825
- Co-authored by: Robert Lipe <robertlipe@gmail.com>, zackees <z@zackees.com>

## Contributing requirements


include/webserver.h
```diff
@@ -240,14 +239,12 @@ class CWebServer : public IService
             response->addHeader("Cache-Control", "no-store, max-age=0");
             response->addHeader("Pragma", "no-cache");
 
-            AddCORSHeaderAndSendResponse(request, response);
+                        this->AddCORSHeaderAndSendResponse(request, response);
```
robertlipe: Is this a funky indent or is the diff just mangled?

robertlipe (approved):
Making the audits optional means that they're going to be ignored by submitters and will thus just continue to decay. I don't know why you worked so hard to NOT just turn it on and leave it on. "Those smoke detectors keep going off. Let's remove the batteries."

This PR is still a solid improvement as far as it goes. Thanx.

include/webserver.h
rbergen: Funky indent. Thanks!

rbergen (commented):
Audits will be enforced in CI soon as I indicated in the PR description, and there is a reason for making that a two-step process, which relates to a smooth adoption in the broader context.

As for the ability to make them not abort the build locally, that relates to different people preferring different development workflows. Which is fine - in the end CI will decide what passes the gauntlet and what doesn't.

In any case, thanks for your review!

PR #892, opened 2026-06-05
Accomodate all effects usefully on 48x16 displays

davepl:
This reworks some of the larger effects like weather, subscribers, and stocks, to work usefully on a 48x16 display.  

I also fixed an effects manager load issue - I think it's been around a long time as a crash shortly after load if the effects set changes, or it doesn't like the data it finds.

The EffectManager loads its saved effect list from SPIFFS, specifically /effects.cfg.

What happened:

On startup, InitEffectsManager() tried to load /effects.cfg.
Because the file was from an older build, its persisted effect-set hash no longer matched the current effect factory hash.
That made LoadEffectsJSONFile() reject the old config and return no JSON. That part is intentional.
The EffectManager then loaded the default effects.
Since no valid persisted effects were loaded, the code tried to save the newly generated default effect list back to /effects.cfg.
That save happened through JSONWriter, asynchronously, shortly after startup.
The crash stack showed JSONWriter -> SaveToJSONFile("/effects.cfg") -> file.flush() -> SPIFFS_write -> SPIFFS garbage collection.
During that SPIFFS write/GC, flash/cache state collided with other active startup work and the ESP32 hit: Cache disabled but cached memory region accessed.
So the old storage mattered because it forced the fallback-to-defaults path, which triggered a fresh /effects.cfg rewrite. The crash itself happened during SPIFFS persistence, not during JSON parsing and not directly in NVS.

The fix I made changes the fallback case: when persisted effects are missing or stale, we write the default effect config synchronously inside InitEffectsManager() before the rest of the high-activity startup tasks are running. Normal later effect changes still use the JSON writer path.  

- Dave

### DIFF
--- a/include/deviceconfig.h
+++ b/include/deviceconfig.h
@@ -102,8 +102,13 @@
 #ifndef BRIGHTNESS_MAX
     #define BRIGHTNESS_MAX          uint8_t(255)
 #endif
-#define POWER_LIMIT_MIN             1000
-#define POWER_LIMIT_DEFAULT         4500
+#define POWER_LIMIT_MIN             1
+#define POWER_LIMIT_LEGACY_DEFAULT  4500
+#ifdef POWER_LIMIT_MW
+    #define POWER_LIMIT_DEFAULT     POWER_LIMIT_MW
+#else
+    #define POWER_LIMIT_DEFAULT     POWER_LIMIT_LEGACY_DEFAULT
+#endif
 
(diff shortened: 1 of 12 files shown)


robertlipe (commented):
This is several PRs all in one, but overall OK.

davepl:
I’m the power of ten men, but I work in a single enlistment.

rbergen (approved):
LGTM

PR #893, opened 2026-06-06
Hiccup - Render thread could block on JSON writer save

davepl:
I changed persistence so SPIFFS writes no longer hold the render mutex:
src/jsonserializer.cpp (line 355): SaveToJSONFile() and RemoveJSONFile() now use a dedicated JSON filesystem mutex instead of g_render_mutex.
src/jsonserializer.cpp (line 397): added WaitForRenderSwapBeforeFilesystemWrite() so HUB75 still waits for an active swap to settle without blocking future frames on file I/O.
src/effectmanager_persistence.cpp (line 124): /current.cfg writes now use the JSON filesystem mutex, not the render mutex.
include/jsonserializer.h (line 90): exposed the persistence mutex/helper for the current-effect writer.


davepl:
The render mutex around SPIFFS writes was a defensive crash mitigation.  The original concern was ESP32 flash behavior: SPIFFS writes can disable cache, and cache-disabled windows are dangerous if another task/ISR touches code/data in flash or PSRAM. Around that same work, JSON save documents were moved into internal RAM and HUB75 writes waited for SmartMatrix buffer swaps. The intent was: “pause drawing while flash is being written so render/panel code doesn’t race a cache-disabled SPIFFS operation.”

That was a blunt guard. It helped reduce one suspected class of crashes, but it was not complete protection because other tasks can still run during SPIFFS writes: WiFi, AsyncTCP, audio, remote input, etc. So g_render_mutex could block rendering and cause the visible 1-second autosave hiccup, while still not guaranteeing immunity from cache-disabled crashes.

The current fix keeps the parts that make sense:

Serialize SPIFFS JSON writes with a dedicated filesystem mutex.
Keep the HUB75 “wait for active swap to settle” step.
Stop using the render mutex as a file-I/O guard.

rbergen (approved):
LGTM

PR #899, opened 2026-06-16
Merge branch 'main' of github.com:PlummersSoftwareLLC/NightDriverStrip

davepl:
Updates the stock server to show the correct, live data, and uses the new V2 api to fetch it.

### DIFF
--- a/include/effects/matrix/PatternStocks.h
+++ b/include/effects/matrix/PatternStocks.h
@@ -46,6 +46,7 @@
 #include <thread>
 #include <vector>
 #include <map>
+#include <mutex>
 #include <sstream>
 
 #include "formatsize.h"
@@ -171,7 +172,16 @@ public:
     float volume = 0.0f;
 
     bool hasCurrentPrice = false;
+    bool hasPreviousClose = false;
+    bool hasOpen = false;
+    bool hasHigh = false;
+    bool hasLow = false;
+    bool hasClose = false;
+    bool hasOfficialClose = false;
     bool hasRegularClose = false;
+    bool hasChange = false;
+    bool hasChangePercent = false;
+    bool hasVolume = false;
 
     vector<StockPoint> points;
 
@@ -230,6 +240,8 @@ private:
     system_clock::time_point nextFetch = system_clock::now();  // Time of last quote pull
 
     std::map<String, StockData> stockData;  // map of stock symbols to quotes
+    mutable std::mutex stockDataMutex;
+    std::mutex quoteFetchMutex;
 
     using StockDataCallback = function<void(const StockData&)>;
 
@@ -247,17 +259,21 @@ private:
     {
         const int topLineHeight = screenHeight / 2;
 
-        if (stockData.empty())
+        StockData currentStock;
         {
-            g().DrawTextInBand("STOCKS", 0, topLineHeight, CRGB::White);
-            g().DrawTextInBand("--", topLineHeight, screenHeight - topLineHeight, CRGB::LightGrey);
-            return;
-        }
+            std::lock_guard dataGuard(stockDataMutex);
+            if (stockData.empty())
+            {
+                g().DrawTextInBand("STOCKS", 0, topLineHeight, CRGB::White);
+                g().DrawTextInBand("--", topLineHeight, screenHeight - topLineHeight, CRGB::LightGrey);
+                return;
+            }
 
-        const size_t stockIndex = std::min(static_cast<size_t>(std::max(0, iCurrentStock)), stockData.size() - 1);
-        auto it = stockData.begin();
-        std::advance(it, stockIndex);
-        const StockData& currentStock = it->second;
+            const size_t stockIndex = std::min(static_cast<size_t>(std::max(0, iCurrentStock)), stockData.size() - 1);
+            auto it = stockData.begin();
+            std::advance(it, stockIndex);
+            currentStock = it->second;
+        }
 
         const float displayPrice = currentStock.DisplayPrice();
         const String priceText = displayPrice >= 10000 ? String(displayPrice, 0) : String(displayPrice, 2);
@@ -270,6 +286,11 @@ private:
     static float JsonFloatOr(JsonVariantConst value, float fallback = 0.0f)
     {
         return value.is<float>() || value.is<double>() || value.is<int>() ? value.as<float>() : fallback;
+    }
+
+    static bool JsonHasNumber(JsonVariantConst value)
+    {
+        return value.is<float>() || value.is<double>() || value.is<int>() || value.is<unsigned int>();
     }
 
     static String JsonStringOr(JsonVariantConst value, const String& fallback = "")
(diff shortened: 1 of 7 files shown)


include/effects/matrix/PatternStocks.h
```diff
@@ -612,9 +798,12 @@ class PatternStocks : public EffectWithId<PatternStocks>
         // have and trigger a reload.
         if (FieldAccess::AssignIfSelected(name, NAME_OF(tickerSymbols), tickerSymbols, value))
         {
-            iCurrentStock = 0;
-            stockData.clear();
-            lastCount = SIZE_MAX;
+            {
+                std::lock_guard dataGuard(stockDataMutex);
+                iCurrentStock = 0;
```
rbergen: **Bug: `iCurrentStock` off-by-one — first ticker always skipped**

`iCurrentStock` is initialized to `0` at the class level (line 236) and reset to `0` here. The first call to `Update()` immediately executes `(0 + 1) % stockCount = 1`, so the display always starts on the **second** ticker — the first symbol in the list is never shown first, neither on startup nor after a ticker-symbols change.

Fix: initialize and reset to `-1` so the first increment yields `0`:
```cpp
// class definition:
int iCurrentStock = -1;

// here in SetField:
iCurrentStock = -1;
```

> **Note:** if you make this change, `UpdateQuoteDisplay`'s bounds guard (line 646) also needs a lower-bound companion. It currently only checks `iCurrentStock >= stockData.size()`, while `DrawCompactQuoteDisplay` already uses `std::max(0, iCurrentStock)` — a negative value reaching `std::advance` on a map iterator is UB.

include/effects/matrix/PatternStocks.h
```diff
@@ -317,7 +467,10 @@ class PatternStocks : public EffectWithId<PatternStocks>
             GetQuote(symbol, [this, callback](const StockData& stockDataReceived)
             {
                 if (!stockDataReceived.symbol.isEmpty())    // Check if the stock data is not empty
+                {
+                    std::lock_guard dataGuard(stockDataMutex);
```
rbergen, line 486: **Bug: stale symbol data re-inserted after ticker-symbols change**

When `SetField()` clears `stockData` (under `stockDataMutex`) and then calls `FlagReader()`, any `GetAllQuotes()` loop that was already running continues to completion — `FlagReader()` only sets a flag consumed on the next network-reader cycle; it does not cancel the running fetch. Each in-flight `GetQuote()` callback will reach this line, acquire `stockDataMutex`, and **re-insert its (now-stale) symbol data** back into the freshly-cleared map.

Fix: guard the insert against `tickerSymbols` to reject symbols that are no longer wanted:
```cpp
std::lock_guard dataGuard(stockDataMutex);
if (tickerSymbols.indexOf(stockDataReceived.symbol) >= 0)
    this->stockData[stockDataReceived.symbol] = stockDataReceived;
```

include/effects/matrix/PatternStocks.h
```diff
@@ -425,29 +579,41 @@ class PatternStocks : public EffectWithId<PatternStocks>
...
         auto pricelen  = pricetext.length();
         constexpr auto textwidth = 5;
 
-        auto changetext = String(data.close - data.previousClose, 2);
+        auto changetext = String(data.DisplayChange(), 2);
         auto changelen  = changetext.length();
 
         constexpr auto formatVolumeLargerThan = 1000000;
-        auto voltext = formatSize(data.volume, formatVolumeLargerThan);
+        auto voltext = data.hasVolume ? formatSize(data.volume, formatVolumeLargerThan, 0) : String("--");
```
rbergen: **Bug: `float volume` implicitly converted to `size_t` — UB on negative value**

`data.volume` is `float` but `formatSize()` takes `size_t`. If the server ever returns a negative volume (e.g. `-1` as a sentinel), `JsonHasNumber` sets `hasVolume = true` and the implicit `float → size_t` conversion here is **undefined behaviour** (C++17 §7.10: converting a negative float to an unsigned integer). On ESP32/Xtensa this typically wraps to a huge value, causing `formatSize`'s suffix loop to exhaust all entries and write garbage to the matrix.

Neither `JsonFloatOr` nor `formatSize` clamps against this; the call site is the right place to guard:
```cpp
auto voltext = (data.hasVolume && data.volume >= 0.0f)
    ? formatSize(static_cast<size_t>(data.volume), formatVolumeLargerThan, 0)
    : String("--");
```

include/effects/matrix/PatternStocks.h
```diff
@@ -203,66 +240,174 @@ class PatternStocks : public EffectWithId<PatternStocks>
...
+
+            JsonObjectConst day = doc["day"].as<JsonObjectConst>();
+            stockData.open   = JsonFloatOr(day["open"]);
+            stockData.hasOpen = JsonHasNumber(day["open"]);
+            stockData.high   = JsonFloatOr(day["high"]);
+            stockData.hasHigh = JsonHasNumber(day["high"]);
+            stockData.low    = JsonFloatOr(day["low"]);
+            stockData.hasLow = JsonHasNumber(day["low"]);
+            stockData.close  = JsonFloatOr(day["close"], stockData.officialClose);
+            stockData.hasClose = JsonHasNumber(day["close"]) || stockData.hasOfficialClose;
```
rbergen: **Bug: `hasClose` conflates intraday close with official close**

When `day["close"]` is absent (e.g. outside trading hours) but `officialClose` is present, this line sets `hasClose = true` and the fallback in `JsonFloatOr` stores `officialClose` in `close`. The debug CLI then prints `, Close X.XX` for a value that is actually the prior day's official close — a mislabelled field.

More importantly, any future code that branches on `hasClose` to distinguish intraday from settlement data will silently use the wrong value.

Consider setting `hasClose` only when `day["close"]` is explicitly present, and displaying `officialClose` under its own guard:
```cpp
stockData.close    = JsonFloatOr(day["close"]);
stockData.hasClose = JsonHasNumber(day["close"]);
// officialClose is already separately available via hasOfficialClose
```

include/effects/matrix/PatternStocks.h
```diff
@@ -471,28 +637,37 @@ class PatternStocks : public EffectWithId<PatternStocks>
...
+        StockData currentStock;
+        {
+            std::lock_guard dataGuard(stockDataMutex);
+            if (stockData.empty())
+                return;
 
-        auto it = stockData.begin();
-        std::advance(it, iCurrentStock);
-        StockData & currentStock = it->second;
+            if (iCurrentStock >= static_cast<int>(stockData.size()))
```
rbergen: **Inconsistent `iCurrentStock` bounds guard**

This only resets when `iCurrentStock >= stockData.size()` (upper bound). `DrawCompactQuoteDisplay` (line 279) uses `std::min(static_cast<size_t>(std::max(0, iCurrentStock)), ...)` which clamps both ends.

If the off-by-one above is fixed by initialising to `-1`, a negative `iCurrentStock` reaching `std::advance` on a bidirectional map iterator is UB. The lower-bound guard needs to be added here to match `DrawCompactQuoteDisplay`:
```cpp
if (iCurrentStock < 0 || iCurrentStock >= static_cast<int>(stockData.size()))
    iCurrentStock = 0;
```

rbergen (commented):
The thread-safety work is solid overall — the lock-copy-release pattern is correct and mutex discipline is consistent. A few correctness bugs worth fixing before merge.

rbergen (approved):
LGTM

PR #901, opened 2026-07-13
This change migrates Mesmerizer configurations from SmartMatrix to a new HUB75 layer, adds DEVKIT, DEVKIT S3

davepl:
Highlights:
Added ESP32-DevKitC V4 Mesmerizer configuration using the supplied HUB75 pin mapping. No PSRAM
Local effects only
Network and optional services disabled
BOOT button advances effects

Added ESP32-S3-DevKitC-1 N16R8 configuration.
8 MB octal PSRAM
16 MB flash
Dual 4 MB OTA application partitions
Native USB CDC logging
Audio disabled
BOOT button advances effects

Unified classic WROVER, Lolin, DEVKIT, and S3 Mesmerizer configurations on the new HUB75 DMA backend.

Restored splash-screen rendering during startup without waiting for Wi-Fi.

Preserved effect title behavior, including ShouldShowTitle() opt-outs for Weather and Stocks.

Corrected periodic serial status output and S3 USB CDC routing.

Added memory protections for low-memory targets:
Conditional network-frame buffer allocation
Allocation-safe logging
Reduced temporary settings-schema allocations
Graceful JSON allocation failure handling
Safer topology validation

Fixed JPEG decoder initialization before the early splash render.

Integrated the latest changes from main onto the new hub75 branch.

### DIFF
--- a/src/effects.cpp
+++ b/src/effects.cpp
@@ -141,6 +141,11 @@
 
 #define EFFECT_SET_VERSION 9
 
+// Central Adafruit-GFX font used by the matrix information effects.
+#if USE_MATRIX
+#include "FontGfx_apple5x7.h"
+#endif
+
 // Inform the linker which effects have setting specs, and in which class member
 
 //#if USE_HUB75 && ENABLE_WIFI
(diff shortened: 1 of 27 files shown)


src/webserver_settings.cpp
```diff
@@ -216,8 +216,14 @@ bool CWebServer::EnsureDeviceSettingSpecsJson()
...
+    try
     {
+        if (!BuildSettingSpecsJson(json, LoadDeviceSettingSpecs()))
+            return false;
+    }
+    catch (const std::bad_alloc&)
+    {
+        debugE("Insufficient memory to build device setting specs JSON.");
         return false;
     }
```
rbergen, line 228: Exactly what type of allocation failure is this supposed to catch? ArduinoJson solemnly swears it handles allocation failures elegantly, and the ways it reports them (`if (jsonDoc.overflowed())` specifically) is checked by the existing code in `BuildSettingSpecsJson()`.

src/effects.cpp
```diff
@@ -188,6 +200,13 @@ void LoadEffectFactories()
         );
     #endif
 
+    #if defined(EFFECTS_IMPERIAL)
```
rbergen, line 203: `EFFECTS_IMPERIAL` does not seem to be defined by any env in platformio.ini. Is this for future use?

rbergen (commented):
I have one question about an effect set define and one potential dead catch block. LGTM otherwise!

src/effects.cpp
davepl: Yes, its partial support for a new project I'm tinkering on.  I don't NEED to check it in now, it just happened to be part of it.

src/webserver_settings.cpp
davepl: Yes, in the default Arduino-ESP32 configuration, new / new[] throws std::bad_alloc on failure, and this catch (const std::bad_alloc&) can catch it.  You COULD just text for null on a malloc but on a new, it throws by default.

PR #902, opened 2026-07-25
M5TAB support

davepl:
This adds M5 TAB support, including its 1280x720 display.  Renders all effects, some better than others.  FIxed many effects to "draw up" in resolution, as many never had to draw outside the 63x32 box before!

### DIFF
--- a/include/effects/matrix/PatternAnimatedGIF.h
+++ b/include/effects/matrix/PatternAnimatedGIF.h
@@ -291,12 +291,24 @@ public:
         uint16_t gifWidth = gif->second._width;
         uint16_t gifHeight = gif->second._height;
 
-        // Map the complete GIF canvas to the complete matrix canvas. GIF
-        // assets were authored for several small aspect ratios (32x32,
-        // 64x32, 64x12); independent axes ensure every one fills modern
-        // widescreen matrices instead of remaining letterboxed.
-        const float scaleX = static_cast<float>(MATRIX_WIDTH) / gifWidth;
-        const float scaleY = static_cast<float>(MATRIX_HEIGHT) / gifHeight;
+        float scaleX;
+        float scaleY;
+        if (MATRIX_WIDTH > 64 || MATRIX_HEIGHT > 32)
+        {
+            // Large framebuffer builds need to use the available surface.
+            scaleX = static_cast<float>(MATRIX_WIDTH) / gifWidth;
+            scaleY = static_cast<float>(MATRIX_HEIGHT) / gifHeight;
+        }
+        else
+        {
+            // Preserve the original small-panel behavior: only downscale,
+            // retain aspect ratio, and center the result.
+            const float scale = std::min(
+                1.0f, std::min(static_cast<float>(MATRIX_WIDTH) / gifWidth,
+                               static_cast<float>(MATRIX_HEIGHT) / gifHeight));
+            scaleX = scale;
+            scaleY = scale;
+        }
 
         // Calculate the destination dimensions after scaling
         uint16_t dstWidth = (uint16_t)(gifWidth * scaleX);
--- a/include/effects/matrix/PatternAnimatedGIF.h
+++ b/include/effects/matrix/PatternAnimatedGIF.h
@@ -192,20 +192,28 @@ class PatternAnimatedGIF : public EffectWithId<PatternAnimatedGIF>
     static void drawPixelCallback(int16_t x, int16_t y, uint8_t red, uint8_t green, uint8_t blue)
     {
         auto& g = g_ptrSystem->GetEffectManager().g(0);
+        const auto& state = SharedGIFDecoderState();
+        const CRGB color(red, green, blue);
 
-        // Apply scaling transformation
-        int16_t scaledX = (int16_t)(x * SharedGIFDecoderState()._scaleX) + SharedGIFDecoderState()._offsetX;
-        int16_t scaledY = (int16_t)(y * SharedGIFDecoderState()._scaleY) + SharedGIFDecoderState()._offsetY;
-
-        if (false == g.isValidPixel(scaledX, scaledY))
+        if (state._scaleX >= 1.0f && state._scaleY >= 1.0f)
         {
-            debugV("drawPixelCallback: scaled pixel out of bounds: %d, %d (from source %d, %d)", scaledX, scaledY, x, y);
+            // Use adjacent transformed source boundaries so fractional scales
+            // cover every destination pixel without gaps.
+            const int16_t x0 = static_cast<int16_t>(x * state._scaleX) + state._offsetX;
+            const int16_t y0 = static_cast<int16_t>(y * state._scaleY) + state._offsetY;
+            const int16_t x1 = static_cast<int16_t>((x + 1) * state._scaleX) + state._offsetX;
+            const int16_t y1 = static_cast<int16_t>((y + 1) * state._scaleY) + state._offsetY;
+
+            g.fillRectangle(x0, y0, x1, y1, color);
             return;
         }
 
-        // If we're scaling down (scale < 1.0), we might want to sample multiple source pixels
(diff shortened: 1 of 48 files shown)


rbergen (commented):
Overall this looks good - solid work getting the Tab5 backend up and running. Two things below I think are blockers and should be fixed before merge. The rest are questions worth answering, plus a few nits. Your call on those.

src/network.cpp
```diff
@@ -290,7 +293,8 @@ namespace nd_network
         }
 
         // (Re)connect if credentials were explicitly provided, or our last attempt was long enough ago
-        if (explicitCredentials || millisAtLastAttempt == 0 || millis() - millisAtLastAttempt >= retryDelay)
+        if (!isStaAssociated() &&
+            (explicitCredentials || millisAtLastAttempt == 0 || millis() - millisAtLastAttempt >= retryDelay))
```
rbergen: Blocker: `!isStaAssociated()` gates the whole reconnect block, including the explicit-credentials path. That contradicts the comment right above it: if we're already associated and Improv submits new credentials, we now skip reconnecting entirely, so we keep serving the old network. Not scoped to the Tab5, it's a global change. Needs a fix, or a narrower guard around just the async-race case it's meant to address.

src/gfxbase.cpp
```diff
@@ -570,6 +609,52 @@ void GFXBase::blurRows(CRGB *leds, uint16_t width, uint16_t height, uint16_t fir
...
+    const bool rowMajor = width < 2 || (xy(0, 0) == 0 && xy(1, 0) == 1 && xy(0, 1) == width);
+    if (rowMajor)
+    {
+        // Keeping one carry value per column lets us traverse both source and
+        // destination rows sequentially. The former column-at-a-time walk
+        // incurred a PSRAM cache miss for nearly every pixel at 1280x720.
+        CRGB *scratch = static_cast<CRGB *>(
+            heap_caps_calloc(width * 2U, sizeof(CRGB), MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT));
+        if (!scratch)
+            throw std::runtime_error("Unable to allocate native blur scratch rows");
```
rbergen: Blocker: The row-major fast path in `blurColumns` mallocs and frees scratch memory on every call (same pattern repeats around line 654). `HUB75GFX::xy()` is row-major too, so this hits every existing HUB75 build, not just the Tab5, and `BlurFrame` runs once a frame in about ten built-in effects. That's heap churn in a per-frame hot path, on `MALLOC_CAP_INTERNAL`, on a device that runs 24/7. Make the buffer persistent and allocate it once.

include/effects/matrix/PatternAnimatedGIF.h
```diff
@@ -283,20 +291,12 @@ class PatternAnimatedGIF : public EffectWithId<PatternAnimatedGIF>
...
-            // Use the smaller scale factor to maintain aspect ratio (best fit)
-            float scale = min(scaleX, scaleY);
-            scaleX = scale;
-            scaleY = scale;
-        }
+        // Map the complete GIF canvas to the complete matrix canvas. GIF
+        // assets were authored for several small aspect ratios (32x32,
+        // 64x32, 64x12); independent axes ensure every one fills modern
+        // widescreen matrices instead of remaining letterboxed.
+        const float scaleX = static_cast<float>(MATRIX_WIDTH) / gifWidth;
```
rbergen: Question: independent-axis scaling now always stretches the GIF to fill the matrix, rather than only downscaling with aspect preserved when it's larger. A 32x32 GIF on a 64x32 panel gets stretched 2x horizontally now, so anything round goes elliptical. Is that intentional across all existing HUB75 hardware, or should it be scoped to wide panels?

include/effects/matrix/PatternSMSmoke.h
```diff
@@ -82,16 +83,24 @@ class PatternSMSmoke : public EffectWithId<PatternSMSmoke>
...
+    // the accumulated smoke motion remains comparable over two frames.
+    horizontalPass = !horizontalPass;
+    if (horizontalPass)
+    {
+      g().MoveFractionalNoiseX(2);
+      g().blurRows(g().leds, WIDTH, HEIGHT, 0, 20);
+    }
+    else
+    {
+      g().MoveFractionalNoiseY(2);
```
rbergen: Question: alternating the move/blur axis per frame instead of doing both is a fair trade at 160x90+, but it changes how the smoke effect looks on small panels that never had a performance problem. Worth gating on matrix size?

include/effects/matrix/PatternWave.h
```diff
@@ -103,56 +101,70 @@ class PatternWave : public EffectWithId<PatternWave>
...
-                    n = quadwave8(x * 2 - theta) / scale;
-                    if (n < MATRIX_HEIGHT)
+                    const uint8_t phase = static_cast<uint8_t>(
+                        rotation == 0 ? x * 2 + theta : x * 2 - theta);
+                    const int y = static_cast<int>(
+                        (static_cast<uint32_t>(quadwave8(phase)) * maxY + 127) / 255);
+                    const CRGB color = graphics.ColorFromCurrentPalette(x + hue);
+
+                    if (previousY >= 0)
+                        graphics.drawLine(x - 1, previousY, x, y, color);
```
rbergen, line 118: Question: `drawLine` interpolation (here and in the mirrored case 1/3 block below) fixes the gaps on wide matrices but also softens the wave on small ones, where it used to draw discrete points. Intentional for all sizes, or worth gating?

src/m5tabgfx.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+    _caption = caption;
+    _captionStartTime = millis();
+    _captionDuration = duration;
+}
+
+float M5TabGFX::CaptionTransparency() const
+{
+    constexpr uint32_t kFadeInTime = 500;
+    constexpr uint32_t kFadeOutTime = 1000;
```
rbergen: Nit: `kFadeInTime`/`kFadeOutTime` are defined twice (again at 132-133). Worth hoisting to class-level constants so they can't drift apart.

include/effects/matrix/PatternCircuit.h
```diff
@@ -213,15 +215,19 @@ class PatternCircuit : public EffectWithId<PatternCircuit>
...
+        // about 3, but selecting those pixels required thousands of random()
+        // calls per frame on larger matrices. A sequential fade is visually
+        // equivalent and substantially faster.
+        g().DimAll(252);
 
         // fill_palette(colors, SNAKE_LENGTH, initialHue++, 5, graphics.currentPalette, 255, LINEARBLEND);
         fill_palette(colors, SNAKE_LENGTH, 0, 4, ForestColors_p, 255, LINEARBLEND);
+        constexpr uint8_t kReferenceMatrixWidth = 64;
+        constexpr uint8_t kMovementStep =
+            std::max(1, (MATRIX_WIDTH + kReferenceMatrixWidth / 2) / kReferenceMatrixWidth);
```
rbergen, line 231: Nit: `kMovementStep` is derived from `MATRIX_WIDTH` only and reused for both axes on the non-square 160x90 surface, so vertical movement will read as proportionally faster than horizontal. Minor, and you already flagged uneven scaling in the PR description.

platformio.ini
```diff
@@ -289,6 +290,22 @@ build_src_flags = ${dev_m5.build_src_flags}
 board_build.partitions = config/partitions_custom_8M.csv
 board_upload.flash_size = 8MB
 
+[dev_m5_tab5]
+extends         = base
+platform        = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.31-1/platform-espressif32.zip
```
rbergen: Nit: `dev_m5_tab5` pins to a direct release-zip URL on a third-party `platform-espressif32` fork. Probably unavoidable until P4 support lands upstream, worth a comment saying so, so it's not mistaken for an oversight later.

